PR #6, opened 2018-11-12
Support arbitrary expression evaluation

jcrwilliams:
These commits add support for arbitrary expression evaluation using var-create, var-update, and var-delete behavior from CDT.


src/GDBDebugSession.ts
```diff
@@ -18,7 +18,6 @@ import { sendTargetAttachRequest } from './mi/target';
 import { sendThreadInfoRequest } from './mi/thread';
 import { sendStackInfoDepth, sendStackListFramesRequest, sendStackListVariables } from './mi/stack';
 import { sendVarCreate, sendVarListChildren, sendVarUpdate, MIVarCreateResponse, sendVarDelete } from './mi/var';
-import { MIVariableInfo} from './mi/base';
```
dschaefer: Missing space before the closing brace }. Fix next time.

jcrwilliams: The line itself was removed so... fixed? :)

dschaefer: Ah, yes. I guess that's what the red background means :)

PR #8, opened 2018-11-13
Compile in two steps, tsc + webpack

simark:
Instead of compiling the typescript and webpacking in one go, do the
process in two steps:

- Compile ts files to corresponding js files
- Webpack the js files

I find it useful to see the intermediary .js files and be able to
inspect them.  Also, debugging the webpack bundle does not work
currently.  Debugging the intermediary js files work.  We'll of couse
want to be able to debug the bundle, but this gives us a way to do it in
the mean time.

A functional change of this patch is that the bundle is now generated
to dist/bundle.js, which is a kind of de facto standard place for it.
it can be renamed to anything when distributed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -567,6 +567,12 @@
       "integrity": "sha1-JuYe0UIvtw3ULm42cp7VHYVf6Nk=",
       "dev": true
     },
+    "builtin-modules": {
+      "version": "1.1.1",
+      "resolved": "https://registry.npmjs.org/builtin-modules/-/builtin-modules-1.1.1.tgz",
+      "integrity": "sha1-Jw8HbFpywC9bZaR9+Uxf46J4iS8=",
+      "dev": true
+    },
     "builtin-status-codes": {
       "version": "3.0.0",
       "resolved": "https://registry.npmjs.org/builtin-status-codes/-/builtin-status-codes-3.0.0.tgz",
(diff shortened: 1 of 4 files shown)


jonahgraham:
ok to test

dschaefer:
> ok to test

Need to add you as an admin :). Feel free to tweak the job to add in this phrase. For now I have it set as

run tests

dschaefer:
Not sure the benefit of running both tsc and webpack. ts-loader essentially calls tsc. And, really, all the adapter is going to be is TypeScript so webpack has nothing else to do. I would think you'd use one or the other, not both.

I like shrinking this down to a single file for production. That's a pretty standard practice with npm packages. If you want to just call tsc in development and drop webpack, I could live with that.

simark:
> Not sure the benefit of running both tsc and webpack. ts-loader essentially calls tsc. And, really, all the adapter is going to be is TypeScript so webpack has nothing else to do. I would think you'd use one or the other, not both.

I have a slight preference for running the tools separately.  I find it's easier to understand the pipeline, and it's possible to run each step independently if needed.  Like if you are trying to debug a webpack issue, you just run webpack in isolation, there is less stuff invovled.  It just offers more flexibility than a single tool that does everything.  And there is no significant impact on the build time.

> I like shrinking this down to a single file for production. That's a pretty standard practice with npm packages. If you want to just call tsc in development and drop webpack, I could live with that.

I completely agree with packing everything to a single file for production.  I just think that having access to intermediary files can be interesting, for example when you need to dig in what the typescript compiler did.

dschaefer:
We've simplified this in #11

PR #10, opened 2018-11-13
Add tslint config file to get us started.

dschaefer:
Feel free to suggest changes here. This is just what we've been using for the last few months. I'm sure it could be made better.


dschaefer:
Sure. I can take a look at that and update this.

dschaefer:
I just grabbed the one from your change. Not seeing any warnings in vscode, but I'm not sure the check is running properly.

simark:
> I just grabbed the one from your change. Not seeing any warnings in vscode, but I'm not sure the check is running properly.

Hmm I do:

ss

dschaefer:
Ah, we didn't have tslint in the package.json file. See them now. I'll fix them all and resubmit.

simark:
> Ah, we didn't have tslint in the package.json file. See them now. I'll fix them all and resubmit.

Ah, I probably had it as a leftover.

dschaefer:
Yikes! Lots of little things but cleaned up nicely.

PR #11, opened 2018-11-13
Switch to use tsc directly.

dschaefer:
Keep webpack around for production when we get there.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -6,6 +6,7 @@
         "lib": [
             "es2015"
         ],
+        "outDir": "out",
         "strict": true,
         "noImplicitAny": true,
         "noImplicitReturns": true,
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -7,6 +7,7 @@
             "es2015"
         ],
         "outDir": "out",
+        "sourceMap": true,
         "strict": true,
         "noImplicitAny": true,
         "noImplicitReturns": true,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -8,7 +8,7 @@ module.exports = {
         extensions: [ '.ts', '.js' ]
     },
     entry: {
-        gdbDebugAdapter: './src/debugAdapter.ts'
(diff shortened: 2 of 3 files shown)


webpack.config.js
```diff
@@ -8,7 +8,7 @@ module.exports = {
         extensions: [ '.ts', '.js' ]
     },
     entry: {
-        gdbDebugAdapter: './src/debugAdapter.ts'
+        debugAdapter: './src/debugAdapter.ts'
```
simark: Don't you find it confusing that two different build artifacts share the same name?

tsconfig.json
```diff
@@ -6,6 +6,7 @@
         "lib": [
             "es2015"
         ],
+        "outDir": "out",
```
simark: Could you add `"sourceMap": true`?

webpack.config.js
dschaefer: two? No. In theory you'll never see the webpack one. That's created on the build machine. They need to be the same so that the clients that use it don't have to have different code for the different versions. I need to work on cdt-gdb-vscode to figure out how this will work.

tsconfig.json
dschaefer: Definitely. Stay tuned.

webpack.config.js
simark: I expect that clients will always use the bundled/minified version, so they can always point to that.  I expected the unpacked version to be useful more for developers of the adapter itself.  So I thought that having `yarn run build` produce both kinds of artifacts (unbundled and bundled js) would be the best of both worlds.

dschaefer: You'll never need both versions at the same time. I would think having them both there would be confusing for people.

For us this is going to get ugly as it is. We have a QNX "extension" to the adapter and then CDT and QNX versions of the vscode extension and will want to work on all four at the same time. This thing is just step one of a long journey for me :). And I still need to figure out how to plug this into the Eclipse version. Thus my desire to keep things as simple as possible.

PR #13, opened 2018-11-13
Split variable management into seperate file

jcrwilliams:
Move the variable management to a separate file to declutter the GDBDebugSession

Change-Id: I077c41fb1273422961ff839e91b365bf17e92a22

### DIFF
--- a/src/varManager.ts
+++ b/src/varManager.ts
@@ -1,0 +1,62 @@
+import {MIVarCreateResponse} from './mi/var';
+
+export interface VarObjType {
+  varname: string;
+  expression: string;
+  numchild: string;
+  value: string;
+  type: string;
+  isVar: boolean;
+}
+
+const variableMap: Map<string, VarObjType[]> = new Map<string, VarObjType[]>();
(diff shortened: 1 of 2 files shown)


dschaefer:
run tests

src/varManager.ts
```diff
@@ -0,0 +1,62 @@
...
+export interface VarObjType {
+  varname: string;
+  expression: string;
+  numchild: string;
+  value: string;
+  type: string;
+  isVar: boolean;
+}
+
+const variableMap: Map<string, VarObjType[]> = new Map<string, VarObjType[]>();
```
dschaefer: Given we have local state, would it make more sense to declare this as a class? Not sure, but checking on the thinking here.

jcrwilliams: I can't forsee a situation where we would want multiple instances of the varManager. Even with a multiple sessions scenario, it just becomes a matter of keying off the session identifier as well.

We can always swap it to a class later if we do find it necessary.

PR #16, opened 2018-11-14
Make module importable, add getExecPath function

simark:
The getExecPath returns the absolute path to the debug adapter entry
point.  The idea is that a javascript/typescript application can depend
on the cdt-gdb-adapter module, and call getExecPath to know which file
to spawn in order to start the adapter.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,8 @@
   "name": "cdt-gdb-adapter",
   "version": "0.1.0",
   "description": "gdb adapter implementing the debug adapter protocol",
-  "main": "index.js",
+  "main": "out/index.js",
+  "types": "out/index.d.ts",
   "scripts": {
     "build": "tsc",
     "watch": "tsc -w",
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,18 @@
+/*********************************************************************
+ * Copyright (c) 2018 Ericsson and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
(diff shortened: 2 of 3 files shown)


dschaefer:
run tests

simark:
Btw, this is inspired by how `vscode-ripgrep` does:

https://github.com/roblourens/vscode-ripgrep/blob/master/lib/index.js

I thought that using a function instead of a variable could give us more flexibility in the future (adding optional parameters that would influence what is returned).

dschaefer:
I'm generally OK with it. VS Code is pretty different with the debug adapter. I need to specify it statically in the package.json contributions section.

This might help with Eclipse since it'll need to start up node. I might need to package it up differently there though since Eclipse doesn't have npm :). That will be interesting.

dschaefer:
ip check didn't run. They're having a lot of issues with that. We can do it manually. Can you confirm you've signed the Eclipse CLA?

simark:
> I'm generally OK with it. VS Code is pretty different with the debug adapter. I need to specify it statically in the package.json contributions section.
> 
> This might help with Eclipse since it'll need to start up node. I might need to package it up differently there though since Eclipse doesn't have npm :). That will be interesting.

Indeed, for other languages (or the VSCode extension) it doesn't really help.

simark:
> ip check didn't run. They're having a lot of issues with that. We can do it manually. Can you confirm you've signed the Eclipse CLA?

Err in theory yes.  I went to check on Eclipse.org and it told me there was a new version.  It told me to revoke it and sign it again.  Then it's in an in-between state, saying this:

    You can't sign/revoke your ECA while we are processing your request. This message will disappear once everything is completed.

simark:
Ah, nevermind, I was able to sign it now.  Should be good.

PR #17, opened 2018-11-16
Add tslint-language-service

simark:
This patch allows seeing tslint warnings and errors in Theia, where
tslint must be used through the tslint-language-sevice plugin of tsc.
It doesn't seem to affect the editing experiecne in VSCode.


dschaefer:
run tests

PR #18, opened 2018-11-16
Add .editorconfig file, standardize formatting

simark:
I see quite a of bit of inconsistency in the formatting.  Let's add a
.editorconfig file, so the code we write uses the right whitespacing
regardless of the editor we use.  I just copied the one I wrote earlier
for Theia to get started, but it doesn't have to be identical.

I used Theia to auto-format all ts and json files and then tslint --fix
to fixup any warnings that it would have inserted.

For example, I the Typescript formatter handles quite badly (IMO) wrapping long
parameter lists, and produces code that result in a warning according to
our current tslint config.  The Typescript formatter will do this:

```
export class Bob {
    constructor(a: number, b: number, c: number, d: number,
        e: number, f: number) {
        console.log('Hello');
    }
}
```

where the second line of parameters is indented with 4 spaces.  I think
is not very readable, because it's easily confused with the first line
of code.  I prefer our current tslint setting, which requires it to be
aligned like this:

```
export class Bob {
    constructor(a: number, b: number, c: number, d: number,
i               e: number, f: number) {
        console.log('Hello');
    }
}
```

This means that formatting entire documents will always change the code
to look like the first snippet above.  It's a minor annoyance, and
hopefully there will be a setting for that one day in the Typescript
formatter.  There is an issue about that here:

https://github.com/Microsoft/TypeScript/issues/11629


simark:
Updated.  @dschaefer would it be possible to look at this soon, since the change of conflicts is quite hight?

dschaefer:
Standardizing on format is a controversial thing. Spaces versus tabs and all that. I need to think a bit about this. I'll get to it soon.

simark:
One thing that should be uncontroversial is that whatever rules are chosen, there should be means to enforce it (such as an `.editorconfig` file, or whatever config files are needed) and reduce the inconsistencies in the code base.

For Typescript code, I think using VSCode's default formatting options (when you tell it to auto-format the document) i good starting point.  Deviating from that creates additional work and annoyances, so should only be done when we think it's really worth it (such as the alignment of parameters mentioned above).

dschaefer:
At the end of the day though, the author needs to run the formatter, no? Or does having an .edtitorconfig file force it to?

simark:
> At the end of the day though, the author needs to run the formatter, no?

Yes.  Though many people around me like to have "format on save" on when writing Typescript.

> Or does having an .edtitorconfig file force it to?

If the editor supports .editorconfig files, it will at least set the correct mode of whitespaces (spaces vs tabs, size of indent, end of line, etc).  So let's say somebody creates a new JSON file, chances are it will have the right whitespacing, as chosen by the project (and not the otherwise default value provided by the editor).  As far as I know, .editorconfig doesn't go much further than that, but it's already a good start.  More complex language-specific formatting need language/tool-specific configuration, such as a .clang-format file for C++.

simark:
Ah, actually I left the parameter alignment as the Typescript formatter formats them.  In retrospect, I think it will be simpler to just use that formatting and disable the tslint rule...

PR #20, opened 2018-11-20
Add first, simple launch test

simark:
Add an integration test that sets a breakpoint, runs, then expect to hit
that breakpoint.

If you forget a step (forget to build the package or the test examples),
the errors are not very descriptive nor helpful, that's something I'd
like to improve in the future.


dschaefer:
Great start. Thanks!

simark:
Thanks!

PR #21, opened 2018-11-21
Fix calls to sendErrorResponse

simark:
At GDBBackend.ts:78, we reject a Promise with an object.  That rejected
promise may become an exception, caught for example in
setBreakPointsRequest at GDBDebugSession.ts:178.  That object is then
passed to sendErrorResponse.  The problem is that sendErrorResponse
expects to receive a string.  So it fails with:

```
(node:2643) UnhandledPromiseRejectionWarning: TypeError: format.replace is not a function
    at Function.formatPII (/home/emaisin/src/cdt-gdb-adapter/node_modules/vscode-debugadapter/lib/debugSession.js:669:23)
    at GDBDebugSession.sendErrorResponse (/home/emaisin/src/cdt-gdb-adapter/node_modules/vscode-debugadapter/lib/debugSession.js:310:41)
    at GDBDebugSession.<anonymous> (/home/emaisin/src/cdt-gdb-adapter/out/GDBDebugSession.js:137:22)
    at Generator.throw (<anonymous>)
    at rejected (/home/emaisin/src/cdt-gdb-adapter/out/GDBDebugSession.js:5:65)
    at <anonymous>
    at process._tickCallback (internal/process/next_tick.js:189:7)
```

Unfortunately, exceptions in Typescript are not typed, so it's difficult to
make sure what kind of object we caught.  To circumvent this, I suggest
we make sure to always throw and reject with Error (or subclasses of
it) containing a message.  This way, we know that we can use err.message
as an error message.

With this fix, a failed request is properly handled, we send a response
with "success":"false" to the client.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -75,15 +75,15 @@ export class GDBBackend extends events.EventEmitter {
                             resolve(result);
                             break;
                         case 'error':
-                            reject(result);
+                            reject(new Error(result.msg));
                             break;
                         default:
-                            reject('Unknown response ' + JSON.stringify(result));
+                            reject(new Error('Unknown response ' + JSON.stringify(result)));
                     }
                 });
                 this.out.write(`${token}${command}\n`);
             } else {
-                reject('gdb is not running.');
+                reject(new Error('gdb is not running.'));
             }
         });
     }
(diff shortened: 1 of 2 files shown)


dschaefer:
Sounds fine. Could you fix the signed off by?

run tests

simark:
> Sounds fine. Could you fix the signed off by?

Woops, done.

dschaefer:
Thanks!

PR #22, opened 2018-11-22
Add 'Set Value' Functionality for Variables

jcrwilliams:
Adds the ability to right-click > Set Value on most variables.

Limitations:
-Defined length arrays display the value of the array length when first
created, and the value of the first array entry that gets updated after
subsequent checks. It would be more useful to display the address the
array is located at in memory.
-Parent and child entries in the Variables tree are not updated when a
value is set, only the variable that was modified. This is purposefully
defined behavior by VS Code but not clear if this is modifiable.
-There exists a hook in the Debug Protocol to set expressions. I have not
yet found a way to trigger a SetExpressionRequest.

Refs eclipse-cdt#15

Change-Id: I6d3c072bf9e6438cd2f5371f87e8cfeb53da6ca5

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -18,7 +18,7 @@ import * as exec from './mi/exec';
 import { sendStackInfoDepth, sendStackListFramesRequest, sendStackListVariables } from './mi/stack';
 import { sendTargetAttachRequest } from './mi/target';
 import { sendThreadInfoRequest } from './mi/thread';
-import { sendVarCreate, sendVarDelete, sendVarListChildren, sendVarUpdate } from './mi/var';
+import { sendVarAssign, sendVarCreate, sendVarDelete, sendVarListChildren, sendVarUpdate } from './mi/var';
 import * as varMgr from './varManager';
 
 export interface LaunchRequestArguments extends DebugProtocol.LaunchRequestArguments {
@@ -48,6 +48,7 @@ export interface FrameVariableReference {
 
 export interface ObjectVariableReference {
     type: 'object';
+    frameHandle: number;
     varobjName: string;
 }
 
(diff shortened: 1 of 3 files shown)


src/GDBDebugSession.ts
```diff
@@ -457,6 +467,51 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+                                        varobjName: varobj.expression })
+                                    : 0,
+                                    indexedVariables: parseInt(varobj.numchild, 10),
+                                };
+            } else {
+                // we shouldn't hit this case
+                response.body = {value: args.value};
+            }
+        } catch (err) {
+            this.sendErrorResponse(response, 1, err);
```
simark: Can you change `err` to `err.message` (see #21) ?

simark:
I just tried a really simple case with Theia as a frontend, it works!

dschaefer:
run tests

simark:
@jcrwilliams, could you look into adding a test for this feature?  You can take inspiration from the existing one, and we can probably start factoring out things that are common to multiple tests into a test utils library.

dschaefer:
Verify job has been updated to run them.

simark:
Ah cool.  Are non-admins able to start tests using this "run tests" magical phrase?

dschaefer:
> Ah cool. Are non-admins able to start tests using this "run tests" magical phrase?

No. But there is a white list. I need to figure out how to set it up so it runs tests from you two automatically. Of course, my hope is that you both become committers and I can add you as admins.

jcrwilliams:
> @jcrwilliams, could you look into adding a test for this feature? You can take inspiration from the existing one, and we can probably start factoring out things that are common to multiple tests into a test utils library.

@simark I will be adding tests for this, yes. Is it possible to commit this as is and have the tests come later in a separate commit? I will not close the referenced issue until tests have been pulled into the main repo.

simark:
> @simark I will be adding tests for this, yes. Is it possible to commit this as is and have the tests come later in a separate commit? I will not close the referenced issue until tests have been pulled into the main repo.

I am fine with this.

dschaefer:
OK. Thanks guys!

PR #23, opened 2018-11-26
Add code code to build test program before running test.

dschaefer:
Test fails without it and we want to keep the build of the test
programs close to the tests that need them.

### DIFF
--- a/src/integration-tests/launch.spec.ts
+++ b/src/integration-tests/launch.spec.ts
@@ -21,7 +21,7 @@ const emptyProgram = path.join(__dirname, '..', '..', 'src', 'integration-tests'
 
 before(function() {
     // Build the test program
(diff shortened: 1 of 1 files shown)


src/integration-tests/launch.spec.ts
```diff
@@ -19,6 +20,9 @@ let dc: DebugClient;
 const emptyProgram = path.join(__dirname, '..', '..', 'src', 'integration-tests', 'test-programs', 'empty');
 
 before(function() {
+    // Build the test program
+    cp.execSync('make', { cwd: path.dirname(emptyProgram)});
```
simark: Could you introduce a variable like

    const testProgramsDir = path.join(__dirname, '..', '..', 'src', 'integration-tests', 'test-programs');

and then define `emptyProgram` using it?

    const emptyProgram = path.join(testProgramsDir, 'empty');

PR #31, opened 2018-11-27
Remove cross-env. Jenkins will set it's own vars.

dschaefer:


simark:
The dependency on cross-env is not needed anymore.

PR #32, opened 2018-11-28
Use -file-exec-and-symbol to specify binary

### LINKED ISSUE #29: Use -file-exec-and-symbols instead of passing program on command line
We currently pass the program name on the command line when using GDB.  When specifying a program that does not exist (or if there is any other error while loading the binary in GDB), we don't really get good error feedback.  I suggest using the `-file-exec-and-symbols` command after GDB was launched to load the program.  We will then get an error response with a proper error message.

simark:
We currently pass the program name on the GDB command line.  If, for
example, the program does not exist, the failure is silent (GDB outputs
an error as CLI output, but that's it).  The failure only manifests
itself when we try to run, but at this point the error is not precise:

    Error: No executable file specified.
    Use the "file" or "exec-file" command.

By using -file-exec-and-symbols, we are able to get a more precise
error:

    /does/not/exist: No such file or directory.

Fixes #29.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -48,7 +48,7 @@ export class GDBBackend extends events.EventEmitter {
 
     public launch(args: LaunchRequestArguments) {
         const gdb = args.gdb ? args.gdb : 'gdb';
-        const proc = spawn(gdb, ['--interpreter=mi2', args.program]);
+        const proc = spawn(gdb, ['--interpreter=mi2']);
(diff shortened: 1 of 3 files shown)


dschaefer:
Sweet. White listing works. Build was triggered automagically.

simark:
yay! And testing is faster than for CDT :)

dschaefer:
Looks good.

PR #34, opened 2018-11-28
Add attach test

simark:
Add some simple "attach" tests, one when we successfully attach and one
when we provide a non-existent pid.

I kept it really simple, because I don't yet know what is the right
behavior we should test for.  For example, if the client sends a
configurationDone after a failed attach request, what should we reply?
So the tests can be improved at a later point.

I started to move things out of the existing test to a shared library of
test utilities.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1510,8 +1510,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "aproba": {
           "version": "1.2.0",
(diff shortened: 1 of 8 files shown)


simark:
We'll need to investigate why it fails on Jenkins... it looks like the debug client doesn't get any reponse at all for the attach request.

simark:
It would be a good opportunity to improve how we do logging and make it available after a Jenkins run :)

dschaefer:
In theory, you could also run in verbose mode and add a listener to LogOutputEvents to see what gdb is doing.

simark:
I have the intuition that #38 will help... let's see.

jonahgraham:
Sorry, this is the PR that hung the build: https://ci.eclipse.org/cdt/job/cdt-gdb-adapter-verify/41/console

I am cancelling it and putting a timelimit on the build time.

dschaefer:
Now that logging is working, we should generate and archive the log.

dschaefer:
Just ran into something similar. It looks like attach was missing the sendFileExecAndSymbols call. I fixed that in my latest update. And, yes, I did see that in the new log file :).

Suggest rebasing and trying again.

jonahgraham:

PR #37, opened 2018-11-29
Add logging with bunyan

simark:
The GDBDebugSession class currently inherits from LoggingDebugSession
(from vscode-debugadapter), but I am pretty sure logging is not
functional.  In particular, we pass it a relative file name, but the
logger code requires it to be absolute:

    https://github.com/Microsoft/vscode-debugadapter-node/blob/ac2d3a31cf5e71999e28c191f0b80a56439954bc/adapter/src/logger.ts#L144

Therefore I conclude it's not used at the moment.

Instead of using LoggingDebugSession, which is quite rudimentary, I
suggest using a dedicated logging library.  This will give us a bit more
control on what we log, where it goes, etc.  I have tried winston and
bunyan, two popular libraries, but got better and easier results with
the bunyan, so I went with that.

With this patch, there is a default stream to stderr which shows
warnings and above.  Passing `logFile` in a launch request enables a log
file with levels `debug` and above to be written, with all the DA and MI
communications.

Note that the logs are written as json files, and the `bunyan` CLI tool
can be used to display them nicely,

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4,6 +4,15 @@
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
+    "@types/bunyan": {
+      "version": "1.8.5",
+      "resolved": "https://registry.npmjs.org/@types/bunyan/-/bunyan-1.8.5.tgz",
+      "integrity": "sha512-7n8ANtxh2c5A/NfCuv8cVtWcgSLdq76MQbtmbInpzXuPw4TSAReUJ+MGHK4m67I4zI3ynCJoABfaeHYJaYSeRg==",
+      "dev": true,
+      "requires": {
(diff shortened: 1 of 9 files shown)


simark:
The next step would be to have the tests automatically log somewhere (one file per test), and have the logs available on Jenkins.

dschaefer:
I'll have to create a CQ request for the new dependency.

dschaefer:
Is this all because we don't pass an absolute path to the log file? It makes sense relative wouldn't work since we don't really know what the cwd is for the adapter.

Also looking at the source for the Logger, we're using an obsolete API to set it up. We should be calling Logger.setup and pass an absolute file path there.

jonahgraham:

PR #42, opened 2018-12-03
Fix up logging by making the logFile an arg.

dschaefer:
The logFile has to be absolute. It's easiest if we let
the user set that since they know where the would want it.
Adding to the launch and attach args and setting it in the
logger.setup.


simark:
I would suggest doing as I did in #37 to reduce code duplication, but otherwise it LGTM.

dschaefer:
I'll leave it as is for now. I need to focus on the features missing from the adapter. We can clean things up as we progress.

PR #43, opened 2018-12-05
Prepare for vendor specific GDB's.

dschaefer:
The QNX gdb has a couple of different commands including the upload
command, so removing that from the base here.

Added an API to allow vendors to override the backend object.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -16,7 +16,7 @@ import {
 import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
 import { GDBBackend } from './GDBBackend';
 import { sendBreakDelete, sendBreakInsert, sendBreakList } from './mi/breakpoint';
-import * as exec from './mi/exec';
+import { sendExecArguments, sendExecContinue, sendExecRun, sendExecNext, sendExecStep, sendExecFinish } from './mi/exec';
 import { sendStackInfoDepth, sendStackListFramesRequest, sendStackListVariables } from './mi/stack';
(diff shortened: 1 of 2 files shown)


simark:
We will need to deal with the quirks of the multiple GDB versions as well, like is done in CDT.  Do you think we can use this feature (having multiple backends) will help for that too?

src/GDBDebugSession.ts
```diff
@@ -16,7 +16,7 @@ import {
 import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
 import { GDBBackend } from './GDBBackend';
 import { sendBreakDelete, sendBreakInsert, sendBreakList } from './mi/breakpoint';
-import * as exec from './mi/exec';
+import { sendExecArguments, sendExecContinue, sendExecRun, sendExecNext, sendExecStep, sendExecFinish } from './mi/exec';
```
simark: Seeing this gave me an idea to reduce the number of imports we need:

https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/44

dschaefer:
> We will need to deal with the quirks of the multiple GDB versions as well, like is done in CDT. Do you think we can use this feature (having multiple backends) will help for that too?

It's a simple approach. Not sure it'll cover all the multiple version cases we'll have but it's a start. I'm not super happy with it yet. I had to override and essentially cut and past the upstream launch and attach request methods to insert the couple of gdb extension calls we have in the middle. I'm sure we can come up with a better way. I just need to demo this on Friday, so keeping it dumb and simple for now.

PR #44, opened 2018-12-05
Export all symbols from mi directory

simark:
Instead of importing all MI commands-related functions under mi/
separately, add an index.ts which exports everything from the mi
directory.  This lets us have a single import for all mi commands.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -974,16 +974,6 @@
         "sha.js": "^2.4.8"
       }
     },
-    "cross-env": {
-      "version": "5.2.0",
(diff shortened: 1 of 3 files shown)


simark:
> Nice, I like it.

Woops, except that I didn't committed all the changes.

dschaefer:
@jcrwilliams what do you think? I kinda like this. I've seen this pattern in other libraries.

dschaefer:
LOL, I could just read the early comments too :).

dschaefer:
Thanks @simark. Sold! :)

PR #45, opened 2018-12-06
Update build instructions in README

simark:


dschaefer:
I'm still wrestling with how to build this thing so that the npm package will be usable both by clients and by extenders.

Right now, for example, the cdt-gdb-vscode extension has the adapter as a submodule. It's build loads up the webpack.config.js file from the adapter and does a multi-config build to produce both the extension and it's own version of the adapter.js file. That's not what we want once we make this an npm package. The vscode extension will just load it directly from the node_modules directory.

For extenders who have a gdb variant and need to subclass the session and/or the backend they need access to the class definitions and other API. I'm again doing a submodule for our QNX extension where again we use webpack to build our adapter which sucks in code from this one through imports. In theory we should be using the npm package as well in which case I need a definitions file generated as well and to make sure everything is exported.

I bring this up since you state npm run build uses webpack. Right now it doesn't. But as I think ahead to the npm package, maybe it should be bundling using webpack to keep downstream users of this thing simple. I need to look around at more TypeScript libraries to see how others are doing this.

dschaefer:
> > I bring this up since you state npm run build uses webpack. Right now it doesn't. But as I think ahead to the npm package, maybe it should be bundling using webpack to keep downstream users of this thing simple. I need to look around at more TypeScript libraries to see how others are doing this.
> 
> Hmm maybe you've read the diff, backwards, I actually remove the mention to webpack.

That would be true :).

> 
> One way to achieve what you want might be to do like vscode's `DebugSession` does, where you pass in the type of the backend you want when constructing a debug session.
> 
> Or, we could use a dependency injection library such as inversify, which would make it easy to take what's in the upstream adapter, and override some types. Also, it helps a lot to set things up for unit testing.

I'd rather not use too much magic. Injection just makes it hard to figure out where things come from. On big systems, I've seen it help and maybe we'll get there as we consider supporting multiple gdb versions.

We'll let that soak for a bit. I need to see more examples of how this works before I form a strong opinion.

In the meantime, let's move this discussion to an issue and push in this PR.

PR #46, opened 2018-12-06
Adjust rules to add space before parentheses of anonymous function

simark:
By default, Theia (typescript-language-server, in fact) and VSCode don't
auto-format anonymous functions the same way:

    Theia:  function() {
    VSCode: function () {

I don't really mind which one we use, I just want it to be consistent
and avoid spurious diffs when reformating.  I chose the VSCode way.

This patch adds a tsfmt.json file, which typescript-language-server
reads.  It also adjusts the tslint rule to match, and fixes all
offenders.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,2 +1,3 @@
 {
+  "typescript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false
 }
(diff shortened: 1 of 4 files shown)


dschaefer:
I rarely see in the wild examples where there is a space between function and (). I'd rather it be the Theia way.

simark:
> I rarely see in the wild examples where there is a space between function and (). I'd rather it be the Theia way.

Ok, I just took the VSCode defaults as the "de facto" defaults.  I'll update my PR to do the opposite.

Issue #25, opened 2018-11-26
Improve C++ support in Variables view


jcrwilliams:
Investigate and address any deficiencies relating to how variables are displayed and set.  Have not yet done any testing/experimentation with how C++ programs behave in this view so this is a blanket issue for any findings that occur.

jcrwilliams:
As it stands, we do not display variables in C++ objects properly, as they exist under a valueless, typeless child elements reflecting their public/protected/private status. I will be adding support for this in an upcoming patch.

Interestingly, my investigation shows that the Eclipse CDT debug scenario chokes on C code that uses public/protected/private as a variable name inside of a structure. eg)
```
struct foo {
    int public;
};
```

simark:
> Interestingly, my investigation shows that the Eclipse CDT debug scenario chokes on C code that uses public/protected/private as a variable name inside of a structure. eg)
> 
> ```
> struct foo {
>     int public;
> };
> ```

Ha, nice test case :)

PR #52, opened 2018-12-19
Revamp vars support to fix issues with displaying/setting objects and structs.

jcrwilliams:
This patch addresses several issues with objects and structs that were
causing elements not to display properly in the variables view, and was
preventing the proper setting of variables in several cases.

This patch also creates some entry points for C++ unit tests. I will be
uploading subsequent patches to expand unit test coverage based on my
by-hand corner case testing.

Issue #25

Change-Id: If2e4d015ff90e1ba5c75e56c4ab149c07fd74d47


src/GDBDebugSession.ts
```diff
@@ -357,7 +357,10 @@ export class GDBDebugSession extends LoggingDebugSession {
                             if (update) {
                                 if (update.in_scope === 'true') {
                                     numVars++;
-                                    varobj.value = update.value;
+                                    if (update.name === varobj.varname) {
```
simark: The `variablesRequest` method is getting quite big... could you look into splitting it?

jcrwilliams: Done. I split the 'frame' and 'object' type requests into a submethod each. This should make the code a bit more comprehensible. :) I intend to do another pass on it and add some better comments when I'm juggling a few less things in my workspace than I am right now.

simark: > Done. I split the 'frame' and 'object' type requests into a submethod each. This should make the code a bit more comprehensible. :) I intend to do another pass on it and add some better comments when I'm juggling a few less things in my workspace than I am right now.

Thanks!  I don't know how this request works in details yet, so I can't comment on your changes.  But improving the readability will help me get there :).

src/integration-tests/vars_cpp.spec.ts
```diff
@@ -0,0 +1,68 @@
...
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *********************************************************************/
+
+import { expect } from 'chai';
+import * as cp from 'child_process';
+import * as path from 'path';
+import { DebugClient } from 'vscode-debugadapter-testsupport';
+import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
+import { getExecPath } from '..';
```
dschaefer: This has been removed from the index.ts which is now used for the external API. Check with the other tests on how to do this now.

simark:
At least before merging the PRs, could we squash the fixups back in the original commit, to avoid having broken intermediate states in the permanent history?  It's also not really useful to have all the commits that address review comments in the repo history.

Also, could we avoid merge commits?

simark:
Oh sorry!  Now that I look at the actual history, I see that it has been squashed.  I did not understand this initially based on Github's interface.

dschaefer:
Yup, I always squash the commits if I can't rebase.

PR #53, opened 2018-12-19
Add npmignore to bring sanity to the packaged tarball.

dschaefer:
yarn pack was giving us garbage.


dschaefer:
Pushing in now so I can test the build.

Issue #27, opened 2018-11-27
Memory View


dschaefer:
Add the ability to fetch blocks of memory using the data-read-memory MI command and friends.

Create React components to show the user the results. We can use these view components in both VS Code and Theia. If we ever figure out how to get a good browser in the Eclipse IDE, we can use it there too.

simark:
I am planning to work on the adapter bits to read memory very soon (probably this week).

dschaefer:
Very cool. Thanks Simon!

dschaefer:
BTW, is your work on this with Theia somewhere in the open? I'd like to see how hard it would be to provide a UI for this that works with both Theia and vscode.

mikhailkhod:
There is no specifications regarding memory fetching for DAP (https://microsoft.github.io/debug-adapter-protocol/specification). I assume, defining it is a part of the task. Is this correct?
Just for clarification, "data-read-memory" command is deprecated. We use "data-read-memory-bytes" in CDT.

simark:
> There is no specifications regarding memory fetching for DAP (https://microsoft.github.io/debug-adapter-protocol/specification). I assume, defining it is a part of the task. Is this correct?

Exact.

> Just for clarification, "data-read-memory" command is deprecated. We use "data-read-memory-bytes" in CDT.

Exact.

simark:
> BTW, is your work on this with Theia somewhere in the open? I'd like to see how hard it would be to provide a UI for this that works with both Theia and vscode.

It's still just a local draft, but in the following days weeks, I should have a PR on Theia to start using this adapter instead of code-debug.

simark:
I have something basic but working on this branch:

https://github.com/simark/cdt-gdb-adapter/tree/simark/zoe_chartrand

I'll wait until I can try it in an actual IDE before submitting it, to see if the request/response formats make sense.

dschaefer:
Cool. Looks good so far. I'll start thinking about how we'd do this in code. My current thinking is to create a simple command that prompts for the start address and length and opens up a webview with a simple dump of the result. Not sure yet how to get a handle on the debug client tho.

jonahgraham:
Might be worth looking at https://github.com/jonahgraham/cortex-debug which does roughly what you have outlined (for disassembly too)

dschaefer:
[...]

PR #54, opened 2018-12-20
Add basic memory reading capability

simark:
This patch adds a custom request (not specified in the DAP)
'cdt-gdb-adapter/Memory' to let the frontend read memory from the
target.

For now, it is assumed that

- The address is an integer (I think we could allow passing an
  expression in the future, anything that GDB can evaluate, like
  "&variable" or "base_ptr + offset").
- The result contains exactly one memory block.

Refs #27


dschaefer:
Looks good!

PR #55, opened 2018-12-21
Increased test coverage for variable getting/setting

jcrwilliams:
Improves the coverage of different use cases for getting and setting
variables in C and C++ programs

Issue #25 

Change-Id: Ib89e7dd11ff98641c5449308cb964ce5350004e9


jcrwilliams:
I'll be adding some additional test cases over the coming weeks to cover more of the corner cases I've found but it might as well toss what I have working so far up.

jcrwilliams:
> You can ping me when this is ready to review.

This should be good to review now.

src/integration-tests/test-programs/vars_cpp.cpp
```diff
@@ -30,10 +30,7 @@ int main() {
 	Foo *fooA = new Foo(1, 2, 'a');
 	Foo *fooB = new Foo(3, 4, 'b');
 	Foo *fooarr[] = {fooA, fooB};
-	// operator= on dereferenced objects is bad practice
-	Foo foo0 = *fooarr[0];
-	Foo foo1 = *fooarr[1];
-	if (foo0.a || foo1.a)
-		cout << "!!!Hello World!!!" << endl;
+    cout << "!!!Hello World!!!" << endl;
```
simark: The new indentation is right (in concordance with our .editorconfig), but different from the other lines.  Could you first make a commit that fixes the indentation in this file?  It can also fix other c/cpp/h files if they have the same problem.

src/integration-tests/var.spec.ts
```diff
@@ -56,45 +52,168 @@ describe('Variables Test Suite', function() {
...
-        expect(vars.body.variables[1].name).to.equal('b');
-        expect(vars.body.variables[1].value).to.equal('2');
-        expect(vars.body.variables[1].type).to.equal('int');
+        expect(vars.body.variables.length, 'There is a different number of variables than expected').to.equal(numVars);
+        expect(vars.body.variables[0].name, 'The variable name is wrong').to.equal('a');
+        expect(vars.body.variables[0].value, 'The variable value is wrong').to.equal('1');
+        expect(vars.body.variables[0].type, 'The variable type is wrong').to.equal('int');
+        expect(vars.body.variables[1].name, 'The variable name is wrong').to.equal('b');
+        expect(vars.body.variables[1].value, 'The variable value is wrong').to.equal('2');
+        expect(vars.body.variables[1].type, 'The variable type is wrong').to.equal('int');
```
simark: I don't think this text really helps.  It makes the code here heavier, and it kind of states the obvious.

I also think it would be good to avoid some duplication and make a function to assert the contents of a variable, like:

```ts
function verifyVariable(variable, expectedName, expectedValue, expectedType): boolean {
    expect(variable.name).eq(expectedName);
    expect(variable.value).eq(expectedValue);
    expect(variable.type).eq(expectedType);
}
```

.vscode/launch.json
```diff
@@ -0,0 +1,36 @@
+{
```
simark: This file should be added in a separate PR.

simark:
Could you re-work your branch to avoid adding things in a commit, only to remove (or change it completely) in a further commit?  It is perfectly fine (and even preferable) to split things in multiple small, well focused commits,  but each commit should build on the previous one, not undo what the previous ones do.

In other words, it's fine to have multiple commits if they incrementally build something complex, but not if they represent the different changes of direction you took while writing the code.

src/integration-tests/test-programs/.gitignore
```diff
@@ -4,3 +4,4 @@ vars
 *.o
 .vscode
 /vars_cpp
+/log
```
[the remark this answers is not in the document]
jcrwilliams: As a means of debugging some problems I encountered while developing these tests, I was using the logging we can generate from GDBDebugSession. It was added to .gitignore to keep a cleaner view of added/modified files and to prevent me from accidentally checking in the logfile itself.

src/integration-tests/var.spec.ts
jcrwilliams: As a bit of push back on having a message associated with the expect(), I found it useful for determining what was failing. Otherwise the test will fail with a message like 'AssertionError: expected 8 to equal 6' which is completely useless information without context. I agree with the critique that my current messages aren't all that useful though. I think with the function you suggested I can provide a better solution.

src/integration-tests/test-programs/vars_cpp.cpp
jcrwilliams: I think the discrepancy is that the unmodified lines are using tab characters. Visually speaking the file looks homogeneous wrt indentation when viewed in editor with tab = 4 spaces. But I agree that combined tab/space indenting is poor practice. Interesting that github's diff view defaults tab = 3 spaces, I suppose to highlight precisely this issue. :)

src/integration-tests/var.spec.ts
simark: I see, in my memory, the error messages were a bit more expressive.  One this that doesn't help is that the stack shows locations in the generated js files.  I just uploaded #58, which makes mocha report the locations in the source .ts files, it should help, since you can directly go to the error line.

If you decide to do as I suggested and use a helper function, then  you could do:

```
function verifyVariable(variable, expectedName, expectedValue, expectedType): boolean {
    expect(variable.name, `The name of ${expectedName} is wrong`).eq(expectedName);
    expect(variable.value, `The value of ${expectedName} is wrong`).eq(expectedValue);
    expect(variable.type, `The type of ${expectedName} is wrong`).eq(expectedType);
}
```

Since it's not duplicated, it doesn't add as much noise.

src/integration-tests/test-programs/vars_cpp.cpp
simark: Exactly, the new lines respect the standards we have established, but the existing lines don't.

src/integration-tests/var.spec.ts
jcrwilliams: This was exactly my thinking.

jcrwilliams:
I rebased this branch onto master - each commit should represent a clean changelog now without all the extraneous things that made it in to my previous set of commits. Further comments welcome :)

dschaefer:
I'll be squashing the commit when I push it. I'm not sure it makes sense to look at all the commits individually unless you curious about the process that went into the PR. I just look at the Files Changed tab here: https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/55/files

dschaefer:
This is our first time with Github but give than every time you push to your branch after creating the PR, the PR gets updated tells me that the desired workflow is to keep creating new commits and pushing them as you update the PR.

This would be similar to Gerrit as you keep pushing updates to your change request, except there you need to keep amending the commit. But doesn't work on Github since you need to create a new branch every time you do that. You can force push it I guess, but then you need to remember to do that all the time.

PR #58, opened 2019-01-09
Add source map support when running tests

simark:
Currently, when we have a test failure, we get a js backtrace:

      at Context.<anonymous> (dist/integration-tests/var.spec.js:71:60)
      at Generator.next (<anonymous>)
      at fulfilled (dist/integration-tests/var.spec.js:13:58)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)

With this patch, mocha makes use of the source maps to show the locations in
the original source files:

      at Context.<anonymous> (src/integration-tests/var.spec.ts:64:49)
      at Generator.next (<anonymous>)
      at fulfilled (dist/integration-tests/var.spec.js:13:58)
      at <anonymous>
      at process._tickCallback (internal/process/next_tick.js:189:7)

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,8 +8,8 @@
     "build": "tsc",
     "watch": "tsc -w",
     "test": "yarn test:integration && yarn test:pty && yarn test:integration --run-in-terminal",
-    "test:integration": "mocha --reporter mocha-jenkins-reporter dist/integration-tests/*.spec.js",
-    "test:pty": "mocha --reporter mocha-jenkins-reporter dist/native/*.spec.js"
+    "test:integration": "mocha --require source-map-support/register dist --reporter mocha-jenkins-reporter dist/integration-tests/*.spec.js",
+    "test:pty": "mocha --require source-map-support/register dist --reporter mocha-jenkins-reporter dist/native/*.spec.js"
   },
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


dschaefer:
The verify job failed. Any thoughts why? Anything weird in the console output of the job?

simark:
Hmmm no idea :(.  It passes here.

dschaefer:
I'll try and rerun it.

rerun tests

simark:
Same failure.  Is it possible to log in the machine to try by hand?

dschaefer:
Nope, security and stuff. I can change the script to gather more info if you think that would help.

dschaefer:
Switching the script to run yarn instead of npm. probably not what's causing this but to eliminate one difference.

rerun tests.

simark:
Do we know what environment this runs?  How node/npm/yarn are installed?  I could try to reproduce it in Docker.

dschaefer:
It's a standard image from the foundation staff. Going to double check if gcc and gdb are installed correctly.

rerun tests

dschaefer:
When I run your change locally, the yarn test actually hangs at the end. Which is weird since there's nothing showing in top so it looks like mocha is hanging waiting for something.

checking which version of node we have. Yarn is at 1.12.1 which should be fine.

rerun tests

simark:
Huh you are right, it hangs here too.  Even without this patch, the test takes a suspiciously big amount of time (a few seconds) before exiting.  But with this patch, it hangs forever.

It doesn't seem to be what is happening on Jenkins though.  There, it looks like the tests don't run at all.

dschaefer:
rerun tests. See if things are better now.

simark:
I have rebased the patch, which had some merge conflicts anyway.  On my computer, it still hangs at the end of the testsuite.

jonahgraham:

PR #63, opened 2019-01-15
Clean up termination.

dschaefer:
On exit we need to interrupt gdb for the exit command to work
without hanging. For remote targets, we also need to kill the inferior.
That should be fine for local targets as well.

This change adds interrupt to the backend and then changes
the disconnnect request to follow the steps above.

Also fix up logger. When subclassing we had trouble getting the right
logger setup.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -8,11 +8,10 @@
  * SPDX-License-Identifier: EPL-2.0
  *********************************************************************/
 import * as path from 'path';
-import { logger } from 'vscode-debugadapter/lib/logger';
 import {
-    Handles, InitializedEvent, Logger, LoggingDebugSession, OutputEvent, Response, Scope, Source,
+    Handles, InitializedEvent, Logger, logger, LoggingDebugSession, OutputEvent, Response, Scope, Source,
     StackFrame, StoppedEvent, TerminatedEvent, Thread,
-} from 'vscode-debugadapter/lib/main';
+} from 'vscode-debugadapter';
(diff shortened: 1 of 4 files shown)


dschaefer:
Note, I have implemented the pauseRequest here but haven't figured out how to invoke it from VS Code. The pause button doesn't do it for some reason.

src/GDBDebugSession.ts
```diff
@@ -536,9 +539,18 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+        args: DebugProtocol.PauseArguments) {
+        this.gdb.interrupt();
+        this.sendResponse(response);
+    }
+
     protected async disconnectRequest(response: DebugProtocol.DisconnectResponse,
         args: DebugProtocol.DisconnectArguments): Promise<void> {
         try {
+            this.gdb.interrupt();
+            await this.gdb.sendSet('confirm', 'off');
```
simark: Which version of GDB is this needed for?

dschaefer: -gdb-exit hung waiting for confirmation. This was while gdb was connected to a remote target since it wanted confirmation that it was about to detach from the inferior and leave it running. I can play around with that a bit more and see if I can avoid it.

simark: No need to feel bad, if there's a bug we need to work around it.  It's just that with my latest build of GDB there does not seem to be a problem doing either `kill` or `-gdb-exit` while the program is stopped at a breakpoint.

In other words, it's fine to have a hack, we just need to document well which situation/version of GDB it applies to, so we know when we no longer need it.

PR #67, opened 2019-01-16
Fix warnings

simark:


simark:
I was missing the `Signed-off-by`, I added that.  And there is a second warning that I fix at the same time.

PR #70, opened 2019-01-22
Add function to resolve line tags in tests

simark:
Hard-coding line numbers in tests makes them quite fragil and hard to
modify.  This patch allows us to place tags in test source files, and
then get the source line number corresponding to a tag.  This makes it
possible to modify the test source file without having to change all the
hard-coded line numbers.

### DIFF
--- a/src/integration-tests/test-programs/vars.c
+++ b/src/integration-tests/test-programs/vars.c
@@ -16,7 +16,7 @@ int main()
 {
     int a = 1;
     int b = 2;
-    int c = a + b;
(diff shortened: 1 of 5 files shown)


dschaefer:
The test failures don't lgtm :). Any thoughts why it's failing here?

simark:
I have rebased, the tests pass now.  I don't think this patch in particular is source of flakiness, the flakiness is already there :).

dschaefer:
Yeah, they're pretty random at the moment. Thanks!

PR #72, opened 2019-01-30
Add `openGdbConsole` option

marechal-p:
Split the changes in 2 commits:
- Adding the feature by mimicking how Eclipse CDT did it.
- Testing the feature with all the tests, extending the `DebugClient` to support reverse requests.

See the commit messages for more information.

Work to be done in a patch after this one:
- Supporting GDB async events on the MI channel.

Without the event support, the GDB console can be used by the user driving the debug session, but the UI will get de-synchronized, because the debug-adapter is the one supposed to notify the client of the debug state being changed, which is not currently handled.

I assumed it would be simpler to first merge this change, while I come up with the next one.


dschaefer:
run tests

marechal-p:
Ah the infamous "works on my machine".

CI failure seems to come from the step where I re-run using the two UIs.

I will add a test for the pty creation, maybe this is not supported on the runner?

dschaefer:
Yeah, not sure. Could also be an issue with the mock debug client it uses.

marechal-p:
@dschaefer I added tests for the pty creation and usage.

If the new tests fail I will investigate the modified `DebugClient`.

dschaefer:
rerun tests. Also, it's showing a merge conflict. Can you rebase? Thanks!

marechal-p:
Thank you, rebased.

dschaefer:
rerun tests. Thanks! :)

dschaefer:
Very cool! Thanks.

PR #74, opened 2019-02-05
Publish native sources

marechal-p:

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,5 +38,12 @@
     "vscode-debugadapter-testsupport": "^1.32.0",
     "webpack": "^4.23.1",
     "webpack-cli": "^3.1.2"
-  }
+  },
+  "files": [
+    "dist/**/*.js",
+    "dist/**/*.js.map",
+    "dist/**/*.d.ts",
+    "src/**/*.cc",
+    "src/**/*.h"
+  ]
 }


marechal-p:
Because the sources for the native code are in src, we end up not publishing them.

Didn't know about this, my bad.

dschaefer:
I don't think our tests test this so I'll just push it.

BTW, I'm still not comfortable publish sources up to npm. Not sure why consumers need it and if they do why npm or yarn links don't work for them.

I don't have a better strategy at the moment and will create a bug to discuss it later.

marechal-p:
I only publish native sources, although there's a `src/integration-test/README.md` that also end up in the mix I don't know why.

The native code has to be build on the machines that depend on the adapter, I don't see how to not publish that.

dschaefer:
> The native code has to be build on the machines that depend on the adapter, I don't see how to not publish that.

True. Still trying to get used to how that works with npm.

Though most clients I assume won't expect their users to run builds to install the adapter. At least that will be true for VS Code and Eclipse IDE. They will need to ship the binaries for all platforms.

dschaefer:
Mind you, the native libraries are so sensitive to the version of node.js you're using it's going to take a different mind set to work.

marechal-p:
> They will need to ship the binaries for all platforms.

Yes, still trying to wrap my head around that and how to properly do it. I wouldn't mind help with this, as I am not very familiar with the native world.

> Mind you, the native libraries are so sensitive to the version of node.js you're using it's going to take a different mind set to work.

Indeed, so far I thought that I used APIs simple enough that it wouldn't make a difference, hopefully. But maybe there are already caveats that I am just unaware of :/

PR #76, opened 2019-02-06
Publish binding.gyp

marechal-p:
We are not publishing the binaries built from native code, but one might
want to actually build it. For that, `binding.gyp` is required.

---

I am sorry again, it seems like it would require this file too in order to be able to build everything.
I cannot think about anything else we would be missing, doesn't mean I didn't miss something...

Conversion to N-API is following.

Note that this commit explicitly adds a dev dependency on `node-gyp`, it worked before, but I feel more comfortable having it explicitly defined?

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "chai": "^4.2.0",
     "mocha": "^5.2.0",
     "mocha-jenkins-reporter": "^0.4.1",
+    "node-gyp": "^3.8.0",
     "ts-loader": "^5.3.0",
     "tslint": "^5.11.0",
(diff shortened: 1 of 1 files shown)


marechal-p:
I don't even know if it would be possible to test such deployments, in order to catch any failures like I could experience. Basically the packed vscode+adapter unzipped package will behave correctly until we try to use the GDB console, at which point it will fail to load the missing `pty.node` binary.

marechal-p:
CI failure seems unrelated.

rerun tests

marechal-p:
CI failure seems unrelated.

rerun tests

Issue #75, opened 2019-02-05
Use N-API in the native pty code.


dschaefer:
N-API should hopefully give us the ability to release the native module independent of the node version used in the client (as long as it >= 8.x).

PR #77, opened 2019-02-07
Convert native code to use N-API

marechal-p:
N-API claims to be ABI stable, meaning that a binary built using it
could be reused accross different Node.js versions.

Fixes: #75

### DIFF
--- a/src/native/pty.cc
+++ b/src/native/pty.cc
@@ -7,42 +7,46 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *********************************************************************/
-#include <node.h>
+#include "napi.h"
 
 #ifndef WINDOWS
 #include "scoped_fd.h"
-#include <termios.h>
 #include <cstdlib>
 #include <cstring>
 #include <fcntl.h>
+#include <termios.h>
 #endif
 
-static void _throw_exc(v8::Isolate *isolate, const char *message) {
-  isolate->ThrowException(
-      v8::Exception::Error(v8::String::NewFromUtf8(isolate, message)));
+static void _throw_exc(const Napi::Env env, const char *message) {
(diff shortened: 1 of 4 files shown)


simark:
I was looking over Paul's shoulder when he did this, so it LGTM.  It might be good (as another PR perhaps) to add a bit of comments to explain the less trivial bits, so newcomers understand why we do this and that.

marechal-p:
Added comments, thank you again Simon.

src/native/pty.cc
```diff
@@ -7,42 +7,46 @@
...
-static void create_pty(const v8::FunctionCallbackInfo<v8::Value> &args) {
-  v8::Isolate *isolate = args.GetIsolate();
+static Napi::Value create_pty(const Napi::CallbackInfo &info) {
+  Napi::Env env = info.Env();
 #ifdef WINDOWS
-  return _throw_exc(isolate, ".create_pty() is not supported on Windows");
+  // Windows does not supports TTYs.
+  _throw_exc(env, ".create_pty() is not supported on Windows");
 #else
+  // The fd will be closed on this scope exit.
```
simark: ... in the error cases.

marechal-p:
Seems good to go, thanks.

dschaefer:
Looks like tomorrow I'll finally have time to try this all out in vscode.

marechal-p:
Left is to find a way to properly pack the extension for vscode, distributing the binaries for every platforms. Running VS Code from the `cdt-gdb-vscode repo` worked the last time I tried, but I don't think it can be distributed as is for everyone... At least the GDB console feature will fail on the other systems for a given build.

marechal-p:
As a npm dependency it is fine though, as it will build for your platform on install.

PR #79, opened 2019-02-08
Reset version back to 0.0.0 so we can start publishing.

dschaefer:


dschaefer:
Rushing this through so I can do some testing. No objections from the Gitter channel.

PR #81, opened 2019-02-11
Fix deps

dschaefer:
I also tagged version 0.0.2 to do an initial publish from my machine.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
   },
   "homepage": "https://github.com/eclipse-cdt/cdt-gdb-adapter#readme",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


dschaefer:
rerun tests. These are pretty flaky lately.

simark:
> rerun tests. These are pretty flaky lately.

I haven't seen the failure, so not sure if it is related, but do you still plan to fix the race condition in var.spec.ts we discussed on gitter?  I don't know how to link directly to a gitter discussion, but that discussion was on Jan 17th.

dschaefer:
I think it failed 3 times today but it was a different subset of the tests each time.

PR #84, opened 2019-02-11
Add NAPI_CPP_EXCEPTIONS. Fixed configs.

dschaefer:
Windows compile failed because we didn't tell it what exception
handling to use. Defining NAP_CPP_EXCEPTOINS.

Then when doing that, the proper /EHsc flag wasn't turned on.
The msvs_settings need to be in a configuration.

### DIFF
--- a/binding.gyp
+++ b/binding.gyp
@@ -4,7 +4,7 @@
             'target_name': 'pty',
             'sources': ['src/native/pty.cc'],
             'conditions': [
-                ['OS=="win"', { 'defines': ['WINDOWS'] }],
+                ['OS=="win"', { 'defines': ['WINDOWS', 'NAPI_CPP_EXCEPTIONS'] }],
             ],
 
             # https://github.com/nodejs/node/blob/master/doc/api/n-api.md#n-api-version-matrix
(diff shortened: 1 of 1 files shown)


binding.gyp
```diff
@@ -4,7 +4,7 @@
             'target_name': 'pty',
             'sources': ['src/native/pty.cc'],
             'conditions': [
-                ['OS=="win"', { 'defines': ['WINDOWS'] }],
+                ['OS=="win"', { 'defines': ['WINDOWS', 'NAPI_CPP_EXCEPTIONS'] }],
```
marechal-p: This will only be defined if we are on Windows, is it intended?

marechal-p: Ah, maybe it is, didn't read the comment... Ok, apparently only Windows needs that.

dschaefer: Not really sure if Linux and Mac need it. I get a #error in the napi header file on Windows without it.

dschaefer: We should look deeper in to that and see what it's actually doing. Maybe it's just a Windows thing to deal with their really weird exception handling.

PR #89, opened 2019-03-11
Add offset member to memory request

marechal-p:
This allows clients to easily move around some location that would be
resolved by GDB.

Because the offset is in bytes, it avoids issues such as doing `&i +
offset` were in this case pointer arithmetics would be involved, making
jumps of several bytes at once according to the pointer type.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -58,6 +58,7 @@ export type VariableReference = FrameVariableReference | ObjectVariableReference
 export interface MemoryRequestArguments {
     address: string;
     length: number;
+    offset?: number;
 }
 
 /**
@@ -570,9 +571,14 @@ export class GDBDebugSession extends LoggingDebugSession {
                 throw new Error(`Invalid type for 'length', expected number, got ${typeof (args.length)}`);
             }
 
+            if (typeof (args.offset) !== 'number' && typeof (args.offset) !== 'undefined') {
+                throw new Error(`Invalid type for 'offset', expected number or undefined, got ${typeof (args.offset)}`);
+            }
+
             const typedArgs = args as MemoryRequestArguments;
 
-            const result = await sendDataReadMemoryBytes(this.gdb, typedArgs.address, typedArgs.length);
+            const result = await sendDataReadMemoryBytes(
+                this.gdb, typedArgs.address, typedArgs.length, typedArgs.offset);
             response.body = {
                 data: result.memory[0].contents,
                 address: result.memory[0].begin,
--- a/src/integration-tests/mem.spec.ts
+++ b/src/integration-tests/mem.spec.ts
@@ -59,7 +59,6 @@ function verifyMemoryReadResult(resp: MemoryResponse, expectedBytes: string, exp
 
     const actualAddress = parseInt(resp.body.address, 16);
     expect(actualAddress).eq(expectedAddress);
-
 }
 
 describe('Memory Test Suite', function() {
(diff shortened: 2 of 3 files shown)


marechal-p:
Ah maybe I could add tests...

simark:
> Ah maybe I could add tests...

Yes please :)

marechal-p:
I am trying to add a new field to the response sent to the clients, notably the current offset.

But I don't understand why we send the hexa values rather than numbers directly?

https://github.com/eclipse-cdt/cdt-gdb-adapter/blob/e450b6af576ec27169fbc2f7622f282213ff7719/src/GDBDebugSession.ts#L69

marechal-p:
I mean it is not a big deal, just wondering if one is better than the other.

dschaefer:
I'm not really sure, I was surprised to see it as well when I built the Memory View. It works though, so I left it at that.

simark:
Because we want to support properly 64-bit addresses, even on runtimes that do not support 64 bits integers.  In node, for example, address 0xfafafafafafafafa would be represented like this:

```
> JSON.parse('{"addr": 18085043209519168250}')
{ addr: 18085043209519167000 }
```
So to avoid that, we send it as a string.  The application can then convert it to a number if it wants to (using the long library in JS, for example).  And since we are sending it as a string, we might as well sending in hex since it's more meaningful to any human who happens to read the DAP logs.

marechal-p:
@simark this is a perfect reason for me, thanks!

marechal-p:
Hmm, gdb question: Are the returned hex values from gdb dependent on the host architecture?

For instance, the offset return is something like `0x0000000000000000` on my machine, would it be different on a 32 bit OS/host?

dschaefer:
And for what it's worth, the gdb remote protocol also sends memory in that format, having written a gdb stub recently. :)

simark:
> For instance, the offset returned is something like `0x0000000000000000` on my machine, would it be different on a 32 bit OS/host?

It will probably just contain less zeroes, but apart from that, nothing special.  It is very likely that you can forward what we receive from GDB directly to the client.

marechal-p:
@simark had the issue in the tests, but just parsing that into an int will fix it.

marechal-p:
```ts
        // Test positive offset
        let mem = (await dc.send('cdt-gdb-adapter/Memory', {
            address: '&array',
            length: 5,
            offset: 5,
        })) as MemoryResponse;

        verifyMemoryReadResult(mem, '48450c2d13', addrOfArray + 5);
        expect(parseInt(mem.body.offset, 16)).eq(5);
```
Somehow I get the correct data but the offset returned by the response is 0, not 5 like I would expect...

src/GDBDebugSession.ts
```diff
@@ -570,12 +572,18 @@ export class GDBDebugSession extends LoggingDebugSession {
...
 
+            if (typeof (args.offset) !== 'number' && typeof (args.offset) !== 'undefined') {
+                throw new Error(`Invalid type for 'offset', expected number or undefined, got ${typeof (args.offset)}`);
+            }
+
             const typedArgs = args as MemoryRequestArguments;
 
-            const result = await sendDataReadMemoryBytes(this.gdb, typedArgs.address, typedArgs.length);
+            const result = await sendDataReadMemoryBytes(
+                this.gdb, typedArgs.address, typedArgs.length, typedArgs.offset);
```
simark: Isn't `typedArgs.offset` possibly undefined, while `sendDataReadMemoryBytes`'s last argument can't be undefined?  How come does Typescript accept it?

simark: Ah, when the argument is undefined, it takes the default value in sendDataReadMemory Bytes, I suppose.

marechal-p: You can confirm this easily:
```
$ node
> function test(a=2) { console.log(a) }
> test()
2
> test(1)
1
> test(undefined)
2
```

marechal-p:
> Somehow I get the correct data but the offset returned by the response is 0, not 5 like I would expect...

So, as it turns out, gdb can return several bits of memory for one request, this to accommodate with unreadable memory parts, among other things I guess.

So this means that the first bit of memory returned will most likely start exactly at the requested address (request.address + request.offset), and only subsequent memory parts will start to see their offset incremented.

This also means that the current message won't be able to transport such information, but that can be addressed later. It will be breaking though.

For now I'll fix the tests.

simark:
FWIW, this LGTM.  I would maybe prefer not expose the "offset" in the response until we process the partial responses correctly (support multiple blocks of memory in the response), since I think it will just confuse people more than anything, but it's not a big deal.

marechal-p:
@simark will remove the extra field from the response :)

dschaefer:
Are we good to go?

simark:
I think @marechal-p intends to remove the offset field from the response, which is not done yet.

dschaefer:
Oh, right. OK.

marechal-p:
> Are we good to go?

Yes, finally :)

simark:
Thanks, LGTM!

dschaefer:
Awesome, thanks!

dschaefer:
I have kicked the build (have to do that manually until I figure out how to keep it from building empty change sets). v0.0.12 is now up on npm.

PR #91, opened 2019-03-19
Add handling for stopped by signal.

dschaefer:
Use the signal name as the stop reason if it's available. Doesn't seem to be any semantics behind that name. In VS Code, it shows up as the stop reason in the Call Stack view (e.g., PAUSED ON SIGSEGV)

### DIFF
--- a/tslint.json
+++ b/tslint.json
@@ -13,6 +13,7 @@
     "interface-name": false,
     "no-empty-interface": false,
     "object-literal-sort-keys": false,
+    "ordered-imports": false,
     "quotemark": [
       true,
       "single",
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,7 +5,7 @@
       "type": "node",
       "request": "launch",
       "name": "Mocha All",
-      "program": "${workspaceFolder}/node_modules/mocha/bin/_mocha",
+      "program": "${workspaceFolder}/node_modules/.bin/mocha",
       "args": [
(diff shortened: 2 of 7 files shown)


dschaefer:
I'm going to add a test first. Stand by.

simark:
> Use the signal name as the stop reason if it's available. Doesn't seem to be any semantics behind that name. In VS Code, it shows up as the stop reason in the Call Stack view (e.g., PAUSED ON SIGSEGV)

That sounds good.

> I'm going to add a test first. Stand by.

:heart:

tslint.json
```diff
@@ -13,6 +13,7 @@
     "interface-name": false,
     "no-empty-interface": false,
     "object-literal-sort-keys": false,
+    "ordered-imports": false,
```
marechal-p: I kinda liked when the imports were ordered, did it cause issues?

simark: > I kinda liked it when the imports were ordered, did it cause issues?

Yeah, me too, this way I don't have to decide where to put them :).  If they are not ordered, there's usually a fixit to make them ordered, so it's quite easy.

dschaefer: Yeah, I prefer to group my imports based on where they come from, node's modules, node_modules, local. And since with a modern IDE like vscode, you never really need to go edit them other than to remove unused one, I found I was wasting a lot of time just to create a list I wasn't happy with. Personal preference I guess.

Issue #26, opened 2018-11-27
Top level variable element in Variables View should display pointer address for arrays


jcrwilliams:
Currently the size of the array is displayed in square brackets. The basis for the correct display value here is Eclipse CDT and Microsoft's own C/C++ Tools for VS Code which display the pointer value.

jonahgraham:
This was completed in #94

PR #94, opened 2019-04-29
Display address of array entry in Variables View

jcrwilliams:
Issue #26 - Show the address of the array rather than the size in square
brackets, as is done in Eclipse CDT.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -27,6 +27,10 @@ export interface MIGDBShowResponse extends MIResponse {
     value?: string;
 }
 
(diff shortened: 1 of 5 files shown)


dschaefer:
run tests

dschaefer:
Turns out we haven't had builds run since we moved the CI machine to the new cluster. Working on that now.

dschaefer:
run tests

dschaefer:
run tests

dschaefer:
Doesn't look like there's a way to clear the failed run of the old build job. New one (default) passed. Going with that :).

PR #95, opened 2019-05-22
Add handling of thread-created notify messages.

dschaefer:
After configurationDone, vscode sends us a threads request.
However, since we're running, our request to get the threads
info from gdb doesn't respond. That's all fine if we have a
breakpoint since the threads request will resolve itself when
the breakpoint is hit. If not, the session hangs.

This change adds handling for notify messages and thread
creation in particular. We store away the threads from the
notifies and return that if we get a threads request while
running.

Also clean up the API to separate out the result/async class
from the data.

Also added a Server launch so we can debug the adapter.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -4,6 +4,19 @@
     {
       "type": "node",
       "request": "launch",
+      "name": "Server",
+      "cwd": "${workspaceFolder}",
+      "program": "${workspaceFolder}/dist/debugAdapter.js",
+      "args": [
+        "--server=4711"
+      ],
+      "outFiles": [
+        "${workspaceFolder}/dist/**/*.js"
+      ]
+    },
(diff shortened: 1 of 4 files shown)


jonahgraham:
Hi @dschaefer, I haven't looked at this in depth, but it looks like you may be affected by something particular in the debug protocol. Short story is that it is not a simple remote function call implementation because you can have two "function calls" from the client at the same time. See https://github.com/Microsoft/vscode/issues/4902#issuecomment-368583522 for a picture. Let me know if you want me to dig in deeper and look at this more.

dschaefer:
Hey @jonahgraham. I think we're OK, here. My main objective was to get the pause request to come out. Apparently it won't do it without threads being returned on the thread request. Now that we do, I have another incoming PR :).

PR #96, opened 2019-05-22
Get pause and setBreakpoint while running working.

dschaefer:
Still not super clear this is the right approach but it works in
the small use case I tried. Pause request simply calls
kill('SIGINT")  on the child process. Works on Linux at least.

Also calls pause on the setBreakpoints request and waits for
the stopped async message. Then sets the breakpoints and
continues.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -7,7 +7,7 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *********************************************************************/
-import { execFile, spawn } from 'child_process';
+import { execFile, spawn, ChildProcess } from 'child_process';
 import * as events from 'events';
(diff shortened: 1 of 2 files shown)


dschaefer:
rerun tests

dschaefer:
Tested with 7.12 on Ubuntu 18.04 where this works. Also tested against lldb-mi on Mac and it works there. And finally it also works with QNX's gdb which is based off of 7.12 as well and works.

It does not work on Windows as expected. We'll need to borrow the spawner/starter code from CDT to set up the messaging appropriately.

Also tested with 7.9.1 and it does not work there. There must be better process handling in 7.12 and up that makes things work properly. Will open a new bug for 7.9.1 and move on.

PR #97, opened 2019-05-23
Get pty.cc building on Mac.

dschaefer:
xcode_settings seemed to be at the wrong level.

Also strerror_r has a different API on Mac than Linux.

### DIFF
--- a/src/native/pty.cc
+++ b/src/native/pty.cc
@@ -31,8 +31,18 @@ static void _throw_exc_format(const Napi::Env env, int error,
   const int ERRMSG_MAX_SIZE = 128;
   char errmsg_buffer[ERRMSG_MAX_SIZE];
   char message[ERRMSG_MAX_SIZE];
+#ifdef _GNU_SOURCE
   char *errmsg = strerror_r(error, errmsg_buffer, ERRMSG_MAX_SIZE);
   snprintf(message, ERRMSG_MAX_SIZE, "%s: %s", function_name, errmsg);
+#else
+  int rc = strerror_r(error, errmsg_buffer, ERRMSG_MAX_SIZE);
+  if (rc) {
+    snprintf(message, ERRMSG_MAX_SIZE, "%s", function_name);
+  } else {
+    snprintf(message, ERRMSG_MAX_SIZE, "%s: %s", function_name, errmsg_buffer);
+  }
+#endif
+
   _throw_exc(env, message);
 }
 #endif
--- a/binding.gyp
+++ b/binding.gyp
@@ -19,15 +19,15 @@
         # Use C++ exceptions:
         'cflags!': [ '-fno-exceptions' ],
         'cflags_cc!': [ '-fno-exceptions' ],
-        'xcode_settings': {
-            'GCC_ENABLE_CPP_EXCEPTIONS': 'YES',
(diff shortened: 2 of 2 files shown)


src/native/pty.cc
```diff
@@ -31,8 +31,8 @@ static void _throw_exc_format(const Napi::Env env, int error,
   const int ERRMSG_MAX_SIZE = 128;
   char errmsg_buffer[ERRMSG_MAX_SIZE];
   char message[ERRMSG_MAX_SIZE];
-  char *errmsg = strerror_r(error, errmsg_buffer, ERRMSG_MAX_SIZE);
-  snprintf(message, ERRMSG_MAX_SIZE, "%s: %s", function_name, errmsg);
+  strerror_r(error, errmsg_buffer, ERRMSG_MAX_SIZE);
```
simark: This isn't totally correct for when using glibc.  From the man page:

> The GNU-specific strerror_r() returns a pointer to a string containing the error message.  This may be either a pointer to a string that the function stores in buf, or a pointer to some (immutable) static string (in which case buf is unused).

So if using the GNU variant, we might end up displaying random data from the errmsg_buffer.

I would suggest trying to use the XSI version all the time (see man page).  And to make sure that the XSI version is being used, let's check the int return value of strerror_r:

```
int ret = strerror_r(...);
if (ret != 0) {
  // abort or just use a default error message
}
```

This way, if we end up using the GNU version, the code won't compile.

dschaefer: Cool. Wasn't sure about that. The Mac does return an int, which is why it didn't compile. I'll look deeper. Thanks!

simark: The posix-defined version returns an int, so using that one makes portability easier.

dschaefer: Hmm, how was this working before? We were using strerror_r.

dschaefer: Oh, read the whole man page... :)

dschaefer: Ugh, turning off _GNU_SOURCE also turned off grantpt and friends. #ifdef will work too I guess.

simark: > Ugh, turning off _GNU_SOURCE also turned off grantpt and friends. #ifdef will work too I guess.

Aww :(.  Then yeah, ifdef.

dschaefer:
rerun tests since the build machines aren't very predictable.

dschaefer:
rerun tests again. Making sure we get the same failures two runs in a row.

dschaefer:
rerun tests. Maybe today's a better day ;p

dschaefer:
It is :). sigh...

thegecko:
I'm having problems building this on Mac even after this PR;

```bash
pty.cc:81:11: error: use of undeclared identifier 'ptsname_r'; did you mean 'ttyname_r'?
```

Swapping out to use `ttyname_r` seemed to fix the issue.

dschaefer:
> I'm having problems building this on Mac even after this PR;
> 
> ```shell
> pty.cc:81:11: error: use of undeclared identifier 'ptsname_r'; did you mean 'ttyname_r'?
> ```
> 
> Swapping out to use `ttyname_r` seemed to fix the issue.

Do you want to create a separate issue for this? I don't think we've tried mac yet.

thegecko:
Ok

PR #102, opened 2019-06-13
Update README.md

marechal-p:
We are now using `yarn` and not `npm`.


dschaefer:
It's just a Readme change so I'll just push it.

PR #103, opened 2019-06-25
Fixed yarn build on MacOS

thegecko:
With reference to a comment in https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/97#issuecomment-500899450, I got `pty.cc` to build on MacOS by replacing `ptsname_r` with `ttyname_r`.

This is untested on Linux or Windows, so may require further conditional statements.

### DIFF
--- a/src/native/pty.cc
+++ b/src/native/pty.cc
@@ -78,9 +78,9 @@ static Napi::Value create_pty(const Napi::CallbackInfo &info) {
     _throw_exc_format(env, error, "tcsetattr");
 
   // see: man ptmx
-  error = ptsname_r(master_fd.get(), slave_name, SLAVE_NAME_MAX_SIZE);
+  error = ttyname_r(master_fd.get(), slave_name, SLAVE_NAME_MAX_SIZE);
   if (error)
-    _throw_exc_format(env, error, "ptsname_r");
+    _throw_exc_format(env, error, "ttyname_r");
   error = grantpt(master_fd.get());
   if (error)
     _throw_exc_format(env, error, "grantpt");


dschaefer:
run tests

marechal-p:
That's what I was going to say, `ttyname_r` does not work because I think it is part of the old API to open ttys, when we use the `ptmx` mechanism.

marechal-p:
Tests are failing on Linux now.

thegecko:
> Tests are failing on Linux now

Not surprising. I don't have a way of testing this on linux. Is there a #DEF to determine MacOS or linux which can be used?

marechal-p:
See:
- https://github.com/eclipse-cdt/cdt-gdb-adapter/blob/master/binding.gyp#L7
- https://github.com/eclipse-cdt/cdt-gdb-adapter/blob/master/src/native/pty.cc#L52

I simply did not expect it to fail on Mac. Maybe change the logic to stop if not Linux?

marechal-p:
I tried to fix issues on Mac after your comment about it last time, I managed to get it working but I had issues with the tests, it would hang for some reason.

https://github.com/marechal-p/cdt-gdb-adapter/tree/mp/pty-refactor

I was desperately trying to find the cause of the hang, all I could see is that a read call is keeping the event loop alive, despite the associated fd being closed.

https://github.com/marechal-p/cdt-gdb-adapter/blob/mp/pty-refactor/src/native/pty.spec.ts#L77-L79

```
There are 4 handle(s) keeping the process running

# Timeout
/home/emaapur/projects/cdt-gdb-adapter/dist/native/pty.spec.js:72                      - setTimeout(() => {
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/lib/interfaces/common.js:112 - opts.fn.call(suite);
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/lib/interfaces/bdd.js:40     - return common.suite.create({
/home/emaapur/projects/cdt-gdb-adapter/dist/native/pty.spec.js:40                      - describe('pty creation', function () {
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/lib/mocha.js:250             - suite.emit('require', require(file), file, self);
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/lib/mocha.js:247             - this.files.forEach(function(file) {
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/lib/mocha.js:576             - this.loadFiles();
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/bin/_mocha:637               - runner = mocha.run(program.exit ? exit : exitLater);

# SIGNALWRAP
/home/emaapur/projects/cdt-gdb-adapter/node_modules/mocha/bin/_mocha:640 - process.on('SIGINT', () => {

# FSREQWRAP
/home/emaapur/projects/cdt-gdb-adapter/dist/fileDuplex.js:65 - fs.read(this.fd, this.buffer, 0, Math.min(this.buffer.length, size), null, (error, bytesRead, readBuffer) => {

# FSREQWRAP
/home/emaapur/projects/cdt-gdb-adapter/dist/fileDuplex.js:65 - fs.read(this.fd, this.buffer, 0, Math.min(this.buffer.length, size), null, (error, bytesRead, readBuffer) => {
```

marechal-p:
Ah maybe I went a bit ahead of myself here: The branch I linked is only part of the refactoring required to get Mac working, it will not just work with it. But it doesn't change what I said in https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/103#issuecomment-505564937: Maybe just disable the feature if not Linux.

thegecko:
> Tests are failing on Linux now.

Where can I see test failures so I can at least try things out?

Is this also likely to fail on Windows based on https://github.com/eclipse-cdt/cdt-gdb-adapter/blob/master/src/native/pty.cc#L52

I'm concerned that supporting Linux only for a relatively niche feature is a large blocker to using this adapter.

dschaefer:
rerun tests

The link to the failing build should show up in the checks.

dschaefer:
In general, the verify job is here: https://ci.eclipse.org/cdt/view/npm%20builds/job/cdt-gdb-adapter-verify/

Looks like it passed the last run.

marechal-p:
> Looks like it passed the last run.

Did it? I see a failure...

marechal-p:
> I'm concerned that supporting Linux only for a relatively niche feature is a large blocker to using this adapter.

https://github.com/eclipse-cdt/cdt-gdb-adapter/blob/master/src/GDBDebugSession.ts#L121

We do check before hand, again: I just assumed that it would work on Mac without doing anything else but it seems it is not the case, so as I said we can make the checks look for Linux only.

thegecko:
Closed in favour of #115

PR #104, opened 2019-06-25
Make Native Pty optional

thegecko:
With our usage of this adapter, we have no need for `spawnInClientTerminal`.

However, there is a hard dependency for this on native `pty` code.

This makes the package difficult to distribute in built form as it has to be built for every target platform (Linux, Windows, MacOS, etc.).

This PR switches `Pty` to be a [dynamic import](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-4.html) so it's only required when needed by `spawnInClientTerminal`.

This allows the adapter to be distributed as pure JavaScript as long as `spawnInClientTerminal` isn't used.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -14,7 +14,6 @@ import { logger } from 'vscode-debugadapter/lib/logger';
 import { AttachRequestArguments, LaunchRequestArguments } from './GDBDebugSession';
 import { MIResponse } from './mi';
 import { MIParser } from './MIParser';
-import { Pty } from './native/pty';
 
 export interface MIExecNextRequest {
     reverse?: boolean;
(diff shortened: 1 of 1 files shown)


src/GDBBackend.ts
```diff
@@ -51,6 +50,7 @@ export class GDBBackend extends events.EventEmitter {
     public async spawnInClientTerminal(args: LaunchRequestArguments | AttachRequestArguments,
         cb: (args: string[]) => Promise<void>) {
         const gdb = args.gdb ? args.gdb : 'gdb';
+        const Pty = await require('./native/pty');
```
simark: Can you please add a comment saying why it's done this way?

thegecko: done

marechal-p:
```ts
const { Pty } = require('./native/pty');
```
I don't understand your notation, but I see why you want it optional.

thegecko: What don't you understand?

marechal-p: Why the await, and why not extracting the `Pty` class like done in the original import?

thegecko: I've changed this to use the TypeScript syntax.

thegecko: Await is needed to wait for the import to complete.

dschaefer:
run tests.

dschaefer:
Sold :)

PR #105, opened 2019-06-25
Add default vscode build task

thegecko:
Simple addition to allow constant building in VSCode


dschaefer:
No need to test this since it's not a code change.

PR #106, opened 2019-06-25
Add request option for gdb client arguments

thegecko:
This PR adds support for passing optional arguments to the gdb client launch command.

There are a few other arguments I plan on adding, so will update the [adapter config](https://github.com/eclipse-cdt/cdt-gdb-vscode/blob/master/package.json#L47) in one go once any changes are merged here.

The PR also refactors the request argument interfaces slightly so common arguments are shared for easier maintenance.

PR #107, opened 2019-06-25
Add optional initialisation commands

thegecko:
Building on #106, this PR allows the user to pass optional initialisation commands to be run by gdb once ready.

This allows immediate control/setup commands such as target download `-target-download` or run-to-main `-break-insert -t --function main`

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -139,6 +139,7 @@ export class GDBDebugSession extends LoggingDebugSession {
 
             await this.spawn(args);
             await this.gdb.sendFileExecAndSymbols(args.program);
+            await this.gdb.sendEnablePrettyPrint();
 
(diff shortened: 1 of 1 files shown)


dschaefer:
run tests

src/GDBDebugSession.ts
```diff
@@ -164,7 +169,13 @@ export class GDBDebugSession extends LoggingDebugSession {
             await this.spawn(args);
             await this.gdb.sendFileExecAndSymbols(args.program);
 
-            this.gdb.sendEnablePrettyPrint();
+            await this.gdb.sendEnablePrettyPrint();
```
dschaefer: Should we do this on the attach request too?

thegecko: Added

PR #109, opened 2019-06-25
Add support for attaching to a remote gdb server

thegecko:
This PR introduces an attach argument `remote` which allows the gdb client to attach to a remote gdb server.

e.g.:

```
attachArguments = {
    ...
    remote: localhost:3456,
    ...
}
```

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -126,6 +126,10 @@ export class GDBBackend extends events.EventEmitter {
         return this.sendCommand('-gdb-exit');
     }
 
+    public sendTargetSelectRemote(remote: string) {
+        return this.sendCommand(`-target-select remote ${remote}`);
+    }
+
     private nextToken() {
         return this.token++;
     }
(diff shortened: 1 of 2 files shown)


dschaefer:
No, only I can say run tests :).

thegecko:
> No, only I can say run tests :).

Darn, thought I'd save you some time!

dschaefer:
Yeah, it's some weird security thing I think where only admins can kick off the verify jobs.

dschaefer:
I'm just wondering about this though. Is it not possible to launch when connected to a remote target. I know for embedded system connecting to hardware debuggers or gdb stubs you attach, but I'm not sure that's generally true especially when considering all the bells and whistles they've added to the Linux gdbserver.

thegecko:
> Is it not possible to launch when connected to a remote target

This is possible, but I would then expect the adapter to handle the gdb server process launching and arguments. I felt that would be unwanted, but happy to implement it in this way instead or as well as attach.

I believe the launch arguments for this case would then look like:

- gdb client executable
- gdb client args
- gdb server executable
- gdb server args
- executable being debugged
- executable arguments

Would the port for connecting client and server then be determined by the adapter in a similar way to [this one][(https://github.com/Marus/cortex-debug)?

thegecko:
As discussed in https://github.com/eclipse-cdt/cdt-gdb-adapter/issues/114 this is better implemented in a subclass

PR #110, opened 2019-06-25
Add thread ID to step commands

thegecko:
Step commands optionally support threadIDs. This PR adds this support.

### DIFF
--- a/src/mi/exec.ts
+++ b/src/mi/exec.ts
@@ -20,18 +20,34 @@ export function sendExecRun(gdb: GDBBackend) {
     return gdb.sendCommand('-exec-run');
 }
 
-export function sendExecContinue(gdb: GDBBackend) {
-    return gdb.sendCommand('-exec-continue');
+export function sendExecContinue(gdb: GDBBackend, threadId?: number) {
+    let command = '-exec-continue';
+    if (threadId) {
+        command += ` --thread ${threadId}`;
+    }
+    return gdb.sendCommand(command);
 }
 
-export function sendExecNext(gdb: GDBBackend) {
-    return gdb.sendCommand('-exec-next');
+export function sendExecNext(gdb: GDBBackend, threadId?: number) {
+    let command = '-exec-next';
+    if (threadId) {
+        command += ` --thread ${threadId}`;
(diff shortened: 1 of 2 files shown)


thegecko:
Can this be merged?

src/mi/exec.ts
```diff
@@ -20,18 +20,34 @@ export function sendExecRun(gdb: GDBBackend) {
...
+    }
+    return gdb.sendCommand(command);
 }
 
-export function sendExecNext(gdb: GDBBackend) {
-    return gdb.sendCommand('-exec-next');
+export function sendExecNext(gdb: GDBBackend, threadId?: number) {
+    let command = '-exec-next';
+    if (threadId) {
+        command += ` --thread ${threadId}`;
```
marechal-p: According to [GDB MI docs](https://sourceware.org/gdb/download/onlinedocs/gdb/GDB_002fMI-Program-Execution.html), `-exec-next` doesn't support a `--thread` argument, unless I am mistaken?

Did you try it/is it actually working?

thegecko: Yes, it's implemented in the adapter we use:

https://github.com/Marus/cortex-debug/blob/12bc8c0c4f17b11b1fb22e218b45732c521863bb/src/backend/mi2/mi2.ts#L274

dschaefer: I guess the real question is does this work on all gdb's? Again, we need tests for this so we can prove that one way or another.

simark: > According to [GDB MI docs](https://sourceware.org/gdb/download/onlinedocs/gdb/GDB_002fMI-Program-Execution.html), `-exec-next` doesn't support a `--thread` argument, unless I am mistaken?

All commands support it, it's documented once for all commands here:

https://sourceware.org/gdb/download/onlinedocs/gdb/Context-management.html#Context-management

marechal-p: Thank you for the link!

thegecko:
bump

dschaefer:
Sorry, I've been on vacation this week. Will get to them all on Monday.

Issue #28, opened 2018-11-27
Use GDB's mi-async mode


simark:
I think that we should use GDB's `set mi-async on` mode (which in the past was `set target-async on`) from the start.  In this mode, when you tell the target to resume, you get a `^done` response, then you can issue other commands while the target is running (though not all commands work while the target is running).

In CDT, I think that mi-async is used in non-stop mode, but not in all-stop mode.  It would simplify things if we could use it all the time here.

One caveat is that it might not be available when debugging Windows natively (not because it's not possible, just because nobody every improved GDB to support it).

simark:
There is an attempt of this: #116

PR #111, opened 2019-06-25
Interrupt execution on pause

thegecko:
This change could prove controversial.

When pausing execution of the adapter, it's set up to kill the process entirely (it's not clear why).

This PR changes that to correctly(?) send an interrupt request instead.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -57,15 +57,6 @@ export class GDBBackend extends events.EventEmitter {
         return this.parser.parse(pty.master);
     }
 
-    public pause() {
-        if (this.proc) {
-            this.proc.kill('SIGINT');
-            return true;
-        } else {
-            return false;
-        }
-    }
-
     public async supportsNewUi(gdbPath?: string): Promise<boolean> {
         const gdb = gdbPath || 'gdb';
         return new Promise<boolean>((resolve, reject) => {
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -221,7 +221,7 @@ export class GDBDebugSession extends LoggingDebugSession {
             const waitPromise = new Promise<void>((resolve) => {
                 this.waitPaused = resolve;
             });
-            this.gdb.pause();
+            await mi.sendExecInterrupt(this.gdb);
             await waitPromise;
         }
 
@@ -402,10 +402,12 @@ export class GDBDebugSession extends LoggingDebugSession {
 
     protected async pauseRequest(response: DebugProtocol.PauseResponse,
         args: DebugProtocol.PauseArguments): Promise<void> {
-        if (!this.gdb.pause()) {
-            response.success = false;
+        try {
+            await mi.sendExecInterrupt(this.gdb, args.threadId);
+            this.sendResponse(response);
+        } catch (err) {
+            this.sendErrorResponse(response, 1, err.message);
         }
-        this.sendResponse(response);
     }
 
     protected scopesRequest(response: DebugProtocol.ScopesResponse,
--- a/src/mi/exec.ts
+++ b/src/mi/exec.ts
@@ -24,6 +24,14 @@ export function sendExecContinue(gdb: GDBBackend) {
     return gdb.sendCommand('-exec-continue');
 }
 
+export function sendExecInterrupt(gdb: GDBBackend, threadId?: number) {
+    let command = '-exec-interrupt';
+    if (threadId) {
+        command += ` --thread ${threadId}`;
+    }
+    return gdb.sendCommand(command);
+}
+
 export function sendExecNext(gdb: GDBBackend) {
     return gdb.sendCommand('-exec-next');
 }


dschaefer:
It doesn't kill the process, it sends it a SIGINT. To get gdb to listen to the stdin again.

Make sure whatever you do here works before and after gdb 7.12 when they changed how all this works.

thegecko:
> To get gdb to listen to the stdin again

When trying this it simply killed gdb. Is the SIGINT approach new since gdb 7.12?

dschaefer:
> > To get gdb to listen to the stdin again
> 
> When trying this it simply killed gdb. Is the SIGINT approach new since gdb 7.12?

Yes, I think that's how it went. We were still trying to figure out how to start gdb so SIGINT didn't kill it but haven't yet. There should be clues in how the Eclipse plug-ins do this.

jonahgraham:
@thegecko are you on Windows. IIRC Windows needs a Ctrl-C sent instead of SIGINT.

However, it would be much better if we can use `-exec-interrupt`. @dschaefer do you need to run in sync mode anywhere, or can we use `-gdb-set target-async on` always.

marechal-p:
cc @simark, is there anything to watch out for when using async mode?

simark:
Well, except for the fact that `mi-async on` (which is the "new" name of `target-async`) works differently than `mi-async off`, I don't really know about things to watch out for.

The main problem is that GDB debugging natively on Windows does not support `mi-async on`, because the way the GDB Windows backend is written is not async-friendly.  It's not an inherent limitation of Windows, it's just that nobody has done the work to make it async-friendly yet.

I had opened #28 to track this.

marechal-p:
> [...] the fact that `mi-async on` (which is the "new" name of `target-async`) works differently than `mi-async off`, [...]

This was my primary concern: Does that drastically change the way the adapter should handle things?

thegecko:
> @thegecko are you on Windows

No, we actually target Windows, Linux and MacOS (but I develop on Mac). The issue here is probably that we are executing a wrapper around gdb in order to manage gdb server, however `SIGINT` still feels like a brutal approach.

dschaefer:
> No, we actually target Windows, Linux and MacOS (but I develop on Mac). The issue here is probably that we are executing a wrapper around gdb in order to manage gdb server, however `SIGINT` still feels like a brutal approach.

Well, when users are using gdb from the command line, certainly in the past anyway, they would hit Control-C to break the inferior and get the gdb prompt. That's a SIGINT, of course, or in the Windows case a Control-C event sent to the console where gdb registers a handler.

marechal-p:
> That's a SIGINT, of course, or in the Windows case a Control-C event sent to the console where gdb registers a handler.

But then what does gdb hooks onto when running as a subprocess like what is done in the adapter?

dschaefer:
> > That's a SIGINT, of course, or in the Windows case a Control-C event sent to the console where gdb registers a handler.
> 
> But then what does gdb hooks onto when running as a subprocess like what is done in the adapter?

That's where we need to go look at the CDT Eclipse plug-ins. Pause works well for all gdb's I've tried with it. I think gdb needs to start in a new process group, or maybe a new session, or something like that. The signal gets sent to the top process in the group which should be the gdb process.

Of course there's a lot more magic to get it working on Windows. I wrote that probably 10 years ago now so my memory is foggy.

thegecko:
FWIW I know the `-exec-interrupt` approach works on all platforms.

Full disclosure: we have all this working with a different adapter, but are hoping to switch to this one if we can get feature parity.

marechal-p:
I wrote a test on my end using your PR, and it seems to be hanging when sending a `pauseRequest`:

Test added to `stop.spec.ts`:

```ts
    it('can stop inferior and get gdb control back', async () => {
        await dc.launchRequest({
            verbose: true,
            gdb: gdbPath,
            program: path.join(testProgramsDir, 'loop'),
            openGdbConsole,
        } as LaunchRequestArguments);
        const { body: { threads } } = await dc.threadsRequest();
        expect(threads).lengthOf(1);
        const mainThread = threads[0];
        console.log(mainThread);
        await dc.pauseRequest({ threadId: mainThread.id });
        console.log('paused');
    });
```

`loop` executable:

```cpp
int main() {
    while (1) { }
    return 0;
}
```

Do you experience something different?

thegecko:
> Do you experience something different?

I do. In my case this works as expected and I see gdb acknowledging the request:

```
-exec-interrupt --thread 4
(gdb) 
~"\nThread "
~"4 \"rtx_idle\" received signal SIGINT, Interrupt.\n"
~"SMC_SetPowerModeWait (base=<optimized out>) at fsl_smc.c:112\n"
~"112\t    __ISB();\n"
*stopped,reason="signal-received",signal-name="SIGINT",signal-meaning="Interrupt",frame={addr="0x00004f5a",func="SMC_SetPowerModeWait",args=[{name="base",value="<optimized out>"}],file="fsl_smc.c",fullname="fsl_smc.c",line="112"},thread-id="4",stopped-threads="all"
```

I wonder if it's because I'm using remote:

https://stackoverflow.com/questions/1101887/stopping-the-inferior-process-in-gdb-without-a-signal

I need to investigate this some more.

[Other](https://github.com/Marus/cortex-debug/blob/12bc8c0c4f17b11b1fb22e218b45732c521863bb/src/backend/mi2/mi2.ts#L246) [adapters](https://github.com/WebFreak001/code-debug/blob/8219160cd7452e09d164751dcaf11a0e303a3148/src/backend/mi2/mi2.ts#L431) also seem to use `-exec-interrupt`

marechal-p:
@thegecko If you can write a test that passes it would help with troubleshooting :)

thegecko:
> If you can write a test 

That isn't going to be easy :(

marechal-p:
Then does that PR effectively do what it is meant to do? :)

If it only works using a remote debug session, would it be better to wait until your remote PR is merged, then rebase, and finally add a test demonstrating that it works in such a scenario?

thegecko:
> would it be better to wait until your remote PR is merged

I've closed the suggestion to add remote support in favour of sub-classing this adapter.

I think this change is better implemented in #116 and will close this in favour if it's going to be merged.

thegecko:
Closing in favour of #116

PR #116, opened 2019-07-04
Enable `mi-async` by default

marechal-p:
Running GDB in synchronous mode is problematic since it won't listen on the MI channel when the debuggee is running. This means that we cannot send a pause/interrupt request via MI, for instance.

This commit enables `target-async` now named `mi-async`.

Closes https://github.com/eclipse-cdt/cdt-gdb-adapter/issues/28


marechal-p:
I forgot to open this PR as a draft, but the goal here is to enable the asynchronous mode to accommodate for the change by @thegecko to add the pause request/interrupt.

Current problem: The variable setting tests are failing. It may have to do with an issue we encountered with @simark some time ago, but it wasn't in the scope of my changes back then, and now I won't remember what was the exact issue :(

src/GDBDebugSession.ts
```diff
@@ -215,15 +219,14 @@ export class GDBDebugSession extends LoggingDebugSession {
...
-        if (neededPause) {
-            // Need to pause first
-            const waitPromise = new Promise<void>((resolve) => {
-                this.waitPaused = resolve;
-            });
-            this.gdb.pause();
-            await waitPromise;
-        }
+        // const neededPause = this.isRunning;
+        // if (neededPause) {
```
marechal-p: Nice side effect of this change is that we would not need to interrupt the debuggee to get gdb listening for the MI commands.

src/integration-tests/stop.spec.ts
```diff
@@ -37,4 +38,19 @@ describe('stop', async () => {
...
+            verbose: true,
+            gdb: gdbPath,
+            program: path.join(testProgramsDir, 'loop'),
+            openGdbConsole,
+        } as LaunchRequestArguments);
+        await dc.configurationDoneRequest();
+        const { body: { threads } } = await dc.threadsRequest();
+        expect(threads).lengthOf(1);
+        const mainThread = threads[0];
+        await dc.pauseRequest({ threadId: mainThread.id });
```
marechal-p: This test now actually returns, no more timeout as encountered with https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/111.

src/mi/exec.ts
```diff
@@ -24,7 +24,18 @@ export function sendExecContinue(gdb: GDBBackend) {
...
+    let command = '-exec-interrupt';
+    if (threadId) {
+        command += ` --thread ${threadId}`;
+    }
+    return gdb.sendCommand(command);
+}
+
+export async function sendExecNext(gdb: GDBBackend, threadId?: number) {
+    if (threadId) {
+        await gdb.sendCommand(`-thread-select ${threadId}`);
```
marechal-p: Is this now required on most commands?

dschaefer:
run tests

marechal-p:
As expected it still fails on the variable tests. It is also not really consistent, at least on my machine.

src/GDBBackend.ts
```diff
@@ -59,13 +59,13 @@ export class GDBBackend extends events.EventEmitter {
...
     }
 
-    public pause() {
-        if (this.proc) {
-            this.proc.kill('SIGINT');
-            return true;
-        } else {
-            return false;
-        }
+    public async pause() {
```
thegecko: Can we drop this function?

src/GDBDebugSession.ts
```diff
@@ -138,6 +138,8 @@ export class GDBDebugSession extends LoggingDebugSession {
             this.gdb.on('notifyAsync', (resultClass, resultData) => this.handleGDBNotify(resultClass, resultData));
 
             await this.spawn(args);
+            await this.gdb.sendAsyncMode('on');
+            await this.gdb.sendGDBSet('non-stop off');
```
thegecko: What does `non-stop off` do?

src/GDBDebugSession.ts
thegecko: In our testing, remote debug of devices using `pyOCD` needs the target to stop before setting breakpoints. Could I ask we keep this in for now and I can test separately if we can remove it once this PR is merged?

src/GDBDebugSession.ts
```diff
@@ -363,7 +366,7 @@ export class GDBDebugSession extends LoggingDebugSession {
     protected async nextRequest(response: DebugProtocol.NextResponse,
         args: DebugProtocol.NextArguments): Promise<void> {
         try {
-            await mi.sendExecNext(this.gdb);
+            await mi.sendExecNext(this.gdb, args.threadId);
```
thegecko: This addition is part of #110, shall we merge that first?

src/mi/exec.ts
thegecko: Yes, please see implementation in #110 which passes `threadid` explicitly

src/GDBDebugSession.ts
thegecko: And use `await mi.sendExecInterrupt(this.gdb);`

src/GDBDebugSession.ts
marechal-p: I cannot explain it better than the [docs](https://sourceware.org/gdb/onlinedocs/gdb/Non_002dStop-Mode.html). It is already in all-stop mode by default, but it was paranoïa to get desperately get the tests to pass. It doesn't seem to help though, will remove.

src/GDBDebugSession.ts
marechal-p: Sure :)

marechal-p:
> There's a few comments and you are introducing the `loop` binary file in this PR which should be left out.

Why should I leave out `loop.c`? It is part of the tests to show that we can pause a running program.

thegecko:
> Why should I leave out loop.c?

Not the .c file, the binary:

https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/116/files#diff-d12c930aebd9c85bb7f703c0ed68197d

marechal-p:
My bad, indeed I should exclude it :grimacing:

marechal-p:
I think the tests are passing now!

src/GDBDebugSession.ts
```diff
@@ -648,9 +647,17 @@ export class GDBDebugSession extends LoggingDebugSession {
                 break;
             case 'signal-received':
                 const name = result['signal-name'] || 'signal';
-                this.sendStoppedEvent(name, parseInt(result['thread-id'], 10));
+                this.sendStoppedEvent(
+                    // When sending `-exec-interrupt`, we seem to get back
+                    // `SIGINT` as stop event. Since we need to notify the debug
+                    // client that a pause event is emitted, let's convert the
+                    // `SIGINT` signal to a pause event.
+                    name === 'SIGINT' ? 'pause' : name,
```
marechal-p: Might be controversial, but it is useful for the newly added "stop" test.

marechal-p: Any opinion on this? @dschaefer @simark

src/GDBBackend.ts
```diff
@@ -108,6 +99,13 @@ export class GDBBackend extends events.EventEmitter {
         });
     }
 
+    public sendMIAsync(state: 'on' | 'off') {
+        // http://sourceware.org/gdb/onlinedocs/gdb/Asynchronous-and-non_002dstop-modes.html
+        // Newer GDB use `mi-async` param, but legacy param is `target-async`. The two params
+        // seem to refer to the same actual param (like some aliases).
+        return this.sendGDBSet(`target-async ${state}`);
```
simark: If we have access to the version of the GDB being used, it would be nice to use the newer name (mi-async), just to avoid propagating deprecated usage (e.g. if people copy this code or inspect the MI logs).

src/GDBDebugSession.ts
simark: Hmm it doesn't look easy, as the pause request expects a `StoppedEvent` with a `pause` reason to be emitted.

I don't have a strong opinion.  I wish we would not handle SIGINT differently from other signals, but I understand that you have to fit a square piece in a triangle hole.

Looking at the current version of the specification for `Stopped Event`, we have both a `reason` and `description` field. One possibility would be to always send `pause` as the reason, and the put the signal name in `description`.  This is not ideal though, as `description` is meant to be a translated, human-readable string...

src/GDBBackend.ts
thegecko: Which version of gdb saw this change?

simark: 7.8: https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=gdb/NEWS;h=34c544c3d51a315bf5cc020f415aca8a13b73895;hb=HEAD#l1850

thegecko: Thanks, our use case uses the older form. Can the version be retrieved/parsed from gdb to determine the best command? Alternatively perhaps a new argument to specify the version?

marechal-p:
Just a heads up: I'll keep working on that around the end of next week, as I will be on vacation.

src/GDBDebugSession.ts
marechal-p: I did as you mentioned:

- `reason` is always `pause`
- `text` is the signal name.
- `description` is the meaning of the signal.

According to the spec, it seems to somewhat fit.

src/GDBBackend.ts
marechal-p: Would it be fine if I were to try the newest one and fallback to the legacy on error?

That's what we've been doing for the `runInTerminal` feature already, we may use the same strategy here?

marechal-p: Done using a fallback mechanic, can you tell me if it brings any problem to you @thegecko ?

marechal-p:
Bump.

Should be good now. I don't know how it handles on Windows, if anyone wants to test.

jonahgraham:
rerun tests 

The last run of the tests are failing so I am rerunning.

marechal-p:
Apparently it is the stop test where I try to pause a `while (1) { }` program.

Do you know why it would work on my machine but not on the runner?

jonahgraham:
I don't know - I just had a closer look and I note it is passing on the local debug, but not the (new) remote debug. Is it passing for you locally doing remote debug? `mocha --test-remote [...]`

jonahgraham:
@marechal-p  Either way I will have a look, but the extra datapoint is useful for me to debug the issue.

src/GDBDebugSession.ts
```diff
@@ -140,6 +140,7 @@ export class GDBDebugSession extends LoggingDebugSession {
             this.gdb.on('notifyAsync', (resultClass, resultData) => this.handleGDBNotify(resultClass, resultData));
 
             await this.spawn(args);
+            await this.gdb.sendMIAsync('on');
```
jonahgraham: This line is missing in GDBTargetDebugSession.ts. Adding it there makes it work for me locally.

```
@@ -162,10 +162,11 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         }
         const target = args.target;
         try {
             this.isAttach = true;
             await this.spawn(args);
+            await this.gdb.sendMIAsync('on');
             await this.gdb.sendFileExecAndSymbols(args.program);
             await this.gdb.sendEnablePrettyPrint();
 
             const targetType = target.type !== undefined ? target.type : 'remote';
             let defaultTarget: string[];
```

jonahgraham:
@marechal-p This PR seems to have suffered in that between when you first wrote it and today I have added remote support across the board. All tests are now run with and without remote mode. But the remote session still needs to be updates (as indicated by line comment above).

jonahgraham:
rerun tests

marechal-p:
Ah, this is my bad again, as I ran the test before rebasing but forgot to re-test after...

Thanks for the fix.

marechal-p:
Should we also test remote + openGdbConsole somehow?

jonahgraham:
> Should we also test remote + openGdbConsole somehow? (Not in this PR)

Probably :-) I'll add that now. I can review this PR too, but as I haven't been involved up until now if someone else has a further opinion they should review that too.

jonahgraham:
I rebased with PR #125 applied

marechal-p:
IMO, the most opinion-worthy changes would be the following:
- https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/116/files#diff-381ca8480838665aac12525afe59fcafR116-R127
- https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/116/files#diff-c76f134d8c9cea20536b5a2761643886R656-R683

Aside from that, I am happy with the change.

jonahgraham:
> Should be good now. I don't know how it handles on Windows, if anyone wants to test.

The changes LGTM - but with the assumption that `-exec-interrupt` works always. DSF-GDB does not use `-exec-interrupt` in all-stop mode unless the "Debugger Console" (aka new-ui) is in use. That basically means never on Windows.

But what was the state in Windows before this change? Supported or not? If we aren't supporting windows this change is probably fine.

jonahgraham:
In the absence of someone addressing Windows issue I will test it out later in the week.

marechal-p:
> DSF-GDB does not use `-exec-interrupt` in all-stop mode unless the "Debugger Console" (aka new-ui) is in use.

Do you know the reason for that? We are running tests with and without the new-ui (`--run-in-terminal`), and the test passes.

jonahgraham:
> Do you know the reason for that?

Not beyond what is in the CDT trail. Prior to GDB 7.12 and the new-ui, CDT always did async off (when in all stop). It was initially turned on in [patch 84382](https://git.eclipse.org/r/#/c/84382/) for [Bug 303808](https://bugs.eclipse.org/bugs/show_bug.cgi?id=303808) something was presumably not quite right as [Bug 511243](https://bugs.eclipse.org/bugs/show_bug.cgi?id=511243) was filed, with the key line being:

> [async mode always on] is not good as async mode may not work properly with GDB 7.12 for e.g. Windows and MAC

Is Alvaro at Ericsson still? Perhaps you can ask him to comment further?

jonahgraham:
> Merge branch 'master' into mp/async-on

I did this - but it was my first time doing it in the github interface - I will probably manually rebase before merging.

jonahgraham:
I don't think I had been paying attention up until now, but as far as I can tell cdt-gdb-adapter does not work on Windows for the simple case of slashes in filenames. I am working on this now, but until that is resolved I can't test this patch. I suspect there may be other issues (For example https://github.com/eclipse-cdt/cdt-gdb-adapter/commit/0e015f11f9416e97ad5f835aa1e54143bdc095aa means yarn test does not work on Windows anymore because that commit introduces bash like syntax for setting environment variables.)

Does anyone have the adapter running and/or building on Windows yet?

That said, I would recommend we just go ahead with this patch (rebased) assuming that indeed no one has Windows working yet anyway. We may then have to revert part of the patch and have a flag that keeps track if running in sync or async mode to know how to interrupt target.

jonahgraham:
@thegecko I noticed you are not on the gitter channel yet, https://gitter.im/eclipse-cdt/debug-adapter - do you have an answer to "Does anyone have the adapter running and/or building on Windows yet?"

thegecko:
@jonahgraham We use this adapter on Windows, Linux and Mac

marechal-p:
Closing because inactive for the moment.

PR #120, opened 2019-07-22
Use protected fields for easy overriding

thegecko:
Signed-off-by: thegecko <rob.moran@arm.com>

This PR modifies all `private` fields so they become `protected`.

As inheriting from this debug adapter is a recommended use case, this change makes overriding base fields a lot easier and avoids having to duplicate code in inherited classes.

Our specific use case was to override the variables functionality to introduce new scopes, but it seemed sensible (and trivial) to extend this to all private fields.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -35,10 +35,10 @@ export declare interface GDBBackend {
 }
 
 export class GDBBackend extends events.EventEmitter {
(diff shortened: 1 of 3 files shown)


dschaefer:
I'm not sure what you mean by "overriding" the fields. But certainly having read/write access from class extensions makes sense.

PR #123, opened 2019-08-04
Store each mocha run in its own junit file

jonahgraham:
This needs to be in the package.json because mocha is
run repeatedly during one run of `yarn test` and therefore
the REPORT_PATH must be different for each invocation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,9 +7,10 @@
   "scripts": {
     "build": "tsc",
     "watch": "tsc -w",
-    "test": "yarn test:integration && yarn test:pty && yarn test:integration --run-in-terminal",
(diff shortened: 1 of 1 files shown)


jonahgraham:
This has a matching commit in cdt-infra: https://github.com/eclipse-cdt/cdt-infra/commit/e8c6038310d3f048bb7f3141114c0de5bdc4f36e

The effect of this commit is that the test report [trend line](https://ci.eclipse.org/cdt/view/npm%20builds/job/cdt-gdb-adapter-master/test/?width=800&height=600) gets a big boost as it now shows all the tests.

Issue #114, opened 2019-06-27
Support remote debugging?


thegecko:
Is there any interest in extending this adapter to support remote debugging using gdb server (or other server implementations)?

I've created a PR(https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/109) which introduces the ability to connect to an existing gdb server, but there was a short discussion around whether the adapter should also manage launching the gdb server process where possible.

So if this feature is in scope of this adapter, should it support remote attach, launch or both?

1. **Attach** would introduce an additional `remote` argument for gdb to connect to and change the startup / load commands accordingly (as in the PR).
Management of starting the remote process and stopping it would be outside of the scope of the adapter.

2. **Launch** would introduce the ability to start a gdb server and would need some further arguments adding, `gdb server executable` and `gdb server arguments` at a minimum.
In this case would the adapter then be responsible for determining the port to open for client-server communication (scanning for unused ones as necessary) or would that still be the responsibility of the hosting application?
Finally, how would the adapter know when the gdb server application is ready? I've seen methods which scan the port being used for activity and an alternative way would be to scan stdout from the server for a known `ready` string (although this can differ between gdb server implementations and may need to be an additional argument).

Implementation of the above could be quite onerous, so I wanted to make sure we have a plan before starting.

@dschaefer @marechal-p your thoughts?

dschaefer:
Full disclosure :), at QNX we've built an extension to this adapter that handles QNX remotes. We use a slightly different protocol (target qnx), and a different way to get the binaries to the remotes. But the bulk of the rest of the logic is inherited from the GDB adapter.

I'm guessing we could do the same for gdbserver, or even remotes that connect to hardware debuggers or gdb stubs. We could have a few debugger types which start up a different subclasses of GDBDebugSession to manage the connection to the remote and whatever is needed to launch.

I think the Ericsson guys have more experience with such things with Linux. Interested in their thoughts.

thegecko:
I'd be interested to know how you extend the adapter as this sounds like exactly what we need to do.

marechal-p:
Just so that I understand a bit better:

> to manage the connection to the remote and whatever is needed to launch.

@dschaefer Are you talking about having the debug adapter launching the gdb server itself, or remote connecting to something already running and then launching the inferior?

dschaefer:
[...]

PR #125, opened 2019-08-06
#114: complete test matrix: add remote && run in terminal

jonahgraham:
This PR completes the test matrix as suggested in https://github.com/eclipse-cdt/cdt-gdb-adapter/pull/116#issuecomment-518806503


marechal-p:
Thanks!

PR #126, opened 2019-08-06
#114: Bump version of adapter now that remote is merged

jonahgraham:


jonahgraham:
I think this still needs to be done so downstream can properly version this dependency - but I think I should coordinate before doing it. I also think that some work has already been done in this area and I don't want to reinvent the wheel here.

dschaefer:
We should really look at getting the adapter published regularly. The 'yarn publish' command takes care of upversioning. It's currently commented out in the Jenkinsfile. I had been doing it by hand on my machine.

We'll also need to make sure more committers than myself have publish rights for our eclipse-cdt org on npm.

jonahgraham:
Closing this in favour of the work being done in #133

PR #127, opened 2019-08-07
 #114: Add more options to remote debugging     

jonahgraham:
These new options are to allow migration of CDT's GDB JTAG launches in the adapter.

There are a few separate commits that clean up some other minor issues too.


jonahgraham:
The gerrit https://git.eclipse.org/r/#/c/147162/ implements the client side of this for the Eclipse IDE.

PR #129, opened 2019-08-15
Add async status event for '+'

thegecko:

This PR adds support for the status event output from MI which uses the `+` character.
Tested for status output from image download progress on remote debugging.


jonahgraham:
Thanks @thegecko!

PR #130, opened 2019-08-16
Always send stop event

thegecko:

This PR changes the gdb stopped handler to always notify the host that the debugger has stopped. There are generic stop events which happen without a reason and this can cause the IDE to get out of sync.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -670,7 +670,7 @@ export class GDBDebugSession extends LoggingDebugSession {
                 }
                 break;
             default:
-                logger.warn('GDB unhandled stop: ' + JSON.stringify(result));
+                this.sendStoppedEvent('generic', parseInt(result['thread-id'], 10));
         }
     }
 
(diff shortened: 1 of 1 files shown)


jonahgraham:
@thegecko this causes a test failure doing remote debugging - my guess is that we used to silently ignore an initial stopped event that is now being issued as generic, and this causes test to fail as it is expecting a breakpoint event.

thegecko:
I concur, the tests still pass fine for non-remote scenarios.
We do need to emit the generic stopped event to keep any UI in sync with the debugger. Ways to fix the tests could be:

- Understand why the remote debugger is stopping and see if it can be suppressed
- Override the base `debugClient` `hitBreakpoint` function to ignore generic stop events

Thoughts?

thegecko:
Removed notify classes in favour of #131

jonahgraham:
@thegecko I think the difference is when you do target remote GDB immediately gives a stopped event because the target is indeed stopped at that point. This does not happen in run mode because there is nothing stopped or running until `run` is issued. So there needs to be a way to identify that first stopped as reflecting the state the adapter is already in at that point. As it turns out the silentlyish ignoring that first stopped event did achieve that.

thegecko:
I tried overriding the debug tests, but it wasn't particularly clean. I was hoping there was a way to simply ignore `generic` stop events, but it isn't easily achieved without overriding a lot of the class.

jonahgraham:
I further wonder what the effect of sending a StoppedEvent from the adapter before the launch completes, and the effect of sending two stopped events to the client without an intermediary run/etc. I'll try to look at this in a bit more detail tomorrow, but I suspect we don't want to pass this initial stopped event to the client at all.

thegecko:
Ok

jonahgraham:
rerun tests

(I have to say it right to Jenkins)

thegecko:
Yup, LGTM

PR #131, opened 2019-08-22
Add known notify classes

thegecko:

Add a few known gdb notify classes to suppress warnings.

Split from #130


dschaefer:
Thanks!

PR #135, opened 2019-09-19
Support allThreadsStopped marker on StoppedEvent

thegecko:
Signed-off-by: thegecko <rob.moran@arm.com>

The StoppedEvent raised from the adapter supports a boolean `allThreadsStopped` flag to indicate to the UI that all threads have stopped.
Gdb returns this in the stopped resultData as `stopped-threads`.

This PR marries the two up to fix scenarios such as the following:

Before:

Screenshot 2019-09-19 at 14 43 58

After:

Screenshot 2019-09-19 at 14 43 12

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -11,13 +11,14 @@ import * as os from 'os';
 import * as path from 'path';
 import {
     Handles, InitializedEvent, Logger, logger, LoggingDebugSession, OutputEvent, Response, Scope, Source,
-    StackFrame, StoppedEvent, TerminatedEvent, Thread,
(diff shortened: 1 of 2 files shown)


jonahgraham:
run tests

jonahgraham:
Thanks @thegecko - I had this on my list to check after raising questions about all vs non-stop so I am glad you fixed this. LGTM, just waiting for build machine to confirm.

However, if would be really nice to have a test for this. Do you have a multi-threaded test we can use here?

thegecko:
> However, if would be really nice to have a test for this. Do you have a multi-threaded test we can use here?

I don't unfortunately and time is very much against me

Issue #12, opened 2018-11-13
Test strategy


dschaefer:
Time to put together a test strategy for the adapter. We want to be able to confidently make changes without breaking the clients.

We have experience with mocha but I hear that it doesn't have good Jenkins reporting which we need for the CI machine.

Anyone else have any ideas?

dschaefer:
There's probably two levels of testing here. Unit and integration. Unit tests the APIs of the components, where integration would actually test against a running gdb and fake out a client.

simark:
For the integration test using a fake client, this seems nice, part of the vscode-debugadapter repo:

https://github.com/Microsoft/vscode-debugadapter-node/tree/master/testSupport

I am planning to start writing some tests with it, any objections?

What kind of test output does your Jenkins job want?  It is possible to write new reporters for mocha, so I assume it's possible to output what you need.

For the unit tests, mocha + chai is a combo that works well, that's what we use in Theia.

dschaefer:
> For the integration test using a fake client, this seems nice, part of the vscode-debugadapter repo:
> 
> https://github.com/Microsoft/vscode-debugadapter-node/tree/master/testSupport
> 
> I am planning to start writing some tests with it, any objections?

That would be awesome! Go ahead.

> 
> What kind of test output does your Jenkins job want? It is possible to write new reporters for mocha, so I assume it's possible to output what you need.
> 
> For the unit tests, mocha + chai is a combo that works well, that's what we use in Theia.

Yeah, that's what we use for our web projects here at QNX. The biggest issue we've had is with finding good reporters that present the results nicely. I'll have to see what the current state of all that is.

dschaefer:
Actually we have a reporter we're using internally. I should be able to open it.

dschaefer:
We have the integration story in place. Thanks @simark!
Be good to have a unit test strategy where we mock out gdb as well.

jonahgraham:
@thegecko Please comment here and I will try to figure it out.

jonahgraham:
In #136 @thegecko mentions issues with MacOS. I didn't know that tests didn't work on MacOS. I am in the process of bringing up a machine to add to the CI to regtest on Windows, but I don't have any immediate plans on Mac - but I could do the same there at some point. 

@thegecko the best would be that tests "just worked" on Mac. What is the issue there? Is it the getting a GDB/GCC/toolchain that works easily, or something else?

thegecko:
[...]

PR #136, opened 2019-09-22
Add support for function breakpoints

thegecko:

Introduced support for adding function breakpoints, see:

https://code.visualstudio.com/docs/nodejs/nodejs-debugging#_function-breakpoints

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -103,6 +103,7 @@ export class GDBDebugSession extends LoggingDebugSession {
 
     protected frameHandles = new Handles<FrameReference>();
     protected variableHandles = new Handles<VariableReference>();
+    protected functionBreakpoints: string[] = [];
     protected logPointMessages: { [ key: string ]: string } = {};
 
     protected threads: Thread[] = [];
@@ -139,6 +140,7 @@ export class GDBDebugSession extends LoggingDebugSession {
         response.body.supportsConditionalBreakpoints = true;
         response.body.supportsHitConditionalBreakpoints = true;
         response.body.supportsLogPoints = true;
+        response.body.supportsFunctionBreakpoints = true;
         // response.body.supportsSetExpression = true;
         response.body.supportsDisassembleRequest = true;
         this.sendResponse(response);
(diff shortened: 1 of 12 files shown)


src/mi/breakpoint.ts
```diff
@@ -66,3 +66,8 @@ export function sendBreakDelete(gdb: GDBBackend, request: {
 export function sendBreakList(gdb: GDBBackend): Promise<MIBreakListResponse> {
     return gdb.sendCommand('-break-list');
 }
+
+export function sendBreakFunctionInsert(gdb: GDBBackend, fn: string) {
+    const command = `-break-insert -t --function ${fn}`;
```
jonahgraham: Intentional temporary breakpoint?

thegecko: No, thanks

thegecko:
Tests would be a lot easier to write if they worked on MacOS or ran on PRs (with the output shown)

jonahgraham:
> Tests would be a lot easier to write if they worked on MacOS or ran on PRs (with the output shown)

Please add details on Issue #12 - or raise a new issue if needed.

thegecko:
> I think this still doesn't meet the DAP spec

Still WiP

thegecko:
Rebased and added tests, still need to resolve the issue with not removing them...

jonahgraham:
> @thegecko pushed 1 commit.
>
>    88fcd1f Support removing function breakpoints

@thegecko does this mean it is ready for review? Let me know when/if it is and I will try to get it in.

thegecko:
@jonahgraham This should be good to go. I've got a change which de-duplicates the breakpoint resolving in `setBreakpoints` and `setFunctionBreakpoints`, but felt that should be a PR after this one. Let me know if you'd prefer to see it added here, however.

jonahgraham:
> Let me know if you'd prefer to see it added here, however.

Lets avoid the duplication in the first place - if you are ready to add it now, please do that. It will save me having to compare the setBreakpoints and setFunctionBreakpoints versions.

thegecko:
@jonahgraham, I've introduced a common `resolveBreakpoints()` function for determining the breakpoints to add and delete across `setBreakpoints()` and `setFunctionBreakpoints()`.

I've introduced a few more tests to exercise this new common function and I've tested it all locally.

It's probably worth focussing your checks on the breakpoint handling overall as there is some churn in that area.

I've also added a few minor bug-fixes and comments.

src/GDBDebugSession.ts
```diff
@@ -328,20 +339,83 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+            const gdbbps = result.BreakpointTable.body.filter((gdbbp) => {
+                // Only function breakpoints
+                return this.functionBreakpoints.indexOf(gdbbp.number) > -1;
+            });
+
+            const { inserts, existing, deletes } = this.resolveBreakpoints(args.breakpoints, gdbbps,
+                (vsbp, gdbbp) => {
+
+                // Always invalidate hit conditions as they have a one-way mapping to gdb ignore and temporary
+                if (vsbp.hitCondition) {
```
jonahgraham: `hitCondition` and `condition` - VSCode [does not yet have](https://github.com/microsoft/vscode/issues/3646) UI for this so it seems acceptable to not implement support for now, but we are at some point going to need our own fork of the debug protocol spec to document our differences and extensions and this is one of them.

src/integration-tests/functionBreakpoints.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+                {
+                    name: 'sub',
+                },
+            ],
+        });
+        await dc.configurationDoneRequest();
+        const event = await dc.waitForEvent('stopped') as StoppedEvent;
+        expect(event.body.reason).to.eq('function breakpoint');
+        const scope = await getScopes(dc);
+        expect(scope.frame.line).to.eq(2);
```
jonahgraham: There is already a helper that does this:

```
        dc.assertStoppedLocation('function breakpoint', {line: 2});
```

PR #137, opened 2019-09-22
Support hit conditional breakpoints

thegecko:

Introduce support for ignore counts via the `supportsHitConditionalBreakpoints` configuration.

### DIFF
--- a/src/mi/breakpoint.ts
+++ b/src/mi/breakpoint.ts
@@ -47,8 +47,10 @@ export async function sendBreakInsert(gdb: GDBBackend, request: {
     location: string;
 }): Promise<MIBreakInsertResponse> {
     // Todo: lots of options
+    const temp = request.temporary ? '-t ' : '';
+    const ignore = request.ignoreCount ? `-i ${request.ignoreCount} ` : '';
     const escapedLocation = gdb.standardEscape(request.location);
-    const result = await gdb.sendCommand<MIBreakInsertResponse>(`-break-insert ${escapedLocation}`);
+    const result = await gdb.sendCommand<MIBreakInsertResponse>(`-break-insert ${temp}${ignore}${escapedLocation}`);
 
     if (request.condition) {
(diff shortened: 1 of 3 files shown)


src/mi/breakpoint.ts
```diff
@@ -47,8 +47,9 @@ export async function sendBreakInsert(gdb: GDBBackend, request: {
     location: string;
 }): Promise<MIBreakInsertResponse> {
     // Todo: lots of options
+    const ignore = request.ignoreCount ? `-t -i ${request.ignoreCount} ` : '';
```
jonahgraham: Is it intentional that temporary flag is used here?

thegecko: I meant to make that temporary based on the existence of `>` or not.

So if the user sets hit count `5`, it will break after 5 hits and not break again.

If the user sets hit count `>5`, it will break after 5 hits and continue to break.

Have updated PR.

thegecko:
@jonahgraham This PR should be ready to go

jonahgraham:
There are failing tests [here](https://ci.eclipse.org/cdt/job/cdt-gdb-adapter-verify/87/console), but it worked on my machine:

rerun tests

jonahgraham:
rerun tests

jonahgraham:
The intermittent failing tests have nothing to do with this change.

jonahgraham:
Thanks @thegecko for this.

jonahgraham:
I have tracked down this particular intermittent failure - it was a problem with the testsuite. PR coming up.

thegecko:
Cool, I'm still trying to devise some tests for the logPoints PR

jonahgraham:
Have a look at PR #144  and #145 - there is still an intermittent failure I have not tracked down ([link](https://ci.eclipse.org/cdt/view/npm%20builds/job/cdt-gdb-adapter-verify/88/console))

PR #139, opened 2019-09-22
Initial support for log points

thegecko:

This PR adds initial support for log points (breakpoints which log a message instead of stopping).

Currently, the new functionality simply logs the message provided when the logpoint is hit, but the [specification](https://microsoft.github.io/debug-adapter-protocol/specification) outlines that it should support `Expressions within {} are interpolated`.

This isn't yet added, I'll follow up with a separate PR to interpolate expressions within `{}`.

### DIFF
--- a/src/integration-tests/logpoints.spec.ts
+++ b/src/integration-tests/logpoints.spec.ts
@@ -1,0 +1,60 @@
+/*********************************************************************
+ * Copyright (c) 2019 Arm and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *********************************************************************/
+
+import { join } from 'path';
+import { expect } from 'chai';
+import { CdtDebugClient } from './debugClient';
+import { LaunchRequestArguments } from '../GDBDebugSession';
+import {
+    standardBeforeEach,
+    gdbPath,
+    testProgramsDir,
+    openGdbConsole
+} from './utils';
+
+describe('logpoints', async () => {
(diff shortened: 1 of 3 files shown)


thegecko:
@jonahgraham Rebased and added tests

src/integration-tests/logpoints.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+            breakpoints: [
+                {
+                    column: 1,
+                    line: 4,
+                    logMessage,
+                },
+            ],
+        });
+        await dc.configurationDoneRequest();
+        let logEvent = await dc.waitForOutputEvent('console');
```
jonahgraham: There is a probable race condition here - i.e. it is possible for the event to arrive before the waitForOutputEvents gets to the point it is listening for events. This is the same kind of race condition that I had to resolve with PR #145. I suspect the way to do this is Promise.all() and call waitForOutputEvent first.

jonahgraham:
@thegecko This looks good, I am just concerned about introducing a new unstable test. If you think I am wrong about that, please let me know.

src/integration-tests/logpoints.spec.ts
thegecko: The helper function I wrote ignores a lot of `console` output before reaching the expected `output` category, so I'd be surprised if a race condition occurred. I've not seen anything during my testing.

Using `Promise.all()` could come with it's own problems as the `waitForEvent()` model uses timeouts to avoid locking up the tests. Running it first would mean a suitable high timeout would be needed in order for the rest of the setup to complete. This in itself could become a race condition :/

It would become apparent in the CI quickly if this issue does arise and I would be happy to revisit this PR in that case.

jonahgraham:
That sounds good enough for me. Thank you for picking this up in the unlikely case that there is a race exposed on CI.

PR #145, opened 2019-11-05
Make sure to stopped event arrives before continuing tests

jonahgraham:
Without the assertStoppedLocation calls there is a race condition between
the calls to nextRequest and what happens after


jonahgraham:
rerun tests

jonahgraham:
rerun tests

PR #146, opened 2019-11-07
Include tsc as part of 'yarn install' step

jonahgraham:
And improve the documentation around the same


jonahgraham:
rerun tests

jonahgraham:
The test failure is fixed by #145, so merging without a passing build.

PR #148, opened 2019-11-21
A handful of small cleanups

jonahgraham:
@thegecko here is the PR with the changes I referenced in #136


jonahgraham:
I'll push to npm in a few hours.

Issue #152, opened 2019-11-29
setBreakpoints and setFunctionBreakpoints don't respect spec


jonahgraham:
I am reviewing the setFunctionBreakpoints and setBreakpoints a little more closely after running into a problem integrating the adapter into Eclipse with LSP4E.

There are a few issues with breakpoint handling that needs addressing, I am using this one issue to track them.

- [x] breakpoints simultaneously in multiple files
- [x] breakpoints simultaneously in multiple files on Windows
- [x] comparing of filename is problematic (see conversation in #156)
- [x] "bad" breakpoints are causing error response instead of breakpoints response with unverified breakpoints (see Breakpoint.verified and Breakpoint.message)
- [x] the order of Breakpoints in the responses' body does not follow "The array elements are in the same order as the elements of the 'breakpoints'"
- [x] breakpoints that are inserted on relocated lines do not have the relocated line reported back in the response
- [x] breakpoints requested on lines that GDB relocates causes subsequent setBreakpoints to reinsert the breakpoint instead of reusing the already relocated breakpoint. This is a particular problem when hit counting is being used
- [x] breakpoints with multiple locations (identified with `<MULTIPLE>` in `addr` field from gdb and sub breakpoints, e.g. 1.1 and 1.2 as ids) causes issues as adapter does not identify them as working together and in some cases tries to delete the multiple values.
- [x] MI parser does not parse return values with `<MULTIPLE>` properly on `-break-insert` - it does handle it properly on `-break-list` because GDB has a different format on `-break-insert` result. The `bkpt=` is not duplicated on the insert case. 
- [ ] Some behaviour is conditional on GDB version - e.g. < 8.2 does not report relocated breakpoint new line number - therefore tests need to be dependent on GDB version and we need to start running multiple versions

jonahgraham:
See PR #154 - I have started adding (now failing) tests for each of the issues already. Only "breakpoints with multiple locations" issue does not cause a test case failure - but the MI log shows a problem.

jonahgraham:
> Only "breakpoints with multiple locations" issue does not cause a test case failure - but the MI log shows a problem.

Latest update to PR #154 makes this fail too because instead of silently logging log-output-stream (&) messages, I changed it to be an output event on the 'log' stream. This allows the test to verify that the stream does not have a warning.

jonahgraham:
[...]

PR #154, opened 2019-11-29
Bug #152: Tests and fixes for numerous breakpoint issues

jonahgraham:
See #152

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -268,7 +268,7 @@ export class GDBDebugSession extends LoggingDebugSession {
                 return this.functionBreakpoints.indexOf(gdbbp.number) === -1;
             });
 
-            const { inserts, existing, deletes } = this.resolveBreakpoints(args.breakpoints || [], gdbbps,
+            const { resolved, deletes } = this.resolveBreakpoints(args.breakpoints || [], gdbbps,
                 (vsbp, gdbbp) => {
 
                     // Always invalidate hit conditions as they have a one-way mapping to gdb ignore and temporary
@@ -316,18 +316,21 @@ export class GDBDebugSession extends LoggingDebugSession {
                 };
             };
 
-            const actual = existing.map((bp) => createState(bp.vsbp, bp.gdbbp));
+            const actual: DebugProtocol.Breakpoint[] = [];
 
-            existing.forEach((bp) => {
-                if (bp.vsbp.logMessage) {
-                    this.logPointMessages[bp.gdbbp.number] = bp.vsbp.logMessage;
+            for (const bp of resolved) {
+                if (bp.gdbbp) {
+                    actual.push(createState(bp.vsbp, bp.gdbbp));
+                    if (bp.vsbp.logMessage) {
+                        this.logPointMessages[bp.gdbbp.number] = bp.vsbp.logMessage;
+                    }
+                    continue;
                 }
-            });
 
-            for (const vsbp of inserts) {
+
                 let temporary = false;
                 let ignoreCount: number | undefined;
-
+                const vsbp = bp.vsbp;
                 if (vsbp.hitCondition !== undefined) {
                     ignoreCount = parseInt(vsbp.hitCondition.replace(ignoreCountRegex, ''), 10);
                     if (isNaN(ignoreCount)) {
@@ -393,7 +396,7 @@ export class GDBDebugSession extends LoggingDebugSession {
                 return this.functionBreakpoints.indexOf(gdbbp.number) > -1;
             });
 
-            const { inserts, existing, deletes } = this.resolveBreakpoints(args.breakpoints, gdbbps,
+            const { resolved, deletes } = this.resolveBreakpoints(args.breakpoints, gdbbps,
                 (vsbp, gdbbp) => {
 
                     // Always invalidate hit conditions as they have a one-way mapping to gdb ignore and temporary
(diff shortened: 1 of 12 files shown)


jonahgraham:
This is not ready to merge yet. It is here for visibility to other contributors.

src/integration-tests/functionBreakpoints.spec.ts
```diff
@@ -45,7 +45,7 @@ describe('function breakpoints', async () => {
             ],
         });
         await dc.configurationDoneRequest();
-        dc.assertStoppedLocation('function breakpoint', { line: 6 });
+        await dc.assertStoppedLocation('function breakpoint', { line: 6 });
```
jonahgraham: when I simplified this code the other day I left out the await - cc @thegecko FYI

src/GDBDebugSession.ts
```diff
@@ -333,14 +333,20 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+                        location: `${file}:${vsbp.line}`,
+                        condition: vsbp.condition,
+                        temporary,
+                        ignoreCount,
+                    });
+                    actual.push(createState(vsbp, gdbbp.bkpt));
+                } catch (err) {
+                    actual.push({
+                        verified: false,
+                        message: err.message,
```
jonahgraham: I don't know if the err.message is sufficient yet - maybe more info is needed?

jonahgraham:
rerun tests

jonahgraham:
:-( There is an unstable test in this PR - passes reliably on my machine (locally and yarn docker:test) and fails most of the time on build machine.

jonahgraham:
> unstable test in this PR 

I think this is the same problem as #145 - just for continueRequest instead of next/step/etc.

jonahgraham:
OK. that worked: CI rerun tests to see if it was a fluke.

jonahgraham:
@thegecko If you want to review this let me know. I'll merge tomorrow if not.

The first three commits are all the new tests, and the other commits fix specific issues. I don't plan to squash these commits, but I'll merge them with a merge commit to keep the related items together.

src/GDBDebugSession.ts
```diff
@@ -407,12 +427,26 @@ export class GDBDebugSession extends LoggingDebugSession {
...
-                this.functionBreakpoints.push(gdbbp.bkpt.number);
-                actual.push(createActual(gdbbp.bkpt));
+            for (const bp of resolved) {
+                if (bp.gdbbp) {
+                    actual.push(createActual(bp.gdbbp));
+                    continue;
+                }
+
+                try {
+                    const vsbp = bp.vsbp;
```
thegecko: nit: If this is only used once, it doesn't need putting in a separate variable IMO

src/GDBDebugSession.ts
```diff
@@ -280,8 +285,8 @@ export class GDBDebugSession extends LoggingDebugSession {
                     const vsbpCond = vsbp.condition || undefined;
                     const gdbbpCond = gdbbp.cond || undefined;
 
-                    // TODO probably need more thorough checks than just line number
-                    return !!(gdbbp.line && parseInt(gdbbp.line, 10) === vsbp.line
+                    // Check with original-location so that relocated breakpoints are properly matched
+                    return !!(gdbbp['original-location'] === `${file}:${vsbp.line}`
```
thegecko: I wonder if `original-location` resolves to the same thing across OS's?

src/GDBDebugSession.ts
```diff
@@ -302,25 +307,34 @@ export class GDBDebugSession extends LoggingDebugSession {
                     this.logPointMessages[gdbbp.number] = vsbp.logMessage;
                 }
 
+                let line = 0;
```
thegecko: 👍

src/GDBDebugSession.ts
```diff
@@ -302,25 +307,34 @@ export class GDBDebugSession extends LoggingDebugSession {
...
-            const actual = existing.map((bp) => createState(bp.vsbp, bp.gdbbp));
+            const actual: DebugProtocol.Breakpoint[] = [];
 
-            existing.forEach((bp) => {
-                if (bp.vsbp.logMessage) {
-                    this.logPointMessages[bp.gdbbp.number] = bp.vsbp.logMessage;
+            for (const bp of resolved) {
+                if (bp.gdbbp) {
+                    actual.push(createState(bp.vsbp, bp.gdbbp));
+                    if (bp.vsbp.logMessage) {
```
thegecko: We no longer need to add the logmessage here as it is done in `createState()` above.

src/GDBDebugSession.ts
jonahgraham: AFAICT it does - however this code is still wrong as we need to apply standardEscape (GDBBackend) to it so it compares the same. On Windows it is a complicated mess of single, double and quadruple backslashes (e.g. `original-location="d:\\\\cdt\\\\git\\\\cdt-gdb-adapter\\\\src\\\\integration-tests\\\\test-programs\\\\count.c:4"` in the log)

jonahgraham:
I have changed to using `original-location` for matching breakpoints. However the way I implemented it requires GDB >= 7.11. See #159

jonahgraham:
OK - I think I have resolved everything.

I have run this with each version of GDB from 7.11 to 8.3 (using `yarn test:integration --gdb-path gdb.x.y` -- the docker container has all those, and [more](https://github.com/eclipse-cdt/cdt-infra/blob/6ee4556e97973003a9b0a06ced79a7baba023151/docker/cdt-infra-all-gdbs/ubuntu-18.04/Dockerfile#L35), on the PATH)

I am going to merge this (once CI confirms) - if #159 comes back needing < 7.11 I will revisit again. I think this changes I made in fa04aad can be made to work with < 7.11 - but I won't bother unless there is a compelling reason.

thegecko:
Awesome, can we have a release on npm?

PR #156, opened 2019-12-03
Fix breakpoints being set in multiple files

thegecko:
With reference to #152, this PR fixes a regression where `setBreakpoints()` would remove all previous breakpoints not for the current file.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -258,12 +258,16 @@ export class GDBDebugSession extends LoggingDebugSession {
             // deleting ones not requested and inserting new ones.
 
             const result = await mi.sendBreakList(this.gdb);
+            const file = args.source.path as string;
             const gdbbps = result.BreakpointTable.body.filter((gdbbp) => {
+                // Ignore other files
+                if (gdbbp.fullname !== file) {
+                    return false;
+                }
                 // Ignore function breakpoints
                 return this.functionBreakpoints.indexOf(gdbbp.number) === -1;
             });
 
-            const file = args.source.path as string;
             const { inserts, existing, deletes } = this.resolveBreakpoints(args.breakpoints || [], gdbbps,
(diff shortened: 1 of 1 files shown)


jonahgraham:
:-( This doesn't work on Windows.

The paths are the same, e.g.: D:\file.c and d:\file.c but of course they are not identical.

thegecko:
OK. both sides need a `.toLowercase()` to fix that.

This mismatch would have caused issues previously, too:

https://github.com/eclipse-cdt/cdt-gdb-adapter/blob/8084f5262b3f6c2fda394d642d171a179caf6f4d/src/GDBDebugSession.ts#L277

jonahgraham:
Yes, you are right, it never worked properly, but there didn't used to be tests that detected it. This one was detected "by accident" as a result of another test. I don't know how to tell from the debug client how many bps in total are installed - so I probably need to test installing and removing bps and make sure the execution does not stop there.

jonahgraham:
> OK. both sides need a .toLowercase() to fix that.

BTW do you know if this is the only canonicalization needed?

jonahgraham:
I agree @jld01 (good to see you here BTW!) - I think that we probably should be comparing against `original-location` (with line number stripped out) as that path is unmodified by GDB. One of my concerns is if a client passes paths with things like symlinks or .. in it GDB will canonicalize that. Also using source path mapping also changes the `fullname` entry so it probably isn't a stable item to compare to.

What would be nice (and perhaps in need of a proposal to DAP) is that when breakpoints are being set, the previously returned ID is passed back. That would remove all this special code.

thegecko:
Node has some build in functions for normalising paths, that could be a way to go regarding symlinks. It may also resolve the case problem, too.

https://nodejs.org/dist/latest-v12.x/docs/api/path.html#path_path_normalize_path

Issue #155, opened 2019-12-02
variableMap (part of varManager.ts) maintains global state between invocations


jonahgraham:
This happens only when using the debug adapter in server mode with a port - for example when debugging. 

The result of variableMap being global is that the session thinks it has already done var create for objects that it hasn't. 

The Variable Manager needs to be created for each session (like the backend is)

jonahgraham:
Fixed with #157

PR #157, opened 2019-12-03
Bug #155: Encapsulate var manager and make it an instance of GDBBackend

jonahgraham:


jonahgraham:
Hi @jcrwilliams - if you are still working on and familiar with the adapter I would appreciate a look at this please.

src/GDBBackend.ts
```diff
@@ -36,10 +37,15 @@ export declare interface GDBBackend {
 
 export class GDBBackend extends events.EventEmitter {
     protected parser = new MIParser(this);
+    protected varMgr = new VarManager(this);
     protected out?: Writable;
     protected token = 0;
     protected proc?: ChildProcess;
 
+    public getVarMgr(): VarManager {
```
thegecko: Could use a getter for this? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

jonahgraham: I could, but then it wouldn't look like a Java programmer wrote it ;0)

jonahgraham: fixed in 120459b

jonahgraham:
> Hi @jcrwilliams - if you are still working on and familiar with the adapter I would appreciate a look at this please.

@jcrwilliams I have merged this, but if you have any comments I can still fix it up.

PR #162, opened 2019-12-19
Fixed stackFrameRequest with thread ID

thegecko:

This PR fixes the `stackFrameRequest()` so that it supports the `thread ID` being passed in.

Without this fix, all stack frames returned are for the main thread and not the one selected from the UI.

Not sure how I'd write any tests for this.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -543,7 +543,8 @@ export class GDBDebugSession extends LoggingDebugSession {
             const levels = args.levels ? (args.levels > depth ? depth : args.levels) : depth;
             const lowFrame = args.startFrame || 0;
             const highFrame = lowFrame + levels - 1;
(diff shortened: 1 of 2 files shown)


jonahgraham:
LGTM. We will need multi-threaded tests sooner rather than later. DSF-GDB has some testcode that is useful: [org.eclipse.cdt.tests.dsf.gdb/data/launch/src](https://github.com/eclipse-cdt/cdt/tree/master/dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/data/launch/src) specifically a simple OS asbstraction layer for threads suitable for tests [Thread.h](https://github.com/eclipse-cdt/cdt/blob/master/dsf-gdb/org.eclipse.cdt.tests.dsf.gdb/data/launch/src/Thread.h)

PR #163, opened 2019-12-20
Stack frame depth fix and thread details

thegecko:
This PR introduces two changes. they are both small, but happy to split them into 2 PRs if either are controversial.

- After fixing the stack trace request in #162 , I've noticed the command to retrieve the stack depth also wasn't supporting threads. This is now fixed so that the depth is returned for the specified thread and not the main one every time.

- Threads return optional details, these are now returned in the name to give the user more info:

Screenshot 2019-12-20 at 09 52 05

I've also taken the liberty of adding myself as a contributor.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,9 @@
     "url": "git+https://github.com/eclipse-cdt/cdt-gdb-adapter.git"
   },
   "author": "Eclipse CDT",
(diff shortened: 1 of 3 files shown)


jonahgraham:
Lgtm. I like the thread details.

PR #164, opened 2020-02-20
Added TypeScript sources to published files

spoenemann:
Including the TypeScript sources in the published package is important to enable debugging when the package is used in another project.

I explicitly listed the source folders to include to avoid `integration-tests` to be included.


jonahgraham:
@spoenemann Thanks!

@eclipse-cdt-bot ok to test

jonahgraham:
It should be on npm soon with `@next` tag

Issue #167, opened 2020-07-01
Cannot Show variables of a Thread with tid > 2


mco-ericsson:
When debugging multiple threads, if you try to display the variables of a thread with a TID > 2, you will see:

`Unknown value for PRINT_VALUES: must be: 0 or \"--no-values\", 1 or \"--all-values\", 2 or \"--simple-values\"`

This is caused by a misuse of the -var-update MI command.

jonahgraham:
I cannot reproduce this. It may have been fixed in the meantime when I added better multi-thread support in #194 - there is now a multithread test that verifies it can read variables from all frames - see https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/integration-tests/multithread.spec.ts

I did find a bug while investigating this one, see #235 and its associated PR.

PR #168, opened 2020-07-01
Bug #167: Fix usage of -var-update

mco-ericsson:
-var-update does not take a threadId as a parameter but it takes
a printValues parameter which can be:
0 or --no-values
1 or --all-values
2 or --simple-values

This patch also defines the PrintValues to be able to send number
to gdb instead of strings for easier/faster parsing.


mco-ericsson:
forgot the signed-of line ...

PR #169, opened 2020-07-01
Bug #167: Fix usage of -var-update

mco-ericsson:
-var-update does not take a threadId as a parameter but it takes
a printValues parameter which can be:
0 or --no-values
1 or --all-values
2 or --simple-values

This patch also defines the PrintValues to be able to send number
to gdb instead of strings for easier/faster parsing.

Signed-off-by: mco-ericsson <manuel.coutand@ericsson.com>

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -700,7 +700,7 @@ export class GDBDebugSession extends LoggingDebugSession {
                     }
                     const children = await mi.sendVarListChildren(this.gdb, {
                         name: parentVarname,
-                        printValues: 'all-values',
+                        printValues: mi.MIVarPrintValues.all,
                     });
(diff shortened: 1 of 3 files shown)


marechal-p:
rerun tests

marechal-p:
@jonahgraham who can run the tests again?

jonahgraham:
run tests

jonahgraham:
> @jonahgraham who can run the tests again?

Committers can be default. What is strange is that you (@marechal-p) are in the whitelist. I guess the whitelist means your own PRs are automatically built?

marechal-p:
rerun tests

marechal-p:
Yes because it really seems like I cannot trigger the bot here, thanks for having done it for us!

PR #184, opened 2021-09-08
Encode data in read memory request in base 64

colin-grant-work:
This PR addresses the issue raised [here](https://github.com/eclipse-theia/theia-cpp-extensions/pull/119#discussion_r695749433) regarding the mismatch between the DAP's ReadMemoryRespnose, which specifies a base64 string, and CDT-GDB's response, which is a hexidecimal string.

Signed-off-by: Colin Grant <colin.grant@ericsson.com>

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -90,6 +90,19 @@ const ignoreCountRegex = /\s|\>/g;
 const arrayRegex = /.*\[[\d]+\].*/;
 const arrayChildRegex = /[\d]+/;
 
+export function hexToBase64(hex: string): string {
+    // The buffer will ignore incomplete bytes (unpaired digits), so we need to catch that early
+    if (hex.length % 2 !== 0) {
+        throw new Error('Received memory with incomplete bytes.');
+    }
+    const base64 = Buffer.from(hex, 'hex').toString('base64');
+    // If the hex input includes characters that are not hex digits, Buffer.from() will return an empty buffer, and the base64 string will be empty.
+    if (base64.length === 0 && hex.length !== 0) {
+        throw new Error('Received illformed hex input: ' + hex);
+    }
+    return base64;
+}
+
 export class GDBDebugSession extends LoggingDebugSession {
     protected gdb: GDBBackend = this.createBackend();
     protected isAttach = false;
@@ -944,7 +957,7 @@ export class GDBDebugSession extends LoggingDebugSession {
             const result = await sendDataReadMemoryBytes(
                 this.gdb, args.memoryReference, args.count, args.offset);
             response.body = {
-                data: result.memory[0].contents,
+                data: hexToBase64(result.memory[0].contents),
                 address: result.memory[0].begin,
             };
             this.sendResponse(response);
--- a/src/integration-tests/mem.spec.ts
+++ b/src/integration-tests/mem.spec.ts
@@ -11,7 +11,7 @@
 import { expect } from 'chai';
 import * as path from 'path';
 import { DebugProtocol } from 'vscode-debugprotocol/lib/debugProtocol';
-import { LaunchRequestArguments, MemoryResponse } from '../GDBDebugSession';
+import { hexToBase64, LaunchRequestArguments, MemoryResponse } from '../GDBDebugSession';
 import { CdtDebugClient } from './debugClient';
 import { expectRejection, gdbPath, openGdbConsole, standardBeforeEach, testProgramsDir } from './utils';
 
@@ -32,9 +32,9 @@ describe('Memory Test Suite', function() {
             program: memProgram,
             openGdbConsole,
         } as LaunchRequestArguments, {
-                path: memSrc,
-                line: 12,
-            });
+            path: memSrc,
+            line: 12,
+        });
         const threads = await dc.threadsRequest();
         // On windows additional threads can exist to handle signals, therefore find
         // the real thread & frame running the user code. The other thread will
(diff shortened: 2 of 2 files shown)


colin-grant-work:
@paul-marechal, would you mind taking a look at this when you get a chance?

src/GDBDebugSession.ts
```diff
@@ -90,6 +90,8 @@ const ignoreCountRegex = /\s|\>/g;
 const arrayRegex = /.*\[[\d]+\].*/;
 const arrayChildRegex = /[\d]+/;
 
+export const hexToBase64 = (hex: string): string => Buffer.from('0'.repeat(hex.length % 2) + hex, 'hex').toString('base64');
```
paul-marechal, line 93: 1. Use a regular function declaration.
2. Add a comment to explain what is going on with `'0'.repeat(hex.length % 2) + hex`?
    No need for it to be in the function's JSDoc as it's an implementation detail, but it had me struggle a bit...
    But from what I understand this is because the `Buffer` API expects to get pairs of hexadecimal characters in order to fully encode a byte, hence the `0` left padding.

src/integration-tests/mem.spec.ts
```diff
@@ -134,4 +134,21 @@ describe('Memory Test Suite', function() {
...
+    it('Converts hex to base64 without loss', () => {
+        const base64ToHex = (base64: string): string => Buffer.from(base64, 'base64').toString('hex');
+        const normalize = (original: string): string => original.replace(/^0+/, '').toLowerCase();
+
+        const testCases = [
+            'f',
+            'fe',
+            'fed',
+            'fedc',
+            '0fedc',
```
paul-marechal, line 147: Just out of curiosity but would gdb send back such weird output (with odd hexadecimal digits?)

Still a good test as it shows the function's resiliency against somewhat malformed data.

paul-marechal (commented):
Looks mostly good for me, but I have some concerns that I hope you can solve.

We need to convert from `hex` to `base64` but one misconception that we need to avoid making is that we are using `hex` to describe individual bytes, not numbers.

What I mean is that I am not sure we can safely expect to parse `hex` data like `abc` since this is technically missing 4 bits worth of information. In this case it might be simple to assume leading `0` are omitted, so the input should be assumed to be `0abc` and this we can convert back to 2 whole bytes.

Now my biggest concern is with `hex` inputs like `0ab`: What does this leading `0` actually mean? Is this really a lazy way to write `00ab` and differentiate from just `ab` without the leading `0`?

I feel like I might be nitpicking but I just want to make sure we don't misinterpret data.

src/integration-tests/mem.spec.ts
colin-grant-work: I don't think GDB will send back anything so weird. And in fact I don't think it will send back anything with an odd number of hex digits, because that would mean there was some strange four-bit  appendix hanging on. On the other hand, I thought the string-->buffer-->string transformation would be smooth, and then I noticed the truncation of odd digits, so I wanted to make sure that the implementation would handle a variety of funny inputs, just in case.

paul-marechal: Not sure what's best then, if we don't expect funny data but still get some: Fail early or display weird data?

colin-grant-work:
@paul-marechal, I've modified the base conversion to throw errors if it detects incomplete bytes (odd number of hex digits) or ill-formed hex (Buffer is empty -> base64 string is empty).

src/GDBDebugSession.ts
```diff
@@ -90,6 +90,17 @@ const ignoreCountRegex = /\s|\>/g;
...
+export function hexToBase64(hex: string): string {
+    if (hex.length % 2 !== 0) {
+        throw new Error('Received memory with incomplete bytes.');
+    }
+    const base64 = Buffer.from(hex, 'hex').toString('base64');
+    if (base64.length === 0 && hex.length !== 0) {
+        throw new Error('Received illformed hex input: ' + hex);
+    }
+    return base64;
+}
```
paul-marechal, line 102: Looking better! Though the `base64` output string will only be empty if `hex` is empty to begin with.

```ts
export function hexToBase64(hex: string): string {
    if (hex.length !== 0 || hex.length % 2 !== 0) {
        throw new Error('Malformed hex data (empty or incomplete)');
    }
    return Buffer.from(hex, 'hex').toString('base64');
}
```

src/integration-tests/mem.spec.ts
```diff
@@ -134,4 +134,34 @@ describe('Memory Test Suite', function() {
...
+
+        const wellFormedTestCases = [
+            'fe',
+            '00',
+            'fedc',
+            '00FE',
+            '29348798237abfeCCD',
+        ];
+
+
```
paul-marechal, line 150: One or none, but 2 empty lines look odd :)

Maybe a hint that we should setup linters at some point...

src/GDBDebugSession.ts
```diff
@@ -941,10 +952,11 @@ export class GDBDebugSession extends LoggingDebugSession {
     protected async readMemoryRequest(response: DebugProtocol.ReadMemoryResponse,
         args: DebugProtocol.ReadMemoryArguments): Promise<void> {
         try {
+
```
paul-marechal, line 955: Added an extra line by mistake?

src/GDBDebugSession.ts
colin-grant-work: The base64 string will also be empty if the hex is ill-formed (includes characters that aren't hex digits). In that case, the `Buffer.from()` call returns an empty buffer. See the test case that tests `zyxd` and does in fact `throw` due to the current check. It isn't something that should come up, but it would definitely be a problem if it did.

paul-marechal: See what I mean by things should fail early... This such a weird behavior from `Buffer` to just return empty string if there's an input it doesn't like :/

I might even had misread the check, but I understand now thanks!

paul-marechal (approved):
LGTM thanks!

Issue #187, opened 2022-04-16
Adapter does not work with older GDBs


jonahgraham:
The current adapter assumes a newish version of GDB. For example using `-source` option to `-break-insert`

jonahgraham:
#188 provides support back to GDB 7.6 on Linux.

jonahgraham:
I couldn't get a native GDB 7.6 to work on Windows, so for Windows I tested by running GDB 10 (from msys2) but having it advertise as GDB 7.6.

This issue is now resolve. If additional issues are found new bugs should be raised.

PR #189, opened 2022-04-17
Issue #178: Add more escaping on Windows for old GDB

jonahgraham:
The handling of \ on Windows causes lots of minor issues. While
CDT only does the double-escaping of paths if there is whitespace,
on cdt-gdb-adapter we need to do it all the time as the effect
on original-location is different. In CDT we don't worry about
original-location, but in cdt-gdb-adapter we need to be able to
identify which DAP breakpoint location matches which GDB breakpoint
reliably.


jonahgraham:
Actually #187 not #178

PR #190, opened 2022-04-29
Add register view feature

quyettrinhrvc:
Please help to review it.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1861,7 +1861,6 @@ export class GDBDebugSession extends LoggingDebugSession {
                     throw new Error('Unable to parse response for reg. values');
                 }
             }
-            //this.sendErrorResponse(response, 115, `List of register values: ${reg_values.toString()}`);
         }
         catch (error) {
             throw new Error('Unable to parse response for reg. names');
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1843,7 +1843,7 @@ export class GDBDebugSession extends LoggingDebugSession {
         }
 
         const result_values = await mi.sendDataListRegisterValues(this.gdb, {
-            fmt: ' x',
+            fmt: 'x',
             frameId: frame.frameId,
             threadId: frame.threadId,
         });
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1817,14 +1817,13 @@ export class GDBDebugSession extends LoggingDebugSession {
         if (!frame) {
             return Promise.resolve(variables);
         }
-        //result:;
+        
         try {
             if (this.registerMap.size === 0) {
                 ////// Refer to e2s plugin to port request code
                 const result_names = await mi.sendDataListRegisterNames(this.gdb, {});
                 let idx = 0;
                 const registerNames = result_names['register-names'];
-                //const registerNames = result_cv[0];
                 for (const regs of registerNames){
                     if(regs !== ''){
                         this.registerMap.set(regs, idx);
@@ -1842,11 +1841,8 @@ export class GDBDebugSession extends LoggingDebugSession {
         try {
             const result_values = await mi.sendDataListRegisterValues(this.gdb, {fmt:' x'});
             const reg_values = result_values['register-values'];
-            //const rv_1 = Object.values(reg_values_cv[0]);
-            //const rv_2 = Object.values(rv_1);
             for (const n of reg_values){
                 const id = Object.values(n)[0];
-                //const id = n_values[0];
                 const reg = this.registerMapReverse.get(parseInt(id));
                 if (reg) {
                     const val = n.value;
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1830,7 +1830,7 @@ export class GDBDebugSession extends LoggingDebugSession {
             if (this.registerMap.size === 0) {
                 const result_names = await mi.sendDataListRegisterNames(
                     this.gdb,
-                    {frameId: frame.frameId, threadId: frame.threadId}
+                    { frameId: frame.frameId, threadId: frame.threadId }
                 );
                 let idx = 0;
                 const registerNames = result_names['register-names'];
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -883,13 +883,17 @@ export class GDBDebugSession extends LoggingDebugSession {
 
         const registers: RegisterVariableReference = {
             type: 'registers',
-            frameHandle: args.frameId
+            frameHandle: args.frameId,
         };
 
         response.body = {
             scopes: [
                 new Scope('Local', this.variableHandles.create(frame), false),
-                new Scope('Registers', this.variableHandles.create(registers), true),
+                new Scope(
+                    'Registers',
+                    this.variableHandles.create(registers),
+                    true
+                ),
             ],
         };
 
@@ -911,13 +915,15 @@ export class GDBDebugSession extends LoggingDebugSession {
                 return;
             }
             if (ref.type === 'registers') {
-                response.body.variables = await this.handleVariableRequestRegister(ref);
-            }
-            else if (ref.type === 'frame') {
-                response.body.variables = await this.handleVariableRequestFrame(ref);
-
+                response.body.variables =
+                    await this.handleVariableRequestRegister(ref);
+            } else if (ref.type === 'frame') {
+                response.body.variables = await this.handleVariableRequestFrame(
+                    ref
+                );
             } else if (ref.type === 'object') {
-                response.body.variables = await this.handleVariableRequestObject(ref);
+                response.body.variables =
+                    await this.handleVariableRequestObject(ref);
             }
             this.sendResponse(response);
         } catch (err) {
@@ -1807,42 +1813,47 @@ export class GDBDebugSession extends LoggingDebugSession {
     }
 
     // Register view
-
+    // Assume that the register name are unchanging over time, and the same across all threadsf
     private registerMap = new Map<string, number>();
     private registerMapReverse = new Map<number, string>();
-    protected async handleVariableRequestRegister(ref: RegisterVariableReference): Promise<DebugProtocol.Variable[]> {
+    protected async handleVariableRequestRegister(
+        ref: RegisterVariableReference
+    ): Promise<DebugProtocol.Variable[]> {
         // initialize variables array and dereference the frame handle
         const variables: DebugProtocol.Variable[] = [];
         const frame = this.frameHandles.get(ref.frameHandle);
         if (!frame) {
             return Promise.resolve(variables);
         }
-        
+
         try {
             if (this.registerMap.size === 0) {
-                ////// Refer to e2s plugin to port request code
-                const result_names = await mi.sendDataListRegisterNames(this.gdb, {});
+                const result_names = await mi.sendDataListRegisterNames(
+                    this.gdb,
+                    {frameId: frame.frameId, threadId: frame.threadId}
+                );
                 let idx = 0;
                 const registerNames = result_names['register-names'];
-                for (const regs of registerNames){
-                    if(regs !== ''){
+                for (const regs of registerNames) {
+                    if (regs !== '') {
                         this.registerMap.set(regs, idx);
                         this.registerMapReverse.set(idx, regs);
                     }
                     idx++;
-                }      
-            }            
-        }
-        catch {
-
-           throw new Error('Unable to parse response for reg. names ');
+                }
+            }
+        } catch {
+            throw new Error('Unable to parse response for reg. names ');
         }
 
         try {
-            const result_values = await mi.sendDataListRegisterValues(this.gdb, {fmt:' x'});
+            const result_values = await mi.sendDataListRegisterValues(
+                this.gdb,
+                { fmt: ' x' }
+            );
             const reg_values = result_values['register-values'];
-            for (const n of reg_values){
-                const id = Object.values(n)[0];
+            for (const n of reg_values) {
+                const id = n.number;
                 const reg = this.registerMapReverse.get(parseInt(id));
                 if (reg) {
                     const val = n.value;
@@ -1850,18 +1861,17 @@ export class GDBDebugSession extends LoggingDebugSession {
                         name: reg,
                         evaluateName: '$' + reg,
                         value: val,
-                        variablesReference: 0
+                        variablesReference: 0,
                     };
                     variables.push(res);
                 } else {
                     throw new Error('Unable to parse response for reg. values');
                 }
             }
-        }
-        catch (error) {
+        } catch (error) {
             throw new Error('Unable to parse response for reg. names');
         }
-        
+
         return Promise.resolve(variables);
     }
 
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1844,6 +1844,8 @@ export class GDBDebugSession extends LoggingDebugSession {
 
         const result_values = await mi.sendDataListRegisterValues(this.gdb, {
             fmt: ' x',
+            frameId: frame.frameId,
+            threadId: frame.threadId,
         });
         const reg_values = result_values['register-values'];
         for (const n of reg_values) {
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1826,50 +1826,41 @@ export class GDBDebugSession extends LoggingDebugSession {
             return Promise.resolve(variables);
         }
 
-        try {
-            if (this.registerMap.size === 0) {
-                const result_names = await mi.sendDataListRegisterNames(
-                    this.gdb,
-                    { frameId: frame.frameId, threadId: frame.threadId }
-                );
-                let idx = 0;
-                const registerNames = result_names['register-names'];
-                for (const regs of registerNames) {
-                    if (regs !== '') {
-                        this.registerMap.set(regs, idx);
-                        this.registerMapReverse.set(idx, regs);
-                    }
-                    idx++;
+        if (this.registerMap.size === 0) {
+            const result_names = await mi.sendDataListRegisterNames(this.gdb, {
+                frameId: frame.frameId,
+                threadId: frame.threadId,
+            });
+            let idx = 0;
+            const registerNames = result_names['register-names'];
+            for (const regs of registerNames) {
+                if (regs !== '') {
+                    this.registerMap.set(regs, idx);
+                    this.registerMapReverse.set(idx, regs);
                 }
+                idx++;
             }
-        } catch {
-            throw new Error('Unable to parse response for reg. names ');
         }
 
-        try {
-            const result_values = await mi.sendDataListRegisterValues(
-                this.gdb,
-                { fmt: ' x' }
-            );
-            const reg_values = result_values['register-values'];
-            for (const n of reg_values) {
-                const id = n.number;
-                const reg = this.registerMapReverse.get(parseInt(id));
-                if (reg) {
-                    const val = n.value;
-                    const res: DebugProtocol.Variable = {
-                        name: reg,
-                        evaluateName: '$' + reg,
-                        value: val,
-                        variablesReference: 0,
-                    };
-                    variables.push(res);
-                } else {
-                    throw new Error('Unable to parse response for reg. values');
-                }
+        const result_values = await mi.sendDataListRegisterValues(this.gdb, {
+            fmt: ' x',
+        });
+        const reg_values = result_values['register-values'];
+        for (const n of reg_values) {
+            const id = n.number;
+            const reg = this.registerMapReverse.get(parseInt(id));
+            if (reg) {
+                const val = n.value;
+                const res: DebugProtocol.Variable = {
+                    name: reg,
+                    evaluateName: '$' + reg,
+                    value: val,
+                    variablesReference: 0,
+                };
+                variables.push(res);
+            } else {
+                throw new Error('Unable to parse response for reg. values');
             }
-        } catch (error) {
-            throw new Error('Unable to parse response for reg. names');
         }
 
         return Promise.resolve(variables);
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -70,9 +70,16 @@ export interface ObjectVariableReference {
     varobjName: string;
 }
 
+export interface RegisterVariableReference {
+    type: 'registers';
+    frameHandle: number;
+    regname?: string;
+}
+
 export type VariableReference =
     | FrameVariableReference
-    | ObjectVariableReference;
+    | ObjectVariableReference
+    | RegisterVariableReference;
 
 export interface MemoryRequestArguments {
     address: string;
@@ -874,9 +881,15 @@ export class GDBDebugSession extends LoggingDebugSession {
             frameHandle: args.frameId,
         };
 
+        const registers: RegisterVariableReference = {
+            type: 'registers',
+            frameHandle: args.frameId
+        };
+
         response.body = {
             scopes: [
                 new Scope('Local', this.variableHandles.create(frame), false),
+                new Scope('Registers', this.variableHandles.create(registers), true),
             ],
         };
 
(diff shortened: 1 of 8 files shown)


src/GDBDebugSession.ts
```diff
@@ -874,9 +881,15 @@ export class GDBDebugSession extends LoggingDebugSession {
             frameHandle: args.frameId,
         };
 
+        const registers: RegisterVariableReference = {
+            type: 'registers',
+            frameHandle: args.frameId
```
jonahgraham, line 886: Missing trailing comma - please run `yarn format` on the code. (other code isn't formatted properly either, but the one call to `yarn format` should fix that.

src/GDBDebugSession.ts
```diff
@@ -1792,6 +1806,65 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+        // initialize variables array and dereference the frame handle
+        const variables: DebugProtocol.Variable[] = [];
+        const frame = this.frameHandles.get(ref.frameHandle);
+        if (!frame) {
+            return Promise.resolve(variables);
+        }
+        
+        try {
+            if (this.registerMap.size === 0) {
+                ////// Refer to e2s plugin to port request code
```
jonahgraham, line 1823: Internal e2s comment doesn't belong here.

src/GDBDebugSession.ts
```diff
@@ -1792,6 +1806,65 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+                for (const regs of registerNames){
+                    if(regs !== ''){
+                        this.registerMap.set(regs, idx);
+                        this.registerMapReverse.set(idx, regs);
+                    }
+                    idx++;
+                }      
+            }            
+        }
+        catch {
```
jonahgraham, line 1836: Don't discard the exception information, use it so that the source of the problem can be tracked down later.

src/GDBDebugSession.ts
```diff
@@ -1792,6 +1806,65 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+
+           throw new Error('Unable to parse response for reg. names ');
+        }
+
+        try {
+            const result_values = await mi.sendDataListRegisterValues(this.gdb, {fmt:' x'});
+            const reg_values = result_values['register-values'];
+            for (const n of reg_values){
+                const id = Object.values(n)[0];
```
jonahgraham, line 1845: Instead of `Object.values(n)[0]` should this be `n.number`?

src/GDBDebugSession.ts
```diff
@@ -1792,6 +1806,65 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+                        value: val,
+                        variablesReference: 0
+                    };
+                    variables.push(res);
+                } else {
+                    throw new Error('Unable to parse response for reg. values');
+                }
+            }
+        }
+        catch (error) {
```
jonahgraham, line 1861: Don't discard the exception information, use it so that the source of the problem can be tracked down later.

src/mi/stack.ts
```diff
@@ -98,3 +106,36 @@ export function sendStackListVariables(
 
     return gdb.sendCommand(command);
 }
+
+export function sendDataListRegisterNames(gdb: GDBBackend, params: {
```
jonahgraham, line 110: To be consistent with how GDB organizes the commands (See https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Data-Manipulation.html) these new methods should be in data.ts.

src/mi/stack.ts
```diff
@@ -98,3 +106,36 @@ export function sendStackListVariables(
...
+    regno?: number[];
+    threadId?: number;
+}): Promise<MIListRegisterValuesResponse> {
+    let command = '-data-list-register-values';
+    command += `${params.fmt}`;
+    if (params.threadId) {
+        command += ` --thread ${params.threadId}`;
+    }
+    if (params.regno) {
+        for (let i = 0; i<params.regno.length; i++) {
```
jonahgraham, line 136: Use a string joining, I think something like `command += params.regno.join(' ')` can replace this for loop.

jonahgraham (changes requested):
This looks good - I have tried it out and it worked fine.

Please add a test, look at mem.spec.ts as a possible starting point. Because the test should be platform independent, avoid hardcoding specific registers names in the test, instead test with some heuristics like making sure that non-empty values are returned from the adapter.

src/GDBDebugSession.ts
quyettrinhrvc: Updated format

src/GDBDebugSession.ts
quyettrinhrvc: Removed redundant comment

src/GDBDebugSession.ts
quyettrinhrvc: Sorry but I cannot catch your idea here.
Could you please explain more?

src/mi/stack.ts
```diff
@@ -98,3 +106,36 @@ export function sendStackListVariables(
...
+        command += ` ${params.regno}`;       
+    }
+    //throw new Error('send data register name check!');
+    return gdb.sendCommand(command);
+}
+
+export function sendDataListRegisterValues(gdb: GDBBackend, params: {
+    fmt: string;
+    regno?: number[];
+    threadId?: number;
```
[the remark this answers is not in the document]
quyettrinhrvc: I updated for "sendDataListRegisterNames" with thread ID + frame ID and it is OK.
But for "sendDataListRegisterValues", it will not show full list of register.
I refer to [this](https://sourceware.org/gdb/onlinedocs/gdb/GDB_002fMI-Data-Manipulation.html) and saw that there is no thread ID and frame ID for this command. How do you think?

src/GDBDebugSession.ts
quyettrinhrvc: Updated this point

src/GDBDebugSession.ts
quyettrinhrvc: Sorry but I cannot catch your idea here.
Could you please explain more?

src/mi/stack.ts
quyettrinhrvc: Moved these functions to data.ts

src/mi/stack.ts
```diff
@@ -98,3 +106,36 @@ export function sendStackListVariables(
...
+}): Promise<MIListRegisterNamesResponse> {
+    let command = '-data-list-register-names';
+    if (params.threadId) {
+        command += ` --thread ${params.threadId}`;
+    }
+    if (params.regno) {
+        command += ` ${params.regno}`;       
+    }
+    //throw new Error('send data register name check!');
```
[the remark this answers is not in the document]
quyettrinhrvc: Remove redundant comment

src/mi/stack.ts
quyettrinhrvc: Updated this point

src/GDBDebugSession.ts
```diff
@@ -1792,6 +1812,69 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+                }
+            }
+        } catch {
+            throw new Error('Unable to parse response for reg. names ');
+        }
+
+        try {
+            const result_values = await mi.sendDataListRegisterValues(
+                this.gdb,
+                { fmt: ' x' }
```
jonahgraham, line 1852: There is another comment about this - but just marking the exact location that thread/frame needs to be supplied to `sendDataListRegisterValues`

src/mi/data.ts
```diff
@@ -110,3 +118,45 @@ export async function sendDataDisassemble(
...
+    gdb: GDBBackend,
+    params: {
+        regno?: number[];
+        frameId: number;
+        threadId: number;
+    }
+): Promise<MIListRegisterNamesResponse> {
+    let command = '-data-list-register-names';
+    if (params.regno) {
+        command += ` ${params.regno}`;
```
jonahgraham, line 132: This line won't work, like you have done below you need to join this array.

However, alternatively (and maybe for the best) remove regno as an argument, your code doesn't use it anyway, so having the extra code here isn't that useful.

src/mi/data.ts
```diff
@@ -110,3 +118,45 @@ export async function sendDataDisassemble(
...
+    }
+    command += ` --frame ${params.frameId}`;
+    if (params.threadId) {
```
jonahgraham, line 135: This if is somewhat problematic in javasript/typescript as the number `0` is false. As threadId is required, simply remove the `if` and add `--thread` always. There are a few other places where you have an if that may be problematic, but as you make arguments required you can just remove the `if`s there too.

jonahgraham (changes requested):
Thanks! Some minor changes, the big thing needed is a test for this still. Let me know if you want some additional guidance on how to write such a test

src/GDBDebugSession.ts
quyettrinhrvc: Since "handleVariableRequestRegister" is called in "variablesRequest", and in this function, the error information is returned after handled "handleVariableRequestRegister". So can we remove the "catch (error)" here? How do you think?

src/mi/stack.ts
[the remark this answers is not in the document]
quyettrinhrvc: When I added frame ID and thread ID to "-data-list-register-values" to get full list of registers and its values
image
image
Currently, it only return some registers but not full list.
Here is what we get without frame ID and thread ID
image
image
Could you please help me check it?

src/GDBDebugSession.ts
quyettrinhrvc: Same as above. I think we can remove "try ... catch(error)" here.

quyettrinhrvc:
> Thanks! Some minor changes, the big thing needed is a test for this still. Let me know if you want some additional guidance on how to write such a test
Honestly I have no more experiences in this testing and I am investigating about it.
Currently, I am referring to var.spec.ts to add test for register (it also is a part of scopeRequest, same as register view).
Is it OK as a reference?
And yes, it is really good if you can share some additional documents or guidance on how to write such a test.
Thank you in advance!

src/GDBDebugSession.ts
jonahgraham: That is a good idea.

src/GDBDebugSession.ts
jonahgraham: Good idea.

jonahgraham:
A couple of the failures are tests that need updating because of this change. Most of them look like Windows issues. Please run the tests without your change to identify which ones are caused by your change. The memory + target failures are probably just windows issues and we need to make tests on Windows more resilient. Please let me know which version of gdb + gdbserver are in use and I can replicate and resolve those failures independent of your work - do this by creating a new issue in this repo with those details.

As for general advice, look at https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/integration-tests/README.md - the docker running is useful as it can avoid the problems of getting the expected tool versions.

For specific advice on your test here is a test method to add to var.spec.ts:

```typescript
    it('can read registers in a program', async function () {
        // read the registers
        const vr = scope.scopes.body.scopes[1].variablesReference;
        const vars = await dc.variablesRequest({ variablesReference: vr });
        expect(
            vars.body.variables.length,
            'There is a different number of variables than expected'
        ).to.be.greaterThanOrEqual(5); // 5 is a good bet to make sure that code has probably worked
        const r0 = vars.body.variables[0];
        const r1 = vars.body.variables[1];
        // can't check specific names or register values easily as that
        // is not cross platform
        expect(r0.evaluateName).to.startWith('$');
        expect(r0.name).to.not.equal(r1.name);
        // add other useful tests here, especially ones that test boundary conditions
    });
```

That test has some additional dependencies, so add them to the project with:

```sh
$ yarn add chai-string @types/chai-string
```

Then you can add the extra imports too:
```typescript
import * as chai from 'chai';
import * as chaistring from 'chai-string';
chai.use(chaistring);
```

Finally, don't forget to fix the # of scopes test:
```diff
diff --git a/src/integration-tests/var.spec.ts b/src/integration-tests/var.spec.ts
index df6a71d..b358bc2 100644
--- a/src/integration-tests/var.spec.ts
+++ b/src/integration-tests/var.spec.ts
@@ -59,7 +62,7 @@ describe('Variables Test Suite', function () {
         expect(
             scope.scopes.body.scopes.length,
             'Unexpected number of scopes returned'
-        ).to.equal(1);
+        ).to.equal(2);
     });
 
     afterEach(async function () {
```

src/mi/stack.ts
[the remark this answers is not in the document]
jonahgraham: I believe the issue here is that `--thread` and `--frame` need to appear first:

```sh
-data-list-register-values --thread 1 --frame 0 x 1
^done,register-values=[{number="1",value="0x5555555551c0"}]
(gdb) 
-data-list-register-values x --thread 1 --frame 0 1
^done,register-values=[{number="0",value="0x55555555513d"},{number="1",value="0x5555555551c0"},{number="0",value="0x55555555513d"},{number="0",value="0x55555555513d"},{number="1",value="0x5555555551c0"}]
```

GDB follows the standard convention of `-` flags needing to come before positional arguments. As the format is positional, and `-` or `--` flags should come before it.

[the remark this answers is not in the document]
quyettrinhrvc: Thanks for your advice!
I updated the convention and it is OK now.

src/mi/data.ts
```diff
@@ -110,3 +118,48 @@ export async function sendDataDisassemble(
...
+        threadId: number;
+    }
+): Promise<MIListRegisterValuesResponse> {
+    let command = '-data-list-register-values';
+
+    command += ` --frame ${params.frameId}`;
+
+    command += ` --thread ${params.threadId}`;
+
+    command += `${params.fmt}`;
```
jonahgraham, line 158: Avoid the unneeded interpolated string, this line can just be:
```typescript
command += params.fmt;
```

src/GDBDebugSession.ts
```diff
@@ -1792,6 +1812,62 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+                if (regs !== '') {
+                    this.registerMap.set(regs, idx);
+                    this.registerMapReverse.set(idx, regs);
+                }
+                idx++;
+            }
+        }
+
+        const result_values = await mi.sendDataListRegisterValues(this.gdb, {
+            fmt: ' x',
```
jonahgraham, line 1846: The space here is a separator - the space should be added by sendDataListRegisterValues

src/mi/data.ts
jonahgraham: The above will become irrelevant depending on how you resolve my comment on 1846.

PS now that you have made this all required, you can also write this whole method a little more succiently like this:

```typescript
let command = `-data-list-register-values --thread ${params.threadId} --frame ${params.frameId} ${params.fmt}`
```

But that is a style issue and not terribly important.

jonahgraham (changes requested):
This is very good progress. Once the tests are in I think this will be ready to go.

quyettrinhrvc:
Sorry for my late reply.
> A couple of the failures are tests that need updating because of this change. Most of them look like Windows issues. Please run the tests without your change to identify which ones are caused by your change. The memory + target failures are probably just windows issues and we need to make tests on Windows more resilient. Please let me know which version of gdb + gdbserver are in use and I can replicate and resolve those failures independent of your work - do this by creating a new issue in this repo with those details.
> 
> As for general advice, look at https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/integration-tests/README.md - the docker running is useful as it can avoid the problems of getting the expected tool versions.
> 
> For specific advice on your test here is a test method to add to var.spec.ts:
> 
> ```ts
>     it('can read registers in a program', async function () {
>         // read the registers
>         const vr = scope.scopes.body.scopes[1].variablesReference;
>         const vars = await dc.variablesRequest({ variablesReference: vr });
>         expect(
>             vars.body.variables.length,
>             'There is a different number of variables than expected'
>         ).to.be.greaterThanOrEqual(5); // 5 is a good bet to make sure that code has probably worked
>         const r0 = vars.body.variables[0];
>         const r1 = vars.body.variables[1];
>         // can't check specific names or register values easily as that
>         // is not cross platform
>         expect(r0.evaluateName).to.startWith('$');
>         expect(r0.name).to.not.equal(r1.name);
>         // add other useful tests here, especially ones that test boundary conditions
>     });
> ```
> 
> That test has some additional dependencies, so add them to the project with:
> 
> ```shell
> $ yarn add chai-string @types/chai-string
> ```
> 
> Then you can add the extra imports too:
> 
> ```ts
> import * as chai from 'chai';
> import * as chaistring from 'chai-string';
> chai.use(chaistring);
> ```
> 
> Finally, don't forget to fix the # of scopes test:
> 
> ```diff
> diff --git a/src/integration-tests/var.spec.ts b/src/integration-tests/var.spec.ts
> index df6a71d..b358bc2 100644
> --- a/src/integration-tests/var.spec.ts
> +++ b/src/integration-tests/var.spec.ts
> @@ -59,7 +62,7 @@ describe('Variables Test Suite', function () {
>          expect(
>              scope.scopes.body.scopes.length,
>              'Unexpected number of scopes returned'
> -        ).to.equal(1);
> +        ).to.equal(2);
>      });
>  
>      afterEach(async function () {
> ```

Thanks for instruction. I understood how to create test program.
However, I am having an issue related to test execution.

> Please let me know which version of gdb + gdbserver are in use and I can replicate and resolve those failures independent of your work - do this by creating a new issue in this repo with those details.

Sorry but how can I add gdb+gdbserver to the test? Could you please teach me?

I tried to execute the fresh version from github and got this result
```
<testsuites name="Mocha Tests">
  <testsuite name="Non-test failures" tests="1" errors="0" failures="1" skipped="0" timestamp="2022-05-09T08:43:46" time="0.003">
    <testcase classname="Non-test failures" name="&quot;before all&quot; hook in &quot;{root}&quot;" time="38.861">
      <failure message="Timeout of 20000ms exceeded. For async tests and hooks, ensure &quot;done()&quot; is called; if returning a Promise, ensure it resolves.">    at Context.&lt;anonymous&gt; (dist\integration-tests\utils.js:112:5)
    at processImmediate (node:internal/timers:464:21)</failure>
    </testcase>
  </testsuite>
</testsuites>
```
image
It seems that testProgramsDir is not correct, I think.
Do you have any idea about it?

src/GDBDebugSession.ts
quyettrinhrvc: Updated

quyettrinhrvc:
```
C:\Users\quyettrinh>gdb --version
GNU gdb (GDB) 7.4
Copyright (C) 2012 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.  Type "show copying"
and "show warranty" for details.
This GDB was configured as "i686-pc-mingw32".
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.

C:\Users\quyettrinh>gdbserver --version
GNU gdbserver (GDB) 7.4
Copyright (C) 2012 Free Software Foundation, Inc.
gdbserver is free software, covered by the GNU General Public License.
This gdbserver was configured as "i686-pc-mingw32"

C:\Users\quyettrinh>make --version
GNU Make 4.2.1
Built for x86_64-unknown-cygwin
Copyright (C) 1988-2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

C:\Users\quyettrinh>g++ --version
g++ (GCC) 4.6.2
Copyright (C) 2011 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
Here is current version of my tools in Windows environment.
So it means that the test will execute with these gdb/gdbserver?

jonahgraham:
> So it means that the test will execute with these gdb/gdbserver?

Yes. That gdb are quite old and developing with newer gdb will be much better. I recommend [msys2](https://www.msys2.org/). I can put together step-by-step instructions for how to setup msys2 if needed. Let me know. There are other options, like [cygwin](https://www.cygwin.com/) and [chocolatey](https://chocolatey.org/), but I use msys2 with the [mingw 64-bit toolchain](https://www.mingw-w64.org/). A number of people use [mingw](https://en.wikipedia.org/wiki/MinGW) (note lack of 64 in name) but I think that is very dated now and I would not recommend it. For most of my heavy development I use Linux, only using Windows to verify behaviour.

quyettrinhrvc:
> I can put together step-by-step instructions for how to setup msys2 if needed. Let me know. 

Yes, please share me guidance if any.
Thank you in advance

quyettrinhrvc:
> > It seems that testProgramsDir is not correct, I think. Do you have any idea about it?
> 
> I am unsure - I don't know how testProgramsDir could be incorrect.
> 
> I haven't verified the tests are in good operation on Windows recently as Linux is my main development environment. I will confirm and get back to you.

I found reason for this issue. Yes, it is not problem with `testProgramsDir`. I used wrong command in makefile so it deleted all built objects and cannot run test. It's OK now.
Btw, I tried to run test with #191 and got this result.
[integration#191.zip](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/files/8657146/integration.191.zip)
Some failures are solved but there are some new failures, could you please check it?

quyettrinhrvc:
I updated var.spec.ts and vars_cpp.spec.ts for register test, and I got this result
[integrationMay10.zip](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/files/8660749/integrationMay10.zip)
Currently, it seems that there is failures of environments only (old version of gdb, gdbserver, etc.).
Would you please help me check it?

jonahgraham:
> I think this happens because the `Array.at` method is very [new](https://tc39.es/ecma262/#sec-array.prototype.at) in javascript.

The problem was we were building against node 17, but we need to support node 14. I fixed this in b4bd5c9bbe5261c25c2fb0ff1c9f62375af58143

> In the meantime please rewrite the test to not use the `Array.at` method.

@quyettrinhrvc I decided to just do this and made a couple of small fixes (934f033fe8a237cdfda35c817c8dc66a0e72cc07 and 1a1041e681fe0f96b23d8d8c564c8afc57da224c) and merged this change. It is now available in npm so you can update your dependency version, see details on version [0.0.16-next.20220511231930.4eac625.0](https://www.npmjs.com/package/cdt-gdb-adapter/v/0.0.16-next.20220511231930.4eac625.0)

Congratulations :tada: on your first merged PR!

quyettrinhrvc:
> Congratulations 🎉 on your first merged PR!

Thank you! 😄 

> I will get back to you soon on setting up Windows host using msys2.

OK, I will wait for this.

quyettrinhrvc:
Hi @jonahgraham ,
I found some guidelines to install mysys2 on RVC Jira.
Is it same as yours?

> Which package manager are you using?
> 
> I recommend [msys2](https://www.msys2.org/) on Windows, but you may have reasons to use others. If you do use another, make sure it is up to date. If it reports up to date, but still give gdb 7.6, that means the whole package manager is probably not actively maintained anymore.
> 
> I use something like these setup instructions on Windows - but I haven't run through this list in a while so may need some massaging. See the msys2 for help https://www.msys2.org/wiki/MSYS2-installation/
> 
> Get msys2 [http://www.msys2.org/ ](http://www.msys2.org/)and install
> Start MSYS2 MSYS (from installer or start menu)
> pacman -Syu # update core packages
> close shell if instructed to, see instructions in shell
> start MSYS2 MSYS again (from start menu)
> pacman -Su # update remaining packages
> pacman -S base-devel # choose all to get all normal dev tools
> {{}}pacman -S mingw-w64-x86_64-toolchain # get all tools, gcc, gdb, etc
> MSYS2 MinGW 64-bit (from start menu) – this is the shell to build from 
> gcc --version -v # show details about compiler being used
> gdb --version # show details about gdb being used
> https://chocolatey.org/{{}}

If there is any change points, please inform me!

jonahgraham:
@quyettrinhrvc - I added details to https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/integration-tests/README.md#setting-up-test-environmnt-on-windows-host - it is based on the same info that you quoted.

quyettrinhrvc:
> @quyettrinhrvc - I added details to https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/integration-tests/README.md#setting-up-test-environmnt-on-windows-host - it is based on the same info that you quoted.

It's very helpful for us.
Thank you very much!

PR #194, opened 2022-06-17
mi-async and non-stop modes

### LINKED ISSUE #28: Use GDB's mi-async mode

jonahgraham:
This adds mi-async mode (on by default where supported) and non-stop mode (off by default). Fixes #28

It includes new parameters to override the defaults, see https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/69 for more info on that part.


src/GDBBackend.ts
```diff
@@ -98,15 +103,63 @@ export class GDBBackend extends events.EventEmitter {
...
-    public pause() {
-        if (this.proc) {
-            this.proc.kill('SIGINT');
-            return true;
+    public async setAsyncMode(isSet?: boolean) {
+        const command = this.gdbVersionAtLeast('7.8')
+            ? 'mi-async'
+            : 'target-async';
+        if (isSet === undefined) {
+            isSet = true;
```
quyettrinhrvc, line 116: I think that default mode should be synchronous mode.
But here asynchronous mode is set as default. Is it correct?
Please refer [here](https://sourceware.org/gdb/onlinedocs/gdb/Asynchronous-and-non_002dstop-modes.html#Asynchronous-and-non_002dstop-modes).

src/GDBBackend.ts
```diff
@@ -98,15 +103,63 @@ export class GDBBackend extends events.EventEmitter {
...
+            this.gdbNonStop = false;
+        }
+    }
+
+    public isNonStopMode() {
+        return this.gdbNonStop;
+    }
+
+    public pause(threadId?: number) {
+        if (this.gdbAsync || this.gdbNonStop) {
```
quyettrinhrvc, line 155: As my investigate, the MI execution commands always executed asynchronously in Non-stop mode.
So do we need to check async here? How do you think?
https://sourceware.org/gdb/onlinedocs/gdb/Non_002dStop-Mode.html#Non_002dStop-Mode

src/integration-tests/test-programs/loopforever.c
```diff
@@ -0,0 +1,29 @@
...
+volatile int var1 = 0;
+volatile int var2 = 0;
+volatile int stop = 0;
+
+
+int inner1(void) {
+    return var1++;
+}
+int inner2(void) {
+    return var1++;
```
quyettrinhrvc, line 12: Should it be "var2"?

src/GDBBackend.ts
```diff
@@ -98,15 +103,63 @@ export class GDBBackend extends events.EventEmitter {
...
+        await this.setAsyncMode(requestArgs.gdbAsync);
+        await this.setNonStopMode(requestArgs.gdbNonStop);
     }
 
-    public pause() {
-        if (this.proc) {
-            this.proc.kill('SIGINT');
-            return true;
+    public async setAsyncMode(isSet?: boolean) {
+        const command = this.gdbVersionAtLeast('7.8')
```
quyettrinhrvc, line 112: How can we know that the command is 'mi-async' for gdb version > 7.8 and 'target-async' for < 7.8?
I checked with Renesas gdb and saw that version is 7.8.2 but command still be 'target-async'.
Could you please confirm it?

src/GDBBackend.ts
```diff
@@ -98,15 +103,63 @@ export class GDBBackend extends events.EventEmitter {
...
+        }
+    }
+
+    public async setNonStopMode(isSet?: boolean) {
+        if (isSet === undefined) {
+            isSet = false;
+        }
+        const onoff = isSet ? 'on' : 'off';
+        try {
+            await this.sendCommand(`-gdb-set non-stop ${onoff}`);
```
quyettrinhrvc, line 138: Should we send "set pagination off" before sending this command?
https://sourceware.org/gdb/onlinedocs/gdb/Non_002dStop-Mode.html#Non_002dStop-Mode and current implementation in Renesas adapter have this command.

quyettrinhrvc (commented):
@jonahgraham I checked your modification and have some comments.
Could you please confirm them?

src/GDBBackend.ts
jonahgraham: GDB indeed defaults to off here, but we want it on so that by default users can do interrupt, especially on Windows host. This matches what CDT does.

src/GDBBackend.ts
jonahgraham: I was wondering how to resolve this check. Is GDB actually in mi-async mode when in non-stop mode, or is it a hybrid async mode that is async only for some stuff.

If it is the former, then if non-stop is set to on in setNonStopMode, then this.gdbAsync should also be set to true. WDYT?

src/integration-tests/test-programs/loopforever.c
jonahgraham: Yes. Done.

src/GDBBackend.ts
jonahgraham: Can you check if Renesas GDB supports both? If not, we can adjust this range. From 7.8 to gdb 10 they are aliases of each other, but starting around gdb 11 using the old one generates a warning:

```
-gdb-set target-async on
~"Warning: 'set target-async', an alias for the command 'set mi-async', is deprecated.\n"
~"Use 'set mi-async'.\n\n"
^done
```

src/GDBBackend.ts
jonahgraham: That is a good idea. Done in next udpate.

src/GDBBackend.ts
jonahgraham: The latest patchset does this. Note the updated docs in the vscode PR: https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/69

jonahgraham:
This is now available in npm: https://www.npmjs.com/package/cdt-gdb-adapter/v/0.0.16-next.20220620151247.50e38d3.0

src/GDBDebugSession.ts
```diff
@@ -1469,16 +1493,55 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+            for (const thread of this.threads) {
+                if (!thread.running) {
+                    this.isRunning = false;
+                }
+            }
+        };
         switch (resultClass) {
             case 'running':
-                this.isRunning = true;
+                if (this.gdb.isNonStopMode()) {
```
quyettrinhrvc, line 1506: @jonahgraham
FYI, when we tried to used latest cdt with renesas cdt, the debug process is stucked. As my understanding, the async mode is not set correctly.
I tried to debug and saw that at this point, gdbAsync and gdbNonStop is not same as configuration in launch.json.
Their values are updated in later step (when call setAsyncMode and setNonStopMode), so in here, they still keep the default value (false).
AsyncIssue
Sorry but is it same as your design expectation? If my understanding is wrong, please help to correct it.

jonahgraham: @quyettrinhrvc the setAsyncMode and setNonStopMode calls are the first things done after launching GDB. However GDB may be sending async notifications _before_ that point. Is that what is happening? Can you share the MI log? Is this related to the unresolved conversation above about which GDB version supports `mi-async` vs `target-async`?

quyettrinhrvc: @jonahgraham please ignore my previous comment.
I found the reason why I cannot debug with renesas adapter and solved it. (I forgot to update `isInitialized ` to `true` in reneasas adapter)
I think that current implementation of cdt adapter is OK now.
We are performing some tests with renesas devices. If there is any issue, I will inform you.
Sorry for this inconvenience!

src/GDBBackend.ts
quyettrinhrvc: Yes, you are right. I checked and current Renesas GDB supports both.

PR #195, opened 2022-06-21
Fix windows multithread cases

jonahgraham:
Follow on work for #28 this fixes errors identified that affects Windows only, including most critically making sure the thread/frame is specified to `-var-create`.

The multithread test, which access variables on different threads and frames verifies the var create is done on the correct thread/frame.


quyettrinhrvc (commented):
@jonahgraham 
Sorry for my lateness.
I checked this pull request and it is OK. I have no comment now.

PR #202, opened 2022-09-08
Improve parser regex performance

colin-grant-work:
If a large message comes in many parts, the current code of the parser is very inefficient: it rechecks the entire buffer, even though when a new chunk arrives, we can guarantee that the regex cannot match the existing buffer, and it uses a capture that will always just capture everything from the beginning of the string to the point of a line break.

This PR addresses both of those deficiencies:
 - When a new chunk arrives, only the new chunk is checked for a newline sequence.
 - No captures are used; instead the `index` field of the result is used to pick the end of the slice sent for parsing.

### DIFF
--- a/src/MIParser.ts
+++ b/src/MIParser.ts
@@ -22,19 +22,21 @@ export class MIParser {
     public parse(stream: Readable): Promise<void> {
         return new Promise((resolve) => {
             this.waitReady = resolve;
-            const lineRegex = /(.*)(\r?\n)/;
+            const lineBreakRegex = /\r?\n/;
             let buff = '';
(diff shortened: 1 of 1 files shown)


jonahgraham (approved):
LGTM - just running tests locally if they pass I will push and it should be published soon after to npm.

jonahgraham:
Its up npm: https://www.npmjs.com/package/cdt-gdb-adapter/v/0.0.16-next.20220908234947.bdf98e3.0

Thanks @colin-grant-work for the performance improvement.

colin-grant-work:
Thanks, @jonahgraham!

PR #205, opened 2022-09-12
Update to support debugging CVEngine

thamho0902:


thamho0902:
We are supporting debugging for CVEngine. However, CVEngine doesn't use arg.program so we need to update the code.

src/GDBTargetDebugSession.ts
```diff
@@ -299,7 +301,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                 }
             }
             await this.gdb.sendCommands(args.preRunCommands);
-            this.sendEvent(new InitializedEvent());
+            if (args.program) {
+                this.sendEvent(new InitializedEvent());
```
jonahgraham, line 305: I don't think sending `InitializedEvent` is optional? Can you elaborate why this change?

src/GDBTargetDebugSession.ts
```diff
@@ -234,7 +234,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         try {
             this.isAttach = true;
             await this.spawn(args);
-            await this.gdb.sendFileExecAndSymbols(args.program);
+            if (args.program) {
```
jonahgraham, line 237: Curiously `args.program` is a required field at the moment. Were you planning on marking the `program` field optional? Or is the intention to pass in an empty string for `program`?

If it is changing to optional, then there are a few updates needed. The declaration needs the `?` added:
https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/3ea8dad6393b9f639f1cb36745f9e0df4070c395/src/GDBDebugSession.ts#L42

And the cdt-gdb-vscode needs to be updated to match. Although this change could be left out as for most users having it be required makes sense. https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/blob/a50dfcdbd8267c45d27e53d09bc4d1a5949ef27b/package.json#L49-L51

jonahgraham (changes requested):

src/GDBTargetDebugSession.ts
thamho0902: When configuring args.program for CVE's program or even without configuring args.program, when debugging, the program will report an error like the logfile below:

**GDB command:** 6 -target-select extended-remote 192.168.20.109:9876
**GDB result:** 6 connected
**To client:** {"seq":0,"type":"event","event":"output","body":{"category":"console","output":"connected to extended-remote target 192.168.20.109:9876"}}
**GDB command:** 7 set sysroot remote:/ 
**To client:** {"seq":0,"type":"event","event":"output","body":{"category":"log","output":"set sysroot remote:/ \n"}}
**To client:** {"seq":0,"type":"event","event":"output","body":{"category":"log","output":"warning: \"remote:\" is deprecated, use \"target:\" instead.\n"}}
**To client:** {"seq":0,"type":"event","event":"output","body":{"category":"log","output":"warning: sysroot set to \"target:/\".\n"}}
**GDB notify async:** cmd-param-changed,param="sysroot",value="target:/"
**GDB unhandled notify:** cmd-param-changed: {"param":"sysroot","value":"target:/"}
**GDB result:** 7 done
**To client:** {"seq":0,"type":"event","event":"initialized"}
**To client:** {"seq":0,"type":"response","request_seq":2,"command":"attach","success":true}
[From client: setFunctionBreakpoints({"breakpoints":[]})
**GDB command:** 8 -break-list
**GDB result:** 8 done,BreakpointTable={nr_rows="0",nr_cols="6",hdr=[{width="7",alignment="-1",col_name="number",colhdr="Num"},{width="14",alignment="-1",col_name="type",colhdr="Type"},{width="4",alignment="-1",col_name="disp",colhdr="Disp"},{width="3",alignment="-1",col_name="enabled",colhdr="Enb"},{width="10",alignment="-1",col_name="addr",colhdr="Address"},{width="40",alignment="2",col_name="what",colhdr="What"}],body=[]}
**To client:** {"seq":0,"type":"response","request_seq":3,"command":"setFunctionBreakpoints","success":true,"body":{"breakpoints":[]}}
**From client:** configurationDone(undefined)
**GDB command:** 9 -exec-continue
**GDB result:** 9 error,msg=**"The program is not being run."**
**To client:** {"seq":0,"type":"response","request_seq":4,"command":"configurationDone","success":false,"message":"The program is not being run.","body":{"error":{"id":100,"format":"The program is not being run.","showUser":true}}}
**From client:** disconnect({"restart":false})

src/GDBTargetDebugSession.ts
thamho0902: Currently, If we mark the `program` field optional then we will need to fix the code quite a bit.

So I think we should keep the `args.program` for GDBTargetDebugSession.ts and we will fix the **"The program is not being run"** error when debugging for CVEngine instead.

When debugging with CVEngine, the CVEngine starts with `startGDBAndAttachToTarget()`, then go to the function `configurationDoneRequest()` and send the command `await mi.sendExecContinue(this.gdb);` to GDB and the program shows an error **"The program is not being run"**. Since the CVEngine program is dependent on the ARM program, the CVE program will not be run at this time.

I found a workaround that we need to create a `CVEngine variable?:boolean;` in the GDBDebugSession.ts and need to configure it on the launch.json of CVEngine. If  in the launch.json configuration `"CVEngine": true;` then the program will not send the command `await mi.sendExecContinue(this.gdb);` of `configurationDoneRequest()` to GDB.

thamho0902:
Dear @jonahgraham,

Could you please help me check again?

Thanks a lot!

src/GDBDebugSession.ts
```diff
@@ -39,6 +39,7 @@ export interface RequestArguments extends DebugProtocol.LaunchRequestArguments {
     gdbArguments?: string[];
     gdbAsync?: boolean;
     gdbNonStop?: boolean;
+    CVEngine?: boolean;
```
jonahgraham, line 42: CVEngine is a Renesas specific term, not suitable for the generic adapter. It is fine to add a parameter to the adapter to provide more control with what happens during `configurationDoneRequest`

src/GDBDebugSession.ts
```diff
@@ -699,7 +702,9 @@ export class GDBDebugSession extends LoggingDebugSession {
     ): Promise<void> {
         try {
             if (this.isAttach) {
-                await mi.sendExecContinue(this.gdb);
+                if(!this.isCVEngine){
```
jonahgraham, line 705: This is exactly the place that it is suitable to override `configurationDoneRequest` in the vendor/Renesas specific code. Unlike the earlier example where large parts of the code were copied, the subclass should do something like:

```typescript
    protected async configurationDoneRequest(
        response: DebugProtocol.ConfigurationDoneResponse,
        _args: DebugProtocol.ConfigurationDoneArguments
    ): Promise<void> {
        if (this.isCVEngine) { // with isCVEngine being a field in the subclass too
            this.sendResponse(response);
        } else {
            super.configurationDoneRequest(response, _args);
        }
    }
```

jonahgraham (changes requested):
As is the code is not ok - I have suggested two ways of resolving your issue.

src/GDBDebugSession.ts
jonahgraham: PS There are a couple of minor formatting issues in and around this method - please make sure you run `yarn format` or setup [vscode to format on save](https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode). Note if that was the only issue with the PR I would just fix formatting and merge. But it would help if code arrived already formatted.

jonahgraham (changes requested):

src/GDBDebugSession.ts
thamho0902: Dear @jonahgraham-san,
So I can override `configurationDoneRequest` in RenesasGDBDebugSession.ts and we don't need extra code in cdt-gdb-apdapter right? If correct I will commit the code to go to renesas-gdb-adapter and close the pull request at cdt-gdb-adapter.

Thank you!

jonahgraham: Yes, that is correct.

thamho0902: Dear @jonahgraham -san,
I updated the code of Renesas-gdb-adapter in Gerrit.
Could you please help me review the code?

jonahgraham:
Thanks @thamho0902  - the solution of ` configurationDoneRequest` being in the subclass looks best.

PR #206, opened 2022-09-12
Run all tests in non-stop mode (in addition to all-stop)

jonahgraham:
Now that we support non-stop mode, run all the tests again in non-stop mode to make sure that they work equally well in non-stop as all-stop.


jonahgraham:
With this PR `yarn test` will run both non-stop and all-stop modes. To just test the non-stop you can do `yarn test:integration-gdb-non-stop`

jonahgraham:
As non-stop doesn't work on Windows these tests will fail there. Maybe I have to do a global skip if non-stop is requested.

jonahgraham:
Similar to this special test:
https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/80836a832c1d7cff86dd335c4c76b864f8fb3c99/src/integration-tests/multithread.spec.ts#L139-L142

PR #215, opened 2022-09-18
Fake fails to test PR flow

jonahgraham:
This PR is to test that the error reporting is reasonable when a PR has test errors.


jonahgraham:
That seems to report back well enough.

PR #218, opened 2022-09-26
Support set Hardware Breakpoint when debugging

thamho0902:
Root-cause: VS Code does not have option to set hardware breakpoint when debugging.
=> Solution: Add option for selecting use hardware breakpoint or software breakpoint by config in launch.json file


jonahgraham (changes requested):
This comment is design decision question, I would appreciate your thoughts on this before we merge it.

I'm not sure this is the correct solution to HW breakpoints. It may be suitable for now, but in particular I think `hardware-breakpoint-limit` should be configurable and set along with turning on hardware breakpoints.

IIRC the other way to get gdb to do this is to set `hardware-breakpoint-limit` and set all memory to be `ro` and let gdb manage it automatically.

src/mi/breakpoint.ts
```diff
@@ -114,18 +114,20 @@ export async function sendBreakInsert(
         threadId?: string;
         source: string;
         line: number;
+        hardwareBreakpoint: boolean;
     }
 ): Promise<MIBreakInsertResponse> {
     // Todo: lots of options
     const temp = request.temporary ? '-t ' : '';
     const ignore = request.ignoreCount ? `-i ${request.ignoreCount} ` : '';
+    const hwBreakpoint = request.hardwareBreakpoint ? '-h -f ' : '';
```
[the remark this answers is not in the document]
danghaibang-banvien: Hi @jonahgraham-san,
1. I will change for function breakpoint in the next patch set.
2. I see in e2studio using the argument "-f" for send command set breakpoint. But when I test with VS Code, It seems not to impact debugging.
3. Do you mean we need to write a regression test for this ticket?
Thank you.

jonahgraham:
> This comment is design decision question, I would appreciate your thoughts on this before we merge it.
> 
> I'm not sure this is the correct solution to HW breakpoints. It may be suitable for now, but in particular I think `hardware-breakpoint-limit` should be configurable and set along with turning on hardware breakpoints.
> 
> IIRC the other way to get gdb to do this is to set `hardware-breakpoint-limit` and set all memory to be `ro` and let gdb manage it automatically.

Please share your comments on the above before I consider merging.

src/mi/breakpoint.ts
```diff
@@ -114,18 +114,20 @@ export async function sendBreakInsert(
         threadId?: string;
         source: string;
         line: number;
+        hardwareBreakpoint: boolean;
```
jonahgraham, line 117: Sorry I missed this before, but please reuse `hardware` which is already present rather than creating a new option.

src/mi/breakpoint.ts
```diff
@@ -154,10 +156,14 @@ export function sendBreakList(gdb: GDBBackend): Promise<MIBreakListResponse> {
 
 export async function sendBreakFunctionInsert(
     gdb: GDBBackend,
-    fn: string
+    fn: string,
+    request: {
+        hardwareBreakpoint: boolean;
```
jonahgraham, line 161: :-( - Unfortunately I didn't realize this before, but the parameters for `sendBreakFunctionInsert` are quite different than `sendBreakInsert`, but they really should be basically the same.

Any suggestion on how to resolve this? You can include that in this change, just note it in the commit comment so when we make a changelog we can see that. 

At the moment this change already breaks the API anyway (new required parameter), so lets see if we can do it properly.

PS I may have to sleep on this and see if we should be making this change backwards compatible or not.

jonahgraham (changes requested):
In addition to these comments, there are [unanswered questions from the last patchset](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/218#pullrequestreview-1120741070).

> 1. I will change for function breakpoint in the next patch set.

Thank you.

> 2. I see in e2studio using the argument "-f" for send command set breakpoint. But when I test with VS Code, It seems not to impact debugging.

Do you know what `-f` does? It may be important in some circumstances, but it is likely just as important for software as hardware breakpoints. If you need `-f`, the `pending` in requests should be used.

From https://sourceware.org/gdb/current/onlinedocs/gdb/GDB_002fMI-Breakpoint-Commands.html:
`-f` If _locspec_ cannot be resolved (for example if it refers to unknown files or functions), create a pending breakpoint. Without this flag, GDB will report an error, and won’t create a breakpoint, if _locspec_ cannot be parsed.



> 3. Do you mean we need to write a regression test for this ticket?

Yes. Always, for all tickets. If for some reason you can't, you should document that fact in the PR somewhere. Non-exhaustive list of possible reasons to not include a new test:
- Code being modified is already well exercised by existing tests
- Code being modified requires custom setup that is beyond the ability to test in the current suite (e.g. performance changes)

>  Additional information, we have tested with CVe Linux, and hardware breakpoint can set and hit normally.

This is useful information.

src/mi/breakpoint.ts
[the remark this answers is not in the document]
jonahgraham: I have moved this thread to the main feedback of the change.

src/mi/breakpoint.ts
danghaibang1999: Yes, I missed this argument, it had option hardware breakpoint before. Thank you.

danghaibang1999:
> Do you know what -f does? It may be important in some circumstances, but it is likely just as important for software as hardware breakpoints. If you need -f, the pending in requests should be used.
I see all breakpoint types in e2 studio always have option '-f'. If the decision is using '-f' option, I think it needs to fix in command send to GDB server. 
(example: const command = `-break-insert ${temp}${hwBreakpoint}${ignore} **-f** ${location}`;)

> Yes. Always, for all tickets. If for some reason you can't, you should document that fact in the PR somewhere. Non-exhaustive list of possible reasons to not include a new test:
I will write the integration test if possible. Or not, I will update document in Jira/Confluence ticket.
Thank you.

danghaibang1999:
> This comment is design decision question, I would appreciate your thoughts on this before we merge it.

I'm not sure this is the correct solution to HW breakpoints. It may be suitable for now, but in particular I think hardware-breakpoint-limit should be configurable and set along with turning on hardware breakpoints.

IIRC the other way to get gdb to do this is to set hardware-breakpoint-limit and set all memory to be ro and let gdb manage it automatically.

I am trying to use command '-gdb-set breakpoint auto-hw on' in launch process but it seems not to affect the type of breakpoint.
It is also for option 'hardware-breakpoint-limit', I am not sure if we have the special config for it or not but the command sent to Renesas GDB server has an error (cannot set option limit or unlimited).
So, I think using the current solution can be helpful, The extension has a limit in UI (has only the option "Add breakpoint" but cannot choose the type of breakpoint). I think the exact way is to improve the breakpoint widget to have an option "Add hardware breakpoint", but it seems not available for now.
I think my solution is fixing temporarily, if users do not care type of breakpoint, it is not configured in the launch file and software breakpoint is used for default.
Thank you.

src/mi/breakpoint.ts
danghaibang1999: I see `sendBreakFunctionInsert ` also supports `temporary`, `pending `, and `hardware ` commands. I think it has the argument same as the `sendBreakInsert` function.

> Any suggestion on how to resolve this? You can include that in this change, just note it in the commit comment so when we make a changelog we can see that.

I propose a solution that adds condition `?`  in parameter `request`. It seems this argument can be used or not. I can work compatible with the previous code. What do you think about it?

jonahgraham:
> I will write the integration test if possible. Or not, I will update document in Jira/Confluence ticket.

Great. Please also document the reasoning here. This is an open source project and we should try not to put cdt-gdb-adapater specific information in systems that are not visible to everyone. It is complicated because I have access to the JIRA, but imagine the reviewer of these PRs didn't.

src/integration-tests/utils.ts
```diff
@@ -239,6 +239,8 @@ export const gdbPath: string | undefined = getGdbPathCli();
 export const gdbServerPath: string = getGdbServerPathCli();
 export const debugServerPort: number | undefined = getDebugServerPortCli();
 export const defaultAdapter: string = getDefaultAdapterCli();
+export const hardwareBreapoint: boolean =
```
jonahgraham, line 242: This doesn't seem to be connected to anything. I think you want to add something to `fillDefaults` a few lines above this.

Additionally, this is an OK way to test the new functionality. Also acceptable is a single test added to `breakpoints.spec.ts` and `functionBreakpoints.spec.ts` that sets `hardwareBreakpoint`.

package.json
```diff
@@ -16,9 +16,10 @@
     "clean": "git clean -dfx",
     "docker:build": "docker run --rm -it -v $(git rev-parse --show-toplevel):/work -w /work/$(git rev-parse --show-prefix) --cap-add=SYS_PTRACE --security-opt seccomp=unconfined quay.io/eclipse-cdt/cdt-infra-eclipse-full:latest yarn",
     "docker:test": "docker run --rm -it -v $(git rev-parse --show-toplevel):/work -w /work/$(git rev-parse --show-prefix) --cap-add=SYS_PTRACE --security-opt seccomp=unconfined quay.io/eclipse-cdt/cdt-infra-eclipse-full:latest yarn test",
-    "test": "yarn test:integration && yarn test:integration-remote-target && yarn test:integration-gdb-async-off && yarn test:integration-gdb-async-off-remote-target && yarn test:integration-gdb-non-stop && yarn test:integration-gdb-non-stop-remote-target",
+    "test": "yarn test:integration && yarn test:integration-remote-target && yarn test:integration-gdb-async-off && yarn test:integration-gdb-async-off-remote-target && yarn test:integration-gdb-non-stop && yarn test:integration-gdb-non-stop-remote-target && yarn test:integration-hw-breakpoint-on",
```
jonahgraham, line 19: Can you make the same two changes to this file to the `test-ci` version please.

PS - if you have ideas on how to simplify all the `test` and `test-ci` scripts in `package.json` that would be great (but do it in a different PR though)

src/integration-tests/utils.ts
```diff
@@ -239,6 +239,8 @@ export const gdbPath: string | undefined = getGdbPathCli();
 export const gdbServerPath: string = getGdbServerPathCli();
 export const debugServerPort: number | undefined = getDebugServerPortCli();
 export const defaultAdapter: string = getDefaultAdapterCli();
+export const hardwareBreapoint: boolean = 
+    process.argv.indexOf('--test-hw-breakpoint-on') === -1;
```
jonahgraham, line 243: I think your conditional is backwards here. `indexOf` returns `-1` when the item is _not_ found.

jonahgraham (changes requested):
> I think my solution is fixing temporarily, if users do not care type of breakpoint, it is not configured in the launch file and software breakpoint is used for default. Thank you.

Sounds good, and I agree and will merge it once review comments are resolved. Thank you for including your reasoning. This is important both for the review, but also for someone who comes along in the future and wants to understand the _why_ we did something this way.

src/integration-tests/utils.ts
danghaibang1999: I have fixed it. Thank you.

package.json
danghaibang1999: Done. Thank you.

src/integration-tests/utils.ts
danghaibang1999: I added new TCs for `breakpoints.spec.ts` and `functionBreakpoints.spec.ts`. It runs ok in case of software breakpoint, but in case of hardware breakpoint, gdb for running test is not supported hardware breakpoint, and all TCs are failed. I cannot check the output of it in this case. Do you have any idea about it? Thank you.

jonahgraham: Here is the log extract:


```
[08:25:52.736 UTC] GDB command: 5 -break-insert -h --source "/home/runner/work/cdt-gdb-adapter/cdt-gdb-adapter/src/integration-tests/test-programs/count.c" --line 4
[08:25:52.737 UTC] GDB result: 5 error,msg="No hardware breakpoint support in the target."
[08:25:52.737 UTC] To client: {"seq":0,"type":"response","request_seq":3,"command":"setBreakpoints","success":true,"body":{"breakpoints":[{"verified":false,"message":"No hardware breakpoint support in the target."}]}}
```

Hardware breakpoints aren't generally supported by host dev, but it should work on remote, at least on linux. You may need to do skip for unsupported platforms, e.g. see:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/9d7b824523c2ec8abf5635c900ed2cd100812457/src/integration-tests/utils.ts#L254-L257

jonahgraham (changes requested):

src/integration-tests/utils.ts
```diff
@@ -223,6 +223,7 @@ export function fillDefaults(
     args.openGdbConsole = openGdbConsole;
     args.gdbAsync = gdbAsync;
     args.gdbNonStop = gdbNonStop;
+    args.hardwareBreakpoint = hardwareBreapoint;
```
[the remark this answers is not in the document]
danghaibang1999: Done. Thank you.

src/integration-tests/functionBreakpoints.spec.ts
```diff
@@ -39,6 +40,31 @@ describe('function breakpoints', async function () {
...
+        let substring: string;
+        if (hardwareBreapoint)
+        {
+            substring = 'type="hw breakpoint"';
+        }
+        else
+        {
+            substring = 'type="breakpoint"';
+        }
+        expect(outputs.body.output.includes(substring)).eq(true);
```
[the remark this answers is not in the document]
danghaibang1999: Done. Thank you.

thamho0902:
Sorry, I have a mistake. Please help me check again.

danghaibang1999:
Hi @jonahgraham-san,
It seems running test in windows have failed because option '--timeout 5000' in file package.json

> 1) "before all" hook in "{root}":
     Error: Command failed: make
c:/programdata/chocolatey/lib/mingw/tools/install/mingw64/bin/../lib/gcc/x86_64-w64-mingw32/11.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: cannot open output file empty.exe: Permission denied
collect2.exe: error: ld returned 1 exit status
mingw32-make: *** [makefile:30: empty] Error 1
      at checkExecSyncError (child_process.js:790:11)
      at Object.execSync (child_process.js:863:15)
      at Context.<anonymous> (src\integration-tests\utils.ts:251:12)
      at processImmediate (internal/timers.js:464:21)

I have no solution for this. Could you please help me give me your idea? Thank you.

package.json
```diff
@@ -28,9 +28,9 @@
...
     "test:integration-gdb-non-stop-remote-target": "cross-env JUNIT_REPORT_PATH=test-reports/integration-gdb-non-stop-remote-target.xml JUNIT_REPORT_STACK=1 JUNIT_REPORT_PACKAGES=1 mocha --exit --reporter mocha-jenkins-reporter -r ts-node/register src/integration-tests/*.spec.ts --test-gdb-non-stop --test-remote",
     "test:pty": "cross-env JUNIT_REPORT_PATH=test-reports/native.xml JUNIT_REPORT_STACK=1 JUNIT_REPORT_PACKAGES=1 mocha --exit --reporter mocha-jenkins-reporter dist/native/*.spec.js",
-    "test-ci": "run-s --continue-on-error test-ci:integration test-ci:integration-remote-target test-ci:integration-gdb-async-off test-ci:integration-gdb-async-off-remote-target test-ci:integration-gdb-non-stop test-ci:integration-gdb-non-stop-remote-target test-ci:integration-hw-breakpoint-on-remote-target",
+    "test-ci": "run-s --continue-on-error test-ci:integration test-ci:integration-remote-target test-ci:integration-gdb-async-off test-ci:integration-gdb-async-off-remote-target test-ci:integration-gdb-non-stop test-ci:integration-gdb-non-stop-remote-target",
```
jonahgraham: Why did you remove the hw breakpoints tests in this commit?

jonahgraham: I added the following commit:

[Restore running hw breakpoint tests as part of test and test-ci](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/218/commits/ea655665ec2fcb3429274e18636fd62b3e40b122) 
[ea65566](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/218/commits/ea655665ec2fcb3429274e18636fd62b3e40b122)

Note that I think the reason that you removed them is because the test-ci:integration-hw-breakpoint-on-remote-target was missing the --skip-make command line option. On the CI machine we can't run make reliably from within the mocha tests, so that is why I decided to split the make step out in commit

src/integration-tests/functionBreakpoints.spec.ts
```diff
@@ -39,6 +40,35 @@ describe('function breakpoints', async function () {
...
+        });
+        expect(bpResp.body.breakpoints.length).eq(1);
+        expect(bpResp.body.breakpoints[0].verified).eq(true);
+        expect(bpResp.body.breakpoints[0].message).eq(undefined);
+        await dc.configurationDoneRequest();
+        let isCorrect;
+        let outputs;
+        while (!isCorrect) {
+            // Cover the case of getting event in Linux environment.
+            // If cannot get correct event, program timeout and test case failed.
```
jonahgraham, line 59: I added the commit:
[Use the global timeout](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/218/commits/7576a29202407e54b6d9838c1abafbab83755a48) 
[7576a29](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/218/commits/7576a29202407e54b6d9838c1abafbab83755a48)
See issue https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/222 and the fix there for more details as to why.

jonahgraham (approved):
Rather than bounce it back to you one more time, I have fixed the remaining items (including formatting, which at some point I will add to the CI so you can be immediately notified if formatting/linting is incorrect)

Please do take the time to review the commits I pushed to your branch so you can know what and why I changed some items.

package.json
danghaibang1411: Hi @jonahgraham-san,
Sorry this is my mistake. When I change the code for fixing build in Github, I forgot this item. Thank for your support.

src/integration-tests/functionBreakpoints.spec.ts
danghaibang1411: Yes, I have seen your commit and fix for the previous build fail in this ticket. Thank you.

PR #221, opened 2022-09-27
Support debug multiple same ARM core

QuocDoBV:
Currently, we need to support multi-process debugging as each core  ARM (CA55 + CR52) for R-CAR S4 device.  Since it is not possible to override the export declare interface GDBBackend method in the RenesasGDBBackend.ts class or the GDBTargetDebugSession.ts class. We need to update the code at the GDBBackend.ts class of the cdt-gdb-adapter.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -41,6 +41,7 @@ export declare interface GDBBackend {
         event: 'execAsync' | 'notifyAsync' | 'statusAsync',
         listener: (asyncClass: string, data: any) => void
     ): this;
+    on(event: 'attachedRequest', listener: () => void): this;
 
     emit(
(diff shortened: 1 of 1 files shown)


src/GDBBackend.ts
```diff
@@ -41,6 +41,10 @@ export declare interface GDBBackend {
         event: 'execAsync' | 'notifyAsync' | 'statusAsync',
         listener: (asyncClass: string, data: any) => void
     ): this;
+    on(
```
jonahgraham, line 44: Extenders can't need to modify cdt-gdb-adapter to add events for their own use. 

As for Renesas' specific use-case I have commented on the internal system.

jonahgraham (changes requested):

PR #224, opened 2022-10-05
Fix Watch view for multiple session debugging.

bangdang1411:
Description: In multiple session debugging, all sessions will send command to get result value of  variables added in Watch view. But only 1 session can send command successfully, another session  will catch the error and send ErrorResponse to client. The ErrorRespone makes another session  fail and cannot update the remaining in Watch view.

Solution: Replace ErrorResponse by using response notify 'Error: could not evaluate expression'. It will show the error notify on another sessions that do not have the variable in source file.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1225,11 +1225,10 @@ export class GDBDebugSession extends LoggingDebugSession {
 
             this.sendResponse(response);
         } catch (err) {
-            this.sendErrorResponse(
-                response,
-                1,
-                err instanceof Error ? err.message : String(err)
-            );
+            // For multiple debug session, another sessions can catch error in case of not have variable in source.
+            // Cannot send ErrorResponse because it make another session are failed.
+            // Send the response notify 'Error: could not evaluate expression' instead of ErrorResponse.
+            this.sendResponse(response);
         }
     }
 


bangdang1411:
Hi @jonahgraham-san,
I pull a commit for fixing the bug relate to using watch view for multiple debugging (G4MH & ARM).
This commit relates to https://github.com/eclipse-cdt-cloud/cdt-amalgamator/pull/3
It also fixes in the same ticket. 
Could you please help me take a look at my commit and give me your idea? Thank you.

src/GDBDebugSession.ts
```diff
@@ -1225,11 +1225,10 @@ export class GDBDebugSession extends LoggingDebugSession {
 
             this.sendResponse(response);
         } catch (err) {
-            this.sendErrorResponse(
-                response,
-                1,
-                err instanceof Error ? err.message : String(err)
-            );
+            // For multiple debug session, another sessions can catch error in case of not have variable in source.
```
jonahgraham, line 1228: No, that is not correct, if there is an error the adapter must report it as an error. The caller can decide to ignore the error and call other adapters if it wants to in a multi-session.

Note that the tests now fail with this change, in particular look at tests `should reject evaluation of expression without a frame` and `should reject evaluation of invalid expression`

If there is additional information needed in the error, please feel free to provide a PR that does that.

jonahgraham (changes requested):
The CI results [are here](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/3188906690/jobs/5207732096). Please check them out.

src/GDBDebugSession.ts
bangdang1411: I understood. I will find a different solution for this. Thanks for your comment.

PR #225, opened 2022-10-31
Disassembly view CVe make ARM thread is terminated

bangdang1411:
+ Description: When using multiple debugging (ARM+CVe) for V4H devices, open Disassemble 
                    view in CVe thread, ARM thread will be terminated automatically.
+ Root cause: In e2studio, ARM and CVe debugging via Renesas GDB server (Passthrough).
                   But in VS code, it connects directly to Linux Applications. ARM thread terminated 
                   in case it's running, If ARM thread is suspended before opening Disassemble view in 
                   CVe thread, the bug does not occur.
+ Solution:    Currently, we have no solution to use Disassemble view for CVe, but it needs
                  to disable Disassemble view of CVe thread to fix degradation in the ARM thread.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -178,6 +178,7 @@ export class GDBDebugSession extends LoggingDebugSession {
     // therefore be resumed after breakpoints are inserted.
     protected waitPausedNeeded = false;
     protected isInitialized = false;
+    protected isNotSupportDisassemble = false;
 
     constructor() {
         super();
@@ -1292,6 +1293,11 @@ export class GDBDebugSession extends LoggingDebugSession {
             let lastStartOffset = -1;
             const instructions: DebugProtocol.DisassembledInstruction[] = [];
             let oneIterationOnly = false;
+            if (this.isNotSupportDisassemble) {
+                response.body = { instructions };
+                this.sendResponse(response);
+                return;
+            }
             outer_loop: while (
                 instructions.length < args.instructionCount &&
                 !oneIterationOnly


bangdang1411:
Hi @jonahgraham-san,
Related to the bug open disassembly view in CVe thread makes ARM thread terminated automatically,
I have no solution to fix it now because VS Code does not use GDB server (Passthrough) for debugging.
I propose the solution to disable disassembly view of CVe thread to fix the degraded bug.
Maybe this bug will be fixed in the future, but this work is necessary for normal work of multiple debugging (ARM + Cve).
If you have any solution or idea to fix it, please teach me about it.
Thank you.

src/GDBDebugSession.ts
```diff
@@ -1292,6 +1293,11 @@ export class GDBDebugSession extends LoggingDebugSession {
             let lastStartOffset = -1;
             const instructions: DebugProtocol.DisassembledInstruction[] = [];
             let oneIterationOnly = false;
+            if (this.isNotSupportDisassemble) {
```
bangdang1411, line 1296: This variable will be set to "true" in RenesasGDBTargetDebugSession if this is a thread of CVe debugging.

jonahgraham: If you want to have custom disassembly behaviour, can you override `disassembleRequest` instead? Perhaps something like this (I wrote this in the comment, so there may be compiler errors):

```typescript
protected async disassembleRequest(
        response: DebugProtocol.DisassembleResponse,
        args: CDTDisassembleArguments
    ) {
         if (this.isNotSupportDisassemble) {
            response.body = { [] };
            this.sendResponse(response);
         } else {
           super.disassembleRequest(response, args);
         }
}
```

However, it seems that if the adapter can't provide disassembly should it reply with an error? I am genuinely not sure and you can provide your thoughts on that. But either way, I don't think you should be modifying cdt-gdb-adapter at all.

---

The above may become irrelevant if you can simply stop the client making `disassemble` requests by advertising `supportsDisassembleRequest` properly. This may be impossible in the context of amalgamator, but I leave this comment for completeness:

Does the `isNotSupportDisassemble` flag change at runtime? If not, then instead of the above, reply to client [`initialize`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Initialize) request with the `supportsDisassembleRequest` as false. If it does change at runtime, you may also send the [`capabilities` event](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Capabilities)

jonahgraham (changes requested):

src/GDBDebugSession.ts
bangdang1411: > If you want to have custom disassembly behaviour, can you override disassembleRequest instead? Perhaps something like this (I wrote this in the comment, so there may be compiler errors):
> 
> protected async disassembleRequest(
>         response: DebugProtocol.DisassembleResponse,
>         args: CDTDisassembleArguments
>     ) {
>          if (this.isNotSupportDisassemble) {
>             response.body = { [] };
>             this.sendResponse(response);
>          } else {
>            super.disassembleRequest(response, args);
>          }
> }

Do you mean this change should be in RenesasGDBTargetDebugSession?
> Does the isNotSupportDisassemble flag change at runtime?

I think yes. this flag is changed after initializing has been done.

jonahgraham: > Do you mean this change should be in RenesasGDBTargetDebugSession?

Yes.

bangdang1411: I will test again and commit the code into Gerrit if it can work normally.

jonahgraham: OK. I will close this PR for now then.

PR #226, opened 2022-11-29
Support Memory Browser for multiple child

trongle0504:
Support Memory Browser for multiple children

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -33,6 +33,7 @@ import {
 import { StoppedEvent } from './stoppedEvent';
 import { VarObjType } from './varManager';
 import { breakpointFunctionLocation, breakpointLocation } from './mi';
+import { ChildDapContents } from 'cdt-amalgamator';
 
 export interface RequestArguments extends DebugProtocol.LaunchRequestArguments {
     gdb?: string;
@@ -86,7 +87,6 @@ export type VariableReference =
 export interface MemoryRequestArguments {
     address: string;
     length: number;
-    child: string;
     offset?: number;
 }
 
@@ -97,7 +97,6 @@ export interface MemoryContents {
     /* Hex-encoded string of bytes.  */
     data: string;
     address: string;
-    child?: string[];
 }
 
 export interface MemoryResponse extends Response {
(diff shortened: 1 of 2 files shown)


src/GDBDebugSession.ts
```diff
@@ -198,7 +200,12 @@ export class GDBDebugSession extends LoggingDebugSession {
         args: any
     ): void {
         if (command === 'cdt-gdb-adapter/Memory') {
-            this.memoryRequest(response as MemoryResponse, args);
+            if (args.address === '') {
```
jonahgraham, line 203: Any changes in cdt-gdb-adapter require tests. Please provide a test along with explanation about what this code is changing.

jonahgraham (changes requested):
I haven't looked at the next layers in the stack yet, but these changes don't look like they belong here. Especially a required new field in an arguments interface requires strong justification/explanation and tests.

src/GDBDebugSession.ts
```diff
@@ -96,6 +97,7 @@ export interface MemoryContents {
     /* Hex-encoded string of bytes.  */
     data: string;
     address: string;
+    child?: string[];
```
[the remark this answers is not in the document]
trongle0504: It doesn't remain after changing the code.

src/GDBDebugSession.ts
trongle0504: It doesn't remain after changing the code

src/GDBDebugSession.ts
```diff
@@ -201,6 +202,10 @@ export class GDBDebugSession extends LoggingDebugSession {
             this.memoryRequest(response as MemoryResponse, args);
             // This custom request exists to allow tests in this repository to run arbitrary commands
             // Use at your own risk!
+        } else if ('cdt-amalgamator/getChildDapNames') {
```
jonahgraham, line 205: The adapter doesn't/shouldn't need to know about custom messages of the amalgamator. The client code can issue this command, and handle the case where it fails.

jonahgraham (changes requested):
I think this change should simply be closed.

trongle0504:
No need changes with new solution

PR #228, opened 2023-01-18
Allow pre-filling launch/attach on command line to adapter

### LINKED ISSUE #227: Support configuration when in `Server Mode`
We have a use case running `cdt-gdb-adapter` in a remote containerised environment and exposing a socket to control the debug interface by running a command of the form:

```bash
node debugTargetAdapter.js --server=50000
```

This works well, but requires the connecting client to specify the program to be debugged (in launch mode) or the local port to attach to (in attach mode).

It would be ideal if configuration of the environment was isolated within the container via environment variables or by passing a config file location as a further switch when running `debugTargetAdapter.js` e.g.:

```bash
node debugTargetAdapter.js --server=50000 --config=config.json
```

Contents of config.json:

```json
{
    "request": "attach",
    "port": 50000
}
```

or 

```json
{
    "request": "launch",
    "program": "/path/to/my/app"
}
```

Is this a valid addition and would there be a better/easier way to implement it?

jonahgraham:
#### `--config=INITIALCONFIG`

Start the adapter using the given configuration as a starting point for the args in `launch` or `attach` request.

For example, the default GDB can be set like this:

```sh
    node debugTargetAdapter.js --config='{"gdb":"arm-none-eabi-gdb"}'
```

The config can be passed on the command line as JSON, or a response file can be used by starting the argument with `@`.
The rest of the argument will be interpreted as a file name to read.
For example, to start the adapter defaulting to a process ID to attach to, create a file containing the JSON and reference it like this:

```sh
    cat >config.json <<END
    {
      "processId": 1234
    }
    END
    node debugAdapter.js --config=@config.json

```

#### `--config-frozen=FROZENCONFIG`

Similar to `--config`, the `--config-frozen` sets the provided configuration fields in the args to the `launch` or `attach` request to the given values, not allowing the user to override them.
Specifying which type of request is allowed (`launch` or `attach`) can be specified with the `request` field.

For example, the adapter can be configured for program to be frozen to a specific value.
This may be useful for starting adapters in a container and exposing the server port.

```sh
    node debugAdapter.js --server=23221 --config-frozen='{"program":"/path/to/my.elf"}'
```

Fixes #227

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -9,7 +9,9 @@
  *********************************************************************/
 import * as os from 'os';
 import * as path from 'path';
+import * as fs from 'fs';
 import {
+    DebugSession,
     Handles,
     InitializedEvent,
     Logger,
@@ -152,6 +154,18 @@ export function base64ToHex(base64: string): string {
 }
 
 export class GDBDebugSession extends LoggingDebugSession {
+    /**
+     * Initial (aka default) configuration for launch/attach request
+     * typically supplied with the --config command line argument.
+     */
+    protected static defaultRequestArguments?: any;
+
+    /**
+     * Frozen configuration for launch/attach request
+     * typically supplied with the --config-frozen command line argument.
+     */
+    protected static frozenRequestArguments?: { request?: string };
+
     protected gdb: GDBBackend = this.createBackend();
     protected isAttach = false;
     // isRunning === true means there are no threads stopped.
@@ -182,6 +196,66 @@ export class GDBDebugSession extends LoggingDebugSession {
     constructor() {
         super();
         this.logger = logger;
+    }
+
+    /**
+     * Main entry point
+     */
+    public static run(debugSession: typeof GDBDebugSession) {
+        GDBDebugSession.processArgv(process.argv.slice(2));
+        DebugSession.run(debugSession);
+    }
+
+    /**
+     * Parse an optional config file which is a JSON string of launch/attach request arguments.
+     * The config can be a response file by starting with an @.
+     */
+    public static processArgv(args: string[]) {
+        args.forEach(function (val, _index, _array) {
+            const configMatch = /^--config(-frozen)?=(.*)$/.exec(val);
+            if (configMatch) {
+                let configJson;
+                const configStr = configMatch[2];
+                if (configStr.startsWith('@')) {
+                    const configFile = configStr.slice(1);
+                    configJson = JSON.parse(
+                        fs.readFileSync(configFile).toString('utf8')
+                    );
+                } else {
+                    configJson = JSON.parse(configStr);
+                }
+                if (configMatch[1]) {
+                    GDBDebugSession.frozenRequestArguments = configJson;
+                } else {
+                    GDBDebugSession.defaultRequestArguments = configJson;
+                }
+            }
+        });
+    }
+
+    /**
+     * Apply the initial and frozen launch/attach request arguments.
+     * @param request the default request type to return if request type is not frozen
+     * @param args the arguments from the user to apply initial and frozen arguments to.
+     * @returns resolved request type and the resolved arguments
+     */
+    protected applyRequestArguments(
+        request: 'launch' | 'attach',
+        args: LaunchRequestArguments | AttachRequestArguments
+    ): ['launch' | 'attach', LaunchRequestArguments | AttachRequestArguments] {
+        const frozenRequest = GDBDebugSession.frozenRequestArguments?.request;
+        if (frozenRequest === 'launch' || frozenRequest === 'attach') {
+            request = frozenRequest;
+        }
+
+        return [
+            request,
+            {
+                ...GDBDebugSession.defaultRequestArguments,
+                ...args,
+                ...GDBDebugSession.frozenRequestArguments,
+            },
+        ];
     }
 
     protected createBackend(): GDBBackend {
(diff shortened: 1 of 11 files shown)


thegecko:
Thanks @jonahgraham, this is exactly what I was thinking.

A couple of observations considering the "server" owner should potentially own the setup configuration.

- Should the server somehow dictate whether the session is a `launch` or `attach`? 
- When configuration is specified in the server, could it be a security concern to allow the client to override the data?

e.g. considering:
```typescript
            args = {
                ...GDBDebugSession.commandLineLaunchAttachArguments,
                ...args,
            };
```

A situation could exist where a remote server specifies the binary to be debugged, but the remote client can override this and point at arbitrary binaries.

For now I don't belive these points should block merging, though!

jonahgraham:
> For now I don't belive these points should block merging, though!

OK. I'll test this draft works and work towards merging.

> Should the server somehow dictate whether the session is a launch or attach?
> When configuration is specified in the server, could it be a security concern to allow the client to override the data?

I'll consider how to address this as I test.

jonahgraham:
Updated description to include details on `--config-frozen` that I added to address @thegecko's concerns raised.

jonahgraham:
@thegecko this is ready for review (docs + tests done)

jonahgraham:
The change (hopefully just the tests) don't work on Windows. I'll have to spin up there to figure out what went wrong, but my guess is it has to do with how we are passing the extra args in the debug adapter launch here:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/2ceef5b03b0cfd8b70cb6fcaf386e19ff17670a9/src/integration-tests/utils.ts#L193-L203

That code relies on shell=true so that all the arguments can be passed as one parameter, but with all the quoting my guess is on Windows the shell handling causes the arguments to be split in the wrong place. I have some WIP on allowing this in cdt-gdb-adapter [here](https://github.com/jonahgraham/cdt-gdb-adapter/tree/args-to-commandline), but the better solution would be to fix it upstream, probably [like this](https://github.com/jonahgraham/vscode-debugadapter-node/commit/9661b841afc2d1f96d60301dc012d107712c5a60), but that isn't tested.

src/GDBDebugSession.ts
```diff
@@ -184,6 +198,68 @@ export class GDBDebugSession extends LoggingDebugSession {
...
+     * @param request the type of request to test against the frozen request type
+     * @param args the arguments from the user to apply initial and frozen arguments to.
+     * @returns new arguments
+     * @throws Error if request is not permitted by frozen config
+     */
+    protected applyRequestArguments(
+        request: 'launch' | 'attach',
+        args: LaunchRequestArguments | AttachRequestArguments
+    ): LaunchRequestArguments | AttachRequestArguments {
+        const frozenRequest = GDBDebugSession.frozenRequestArguments?.request;
```
thegecko, line 247: Can this check be removed?

The reason is a little convoluted, but I'll try...

- The server owner sets up the debugger container and chooses `attach` or `launch` for GDB. In my test case I'm using `launch`.
- From a remote VSCode client connection, you are always `attach`ing to the remote running instance, so that's the way I'm specifying to connect to a remote system rather than launching something.

So the above scenario throws an error when in fact the launch type from the client should just be ignored. Therefore I think this check shouldn't be enforced.

thegecko (changes requested):
A small request to remove a check, but apart from that this works well for me. I'll share an example of it in use shortly

src/GDBDebugSession.ts
jonahgraham: I'll wait for your example as that may clarify, but not sure I understand. If VSCode does attach request, how does the adapter run the launch request code?

BTW here I am talking about whether the adapter launches or attaches to the target, as opposed to whether VSCode launches or attaches to the adapter. Is that where my confusion lies?

How are you attaching VSCode to the adapter, are you using `debugServer` or will you implement a custom launch code using `DebugAdapterDescriptor`/`DebugAdapterServer`?

Regardless, I am fine to remove the check.

thegecko: Example uploaded here:

https://github.com/thegecko/cdt-gdb-adapter-docker

jonahgraham: Thanks. I had a look at that and I understand almost what is happening.

What I don't understand is when the "attach" request in the launch.json turns into a "launch" request in cdt-gdb-adapter. IIUC VSCode will start the arm-debug adapter, and then send it an "attachRequest", so presumably arm-debug then connects to the given address and sends an initializeRequest + launchRequest?

thegecko: I thought you might notice this. It's currently a hack which will be removed from our adapter shortly.

By the way, happy to jump on a call and discuss all this, I want to get it right (and not just for our use case!)

jonahgraham: I can do right now if you are available. Send me an email, or I have a zoom account I can send you an invite.

jonahgraham:
> The change (hopefully just the tests) don't work on Windows. 

Windows issues have been fixed, all CI green now.

jonahgraham:
@thegecko As discussed I have refactored launch/attach request to have a true common code for both of them in 594348825e0a76185e0e0469e91bb39a3bf21a17

The final commit 55510156bd2231a9f30bdc408a2387505c5f586d in this series is the implementation of the feature request. The preceding commits are all the cleanup/refactoring that I did to make it easier to read and review 55510156bd2231a9f30bdc408a2387505c5f586d

thegecko (approved):
Thanks @jonahgraham. I tested passing an `attach` config with a frozen `launch` server config and all works as expected.

I'd be keen to see a new release when this gets merged :)

PR #230, opened 2023-01-24
Add binary entries

thegecko:
This PR adds binary entries for the debugger entry points.

This means that when the package is installed globally, you can execute the debugger using `cdtDebugTargetAdapter` rather than fully qualifying the javascript file (e.g. `/usr/local/lib/node_modules/cdt-gdb-adapter/dist/debugTargetAdapter.js`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ Build is pretty simple.
 yarn
 ```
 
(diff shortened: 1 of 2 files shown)


jonahgraham:
Thanks Rob.

PR #234, opened 2023-01-29
Ensure terminated event is sent to client

### LINKED ISSUE #233: The frontend ui don't stop when program over too quickly.
The example：
企业微信截图_16749722914597

The gdb exited, but frontend ui is running status.

Noneghost:
Consider that when no threads are running the inferior is not running. This is needed so we don't suppress sending events to client on async stopped events from gdb
    
Fixes #233

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1633,7 +1633,7 @@ export class GDBDebugSession extends LoggingDebugSession {
 
     protected handleGDBAsync(resultClass: string, resultData: any) {
         const updateIsRunning = () => {
-            this.isRunning = true;
+            this.isRunning = this.threads.length ? true : false;
             for (const thread of this.threads) {
                 if (!thread.running) {
                     this.isRunning = false;
--- a/src/integration-tests/terminated.spec.ts
+++ b/src/integration-tests/terminated.spec.ts
@@ -1,0 +1,67 @@
+/*********************************************************************
+ * Copyright (c) 2023 Kichwa Coders Canada Inc. and others
+ *
+ * This program and the accompanying materials are made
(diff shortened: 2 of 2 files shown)


jonahgraham:
Hi @Noneghost, thank you for the PR. Before we can consider the PR we need the ECA to be signed. See [here](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement) for more details.

Noneghost:
> Hi @Noneghost, thank you for the PR. Before we can consider the PR we need the ECA to be signed. See [here](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/CONTRIBUTING.md#eclipse-contributor-agreement) for more details.

Hi @jonahgraham , I haved signed eca 3.1.0.

jonahgraham (changes requested):
Thanks for the fix, it looks correct to me and I can manually that this fixes the problem. AFAICT it isn't limited to just short running programs. If I add a breakpoint in main somewhere and then continue I also don't get the terminated event from the adapter.

A new test is needed to make sure that terminated event arrives. Using launch.spec.ts as a starting point and then issuing a continue request and verifying a terminate event arrives would probably suffice.

If there is a timing thing going on that is much more annoying as it is difficult to test automatically.

GitMensch:
So, possibly take in, if we already know that there's an issue which is fixed by this? (I'm seeing the same issue - process is terminated but still "active" in the UI).

jonahgraham:
> So, possibly take in, [...]

Sorry, I don't know what you mean by this part?

GitMensch:
If you know that the known bug is fixed I suggest to pull that as-is, and adding "more missing things" later.

Issue #239, opened 2023-01-31
Not all GDB errors are shown in the debug console?


GitMensch:
I've wondered why a launch configuration did not work, and only after `"openGdbConsole": true,` was enabled I've seen it _there_: "symbol lookup error: undefined symbol" (there was a bad .so in LD_LIBRARY_PATH).

Any idea why this was not reported by the debug adapter to the ui?`

~~~
To client: {"seq":0,"type":"response","request_seq":2,"command":"launch","success":true}
From client: setFunctionBreakpoints({"breakpoints":[]})
GDB command: 5 -break-list
GDB result: 5 done,BreakpointTable={nr_rows="0",nr_cols="6",hdr=[{width="7",alignment="-1",col_name="number",colhdr="Num"},{width="14",alignment="-1",col_name="type",colhdr="Type"},{width="4",alignment="-1",col_name="disp",colhdr="Disp"},{width="3",alignment="-1",col_name="enabled",colhdr="Enb"},{width="10",alignment="-1",col_name="addr",colhdr="Address"},{width="40",alignment="2",col_name="what",colhdr="What"}],body=[]}
To client: {"seq":0,"type":"response","request_seq":3,"command":"setFunctionBreakpoints","success":true,"body":{"breakpoints":[]}}
From client: setExceptionBreakpoints({"filters":[]})
To client: {"seq":0,"type":"response","request_seq":4,"command":"setExceptionBreakpoints","success":true}
From client: configurationDone({})
GDB command: 6 -exec-run
GDB notify async: thread-group-started,id="i1",pid="23213"
GDB notify async: thread-created,id="1",group-id="i1"
GDB notify async: library-loaded,id="/lib64/ld-linux-x86-64.so.2",target-name="/lib64/ld-linux-x86-64.so.2",host-name="/lib64/ld-linux-x86-64.so.2",symbols-loaded="0",thread-group="i1",ranges=[{from="0x00007ffff7fd3090",to="0x00007ffff7ff22a0"}]
GDB result: 6 running
GDB exec async: running,thread-id="all"
To client: {"seq":0,"type":"response","request_seq":5,"command":"configurationDone","success":true}
From client: threads({})
To client: {"seq":0,"type":"response","request_seq":6,"command":"threads","success":true,"body":{"threads":[{"id":1,"name":"1","running":true}]}}
[...]

PR #241, opened 2023-01-31
Add a test executable with a broken shared library

jonahgraham:
This is based on the usecase described in #239 and will form the basis of a test to make sure we don't lose such error messages.

To show the error in GDB outside of the adapter, do this:

```sh
$ make BrokenSO
gcc -shared -o libsomea.so -fPIC -g3 BrokenSO_libsomea.c
gcc -shared -o libsomeb.so -fPIC -g3 BrokenSO_libsomeb.c
gcc -o BrokenSO BrokenSO_main.c -g3 -L. -lsomea
mv libsomeb.so libsomea.so

$ LD_LIBRARY_PATH=$PWD gdb --quiet BrokenSO
Reading symbols from BrokenSO...
(gdb) b main
Breakpoint 1 at 0x1151: file BrokenSO_main.c, line 4.
(gdb) r
Starting program: /scratch/debug/git/cdt-gdb-adapter/src/integration-tests/test-programs/BrokenSO
/scratch/debug/git/cdt-gdb-adapter/src/integration-tests/test-programs/BrokenSO: symbol lookup error: /scratch/debug/git/cdt-gdb-adapter/src/integration-tests/test-programs/BrokenSO: undefined symbol: a
[Inferior 1 (process 212237) exited with code 0177]
(gdb)
```

### DIFF
--- a/src/integration-tests/test-programs/BrokenSO_libsomea.c
+++ b/src/integration-tests/test-programs/BrokenSO_libsomea.c
@@ -1,0 +1,1 @@
+void a(void) {}
--- a/src/integration-tests/test-programs/BrokenSO_libsomeb.c
+++ b/src/integration-tests/test-programs/BrokenSO_libsomeb.c
@@ -1,0 +1,1 @@
+void b(void) {}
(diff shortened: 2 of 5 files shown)


jonahgraham:
I don't think this test code adds anything valuable to the mix as there is no additional info returned by gdb, this code is pretty much equivalent of:

```c
int main() {
    fprintf(stderr, "error message\n");
    return 0177;
}
```

The big difference is that the code doesn't get to main.

PR #254, opened 2023-02-08
Update 'result' listener

trongle0504:
Support 'result' listener at handleLine() function of MIParser class so that user can parse the returned data.

### DIFF
--- a/src/GDBBackend.ts
+++ b/src/GDBBackend.ts
@@ -38,7 +38,7 @@ export declare interface GDBBackend {
         listener: (output: string, category: string) => void
     ): this;
     on(
-        event: 'execAsync' | 'notifyAsync' | 'statusAsync',
+        event: 'execAsync' | 'notifyAsync' | 'statusAsync' | 'result',
         listener: (asyncClass: string, data: any) => void
     ): this;
 
@@ -48,7 +48,7 @@ export declare interface GDBBackend {
         category: string
     ): boolean;
     emit(
-        event: 'execAsync' | 'notifyAsync' | 'statusAsync',
+        event: 'execAsync' | 'notifyAsync' | 'statusAsync' | 'result',
         asyncClass: string,
         data: any
     ): boolean;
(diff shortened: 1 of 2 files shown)


jonahgraham (changes requested):
Thanks for the PR @trongle0504 . Can you please clarify the purpose? The result you are emitting is a synchronous thing, but you have added handling to the asynchronous code. AFAIK there won't be enough information in the result for any consumers to be able to process it. 

For example, if you have an MI sequence like this:

```
  123-add-inferior
  =thread-group-added,id="i2"
  ~"[New inferior 2]\n"
  123^done,inferior="i2"
```

The way to pick up the result is to await the result of [`sendCommand`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/b2e5222c837ffbc802ba0ea67866b14fe2507208/src/GDBBackend.ts#L204) and type it to the right type.

For example, look at:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/4eac6259e552ecefbfcb33b738b9d0ffe9fa5f78/src/mi/data.ts#L53-L62

This command sends `-data-read-memory-bytes` and types the return type to the expected structured data:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/4eac6259e552ecefbfcb33b738b9d0ffe9fa5f78/src/mi/data.ts#L14-L21

Then the consumer of that API doesn't have to do much special as the return value is a nicely typed object:

image

allowing the API consumer to do little work:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/9202c1c61e34fe4e99e9cf8b064dcff1745df5c9/src/GDBDebugSession.ts#L1621-L1631

If the above does not cover what you want, please can you please explain what it is you are trying to do?

trongle0504:
Thank you for your comment. I now have a clearer understanding of the problem. I will be closing this request.

PR #255, opened 2023-02-09
Support resume all sessions

trongle0504:
Adding a "Resume All" button to support resuming all sessions using the amalgamator.


src/GDBDebugSession.ts
```diff
@@ -999,6 +1003,23 @@ export class GDBDebugSession extends LoggingDebugSession {
         }
     }
 
+    protected async continueAllSessionsRequest(
+        response: DebugProtocol.ContinueResponse,
+        args: DebugProtocol.ContinueArguments
+    ): Promise<void> {
+        try {
+            await mi.sendExecContinue(this.gdb, args.threadId);
+            this.sendEvent(new ContinuedEvent(args.threadId));
```
jonahgraham, line 1012: Don't send events here, make the amalgamator send a normal continue request and then generate the extra events needed as discussed in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/80#discussion_r1102220588

jonahgraham (changes requested):
I believe all the changes to cdt-gdb-adapter are unneeded.

src/GDBDebugSession.ts
trongle0504: I have removed the event submission here. However, in this repository, we need to check the thread states at the `continueRequest()` function before sending `mi.sendExecContinue() ` to GDB. The purpose of this is to avoid the error `'Cannot execute this command while the selected thread is running.'`

src/GDBDebugSession.ts
```diff
@@ -988,7 +988,9 @@ export class GDBDebugSession extends LoggingDebugSession {
         args: DebugProtocol.ContinueArguments
     ): Promise<void> {
         try {
-            await mi.sendExecContinue(this.gdb, args.threadId);
+            if(this.isRunning === false) {
```
jonahgraham, line 991: Please don't suppress this error. If the client asks to resume, but the thread is running already the adapter should send back an error indicating that.

This is especially true since your check doesn't match the continue arguments. You are checking against the state of all threads, but may be resuming just a single thread.

jonahgraham (changes requested):

trongle0504:
This pull request is closed because the code was executed at cdt-amalgamator.

PR #257, opened 2023-02-27
Update status of multi threads in Call Stack View

bangdang1411:
For multiple cores debugging, the state of cores is not updated correctly with the state in the GDB server. Cannot use "Resume" or "Suspend" button in case of multiple cores debugging. That has not happened in cdt-amalgamator. This solution refers to cdt-amalgamator repository.

Root cause: Call Stack View does not update the state RUNNING if the process *Initialize* is not completed. cdt-gdb-adapter does not support sending "ContinuedEvent" when GDB returns async message "running"

Solution: Return Continues Response with the arg "allThreadsContinued" is 'false', which makes all threads not set to "RUNNING" state. Send ContinuedEvent when handling async message "running" from the GDB server. Remove the condition that does not send StoppedEvent if the process Initialize is not completed.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -1007,9 +1007,6 @@ export class GDBDebugSession extends LoggingDebugSession {
     ): Promise<void> {
         try {
             await mi.sendExecContinue(this.gdb, args.threadId);
-            response.body = {
-                allThreadsContinued: false,
-            }
             this.sendResponse(response);
         } catch (err) {
             this.sendErrorResponse(
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -989,7 +989,12 @@ export class GDBDebugSession extends LoggingDebugSession {
     ): Promise<void> {
         try {
             await mi.sendExecContinue(this.gdb, args.threadId);
-            const isAllThreadsContinued = args.threadId ? false : true;
+            let isAllThreadsContinued;
+            if (this.gdb.isNonStopMode()) {
+                isAllThreadsContinued = args.threadId ? false : true;
+            } else {
+                isAllThreadsContinued = true;
+            }
             response.body = {
                 allThreadsContinued: isAllThreadsContinued,
             }
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -23,7 +23,6 @@ import {
     Source,
     StackFrame,
     TerminatedEvent,
-    Event,
 } from '@vscode/debugadapter';
 import { DebugProtocol } from '@vscode/debugprotocol';
 import { GDBBackend } from './GDBBackend';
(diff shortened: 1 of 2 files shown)


bangdang1411:
Hi @jonahgraham-san,
This is the first solution to fix some issues when debugging with multiple cores devices.
I am not sure that we can use it.
Anyway, I think we can discuss more in this PR.
Please help me take a look at it and give me your comment. Thank you.

jonahgraham:
Can you provide a reproducible test that demonstrates this? Ideally as an integration test, but if not at least as a step-by-step test, or if even that is impossible, some MI/DAP logs that show the problem.

There is [multithread](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/f1481573fa3f07c4afc812e848d0eb1a7cd3437e/src/integration-tests/multithread.spec.ts) that you can use as a starting point for a test.

jonahgraham:
PS This change has caused a bunch of test failures - see https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/4283245156/jobs/7465169867

bangdang1411:
> Can you provide a reproducible test that demonstrates this? Ideally as an integration test, but if not at least as a step-by-step test, or if even that is impossible, some MI/DAP logs that show the problem.
> 
> There is [multithread](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/f1481573fa3f07c4afc812e848d0eb1a7cd3437e/src/integration-tests/multithread.spec.ts) that you can use as a starting point for a test.

I have mentioned you in the Jira ticket. Please help me see the document that describes the solution and the log file of it.
Thank you.

jonahgraham:
> I have mentioned you in the Jira ticket. 

This is an open source project and should operate in the open. Please provide details accessible to all.

bangdang1411:
It includes some information of renesas-gdb-adapter. Do you think it need to accessible by everyone?
I think I need to prepare some information before push it in Github. Thank you.

bangdang1411:
> @bangdang1411, I think my previous comment summed it up best:
> 
> > Can you provide a reproducible test that demonstrates this? Ideally as an integration test, but if not at least as a step-by-step test, or if even that is impossible, some MI/DAP logs that show the problem.
> 
> It is fine to mention the renesas-gdb-adapter, but it would be best to have a test case that doesn't rely on it.

Sorry, I am investigating the reason that make the test in Linux is failed. Also, I am writing the integration test for it but not completed for now.
You can see my idea first until I can commit the test.
Thank you.

bangdang1411:
The current commit only fixes for the bug that selects Resume 1 thread but all threads are Resuming.

src/GDBDebugSession.ts
```diff
@@ -989,6 +989,10 @@ export class GDBDebugSession extends LoggingDebugSession {
     ): Promise<void> {
         try {
             await mi.sendExecContinue(this.gdb, args.threadId);
+            const isAllThreadsContinued = args.threadId ? false : true;
```
jonahgraham, line 992: I am pretty sure this is incorrect in the all-stop case. In allstop even if you continue a single thread all threads are continued.

jonahgraham (changes requested):
> The current commit only fixes for the bug that selects Resume 1 thread but all threads are Resuming.

This is a pretty self-contained change now. Please provide a test so I can effectively review it and get it merged.

bangdang1411:
I attached the log file when using this solution. Thank you.
[RH850_LogFile.txt](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/files/10870354/RH850_LogFile.txt)

jonahgraham (approved):
This looks good. Once the CI completes successfully it can be merged.

Issue #253, opened 2023-02-08
Ending an OpenOCD process at the end of debugging


jonahgraham:
When launching target server (gdbserver, openocd, other) the current implementation of cdt-gdb-adapter assumes that the server will be launched in a way that the server terminates once gdb disconnects. That is what happens with gdbserver and the `--once` argument.

For targets like openocd, having a way to exit cleanly is needed.

### Discussed in https://github.com/eclipse-cdt-cloud/cdt-cloud/discussions/27

<div type='discussions-op-text'>

<sup>Originally posted by **ConorDavenport** February  7, 2023</sup>
Hi,

I'm using Theia Blueprint with the cdt-gdb-vscode extension to debug an application on the Polarfire SoC. The debug configuration can be seen below. An OpenOCD instance is started and gdb connects to it. I am able to step through code and debug as expected, however when ending the debug session the instance of OpenOCD stays active and needs to be killed manually. Is there any way to configure the debug session to kill the OpenOCD process when the deubg session ends? Thanks!

```    
        "configurations": [{
        "name":"debug-smp",
        "type":"gdbtarget",
        "request":"launch",
        "program": "path/to/executable",
        "gdb": "path/to/gdb",
        "logFile": "path/to/logfile",
        "openGdbConsole": true,
        "initCommands": [
            "set mem inaccessible-by-default off",
            "set $target_riscv=1",
            "set architecture riscv:rv64"
        ],
        "preRunCommands": [
            "thread apply all set $pc=_start",
            "thread apply all set $mie=0",
            "thread apply all set $mip=0",
            "thread apply all set $mstatus=0"
        ],
        "imageAndSymbols": {
            "symbolFileName": "path/to/executable",
            "imageFileName": "path/to/executable"
        },
        "target": {
            "host": "localhost",
            "port": "3333",
            "server": "path/to/openocd",
            "serverParameters": ["--command", "set DEVICE MPFS", "--file", "board/microsemi-riscv.cfg", "--command", "set COREID 1"],
            "serverStartupDelay": 5000,
            "cwd": "${workspaceFolder}"
        }
    }]
    }
```

ArsenInstigate:
Hi @jonahgraham. I override disconnectRequest in GDBTargetDebugSession.ts of cdt-gdb-adapter.

`
    protected async disconnectRequest(
        response: DebugProtocol.DisconnectResponse,
        _args: DebugProtocol.DisconnectArguments
[...]

PR #260, opened 2023-03-06
Support kill() process when GDB does not exit

bangdang1411:
Following the discussion in PR #256
I would like to commit the new change about it. I think it is easy to review because the change is different from #256

Solution: Override disconnectRequest() with option exitGDBServer() if this request does not use for restart Request. It has some benefits:
1. disconnectRequest does not wait over 2 seconds for response. exitGDBServer() will handle separately with the request. 
2. restartRequest will be implemented in RenesasGDBTargetDebugSession, that mean it is not depended on timeout of others Request.
3. does not need to implement terminateRequest because it seems cannot update CallStack view when GDB exited.

### DIFF
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -19,10 +19,6 @@ import * as mi from './mi';
 import { DebugProtocol } from '@vscode/debugprotocol';
 import { spawn, ChildProcess } from 'child_process';
 
-function delay(ms: number) {
-    return new Promise( resolve => setTimeout(resolve, ms) );
-}
-
 export interface TargetAttachArguments {
     // Target type default is "remote"
     type?: string;
@@ -81,6 +77,7 @@ export interface TargetLaunchRequestArguments
 export class GDBTargetDebugSession extends GDBDebugSession {
     protected gdbserver?: ChildProcess;
     protected isServerTerminated: boolean = false;
+    protected waitTerminated?: (value?: void | PromiseLike<void>) => void;
 
     protected async attachOrLaunchRequest(
         response: DebugProtocol.Response,
@@ -242,6 +239,10 @@ export class GDBTargetDebugSession extends GDBDebugSession {
 
             this.gdbserver.on('exit', (code) => {
                 this.isServerTerminated = true;
+                if (this.waitTerminated){
+                    this.waitTerminated();
+                    this.waitTerminated = undefined;
+                }
                 const exitmsg = `${serverExe} has exited with code ${code}`;
                 this.sendEvent(new OutputEvent(exitmsg, 'server'));
                 if (!gdbserverStartupResolved) {
@@ -355,8 +356,10 @@ export class GDBTargetDebugSession extends GDBDebugSession {
     ): Promise<void> {
         try {
             await this.gdb.sendGDBExit();
-            if (!_args.restart) {
-                this.exitGDBServer();
+            // Kill a process in case of termination, should not in case of restart
+            if (_args.restart !== true) {
+                // Call async function for checking process exit gdb after send response to client
+                this.exitGDBServer(true);
             }
             this.sendResponse(response);
         } catch (err) {
@@ -368,16 +371,23 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         }
     }
 
-    protected async exitGDBServer(): Promise<void> {
-        if (!this.isServerTerminated) {
-            let timeout = setTimeout(() => {
+    public async exitGDBServer(isKillProcess: boolean): Promise<void> {
+        if (this.isServerTerminated) {
+            return;
+        }
+        let timeout;
+        if (isKillProcess) {
+            // Kill process after 4 seconds timeout
+            timeout = setTimeout(() => {
                 this.gdbserver?.kill();
                 return;
-            }, 2000);
-            while (!this.isServerTerminated) {
-                await delay(200);
-            }
-            clearTimeout(timeout);
+            }, 4000);
         }
+        const waitTerminated = new Promise<void>((resolve) => {
+            this.waitTerminated = resolve;
+        });
+        // Waiting for event 'exit' to resolve the promise
+        await waitTerminated;
+        if (timeout) clearTimeout(timeout);
     }
 }
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -371,7 +371,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         }
     }
 
-    public async exitGDBServer(isKillProcess: boolean): Promise<void> {
+    protected async exitGDBServer(isKillProcess: boolean): Promise<void> {
         if (this.isServerTerminated) {
             return;
         }
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -19,6 +19,10 @@ import * as mi from './mi';
 import { DebugProtocol } from '@vscode/debugprotocol';
 import { spawn, ChildProcess } from 'child_process';
 
+function delay(ms: number) {
+    return new Promise( resolve => setTimeout(resolve, ms) );
+}
+
 export interface TargetAttachArguments {
     // Target type default is "remote"
     type?: string;
@@ -76,6 +80,7 @@ export interface TargetLaunchRequestArguments
 
 export class GDBTargetDebugSession extends GDBDebugSession {
     protected gdbserver?: ChildProcess;
+    protected isServerTerminated: boolean = false;
 
     protected async attachOrLaunchRequest(
         response: DebugProtocol.Response,
@@ -177,6 +182,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         // Wait until gdbserver is started and ready to receive connections.
         await new Promise<void>((resolve, reject) => {
             this.gdbserver = spawn(serverExe, serverParams, { cwd: serverCwd });
+            this.isServerTerminated = false;
             let gdbserverStartupResolved = false;
             let accumulatedStderr = '';
             let checkTargetPort = (_data: any) => {
@@ -235,6 +241,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             }
 
             this.gdbserver.on('exit', (code) => {
+                this.isServerTerminated = true;
                 const exitmsg = `${serverExe} has exited with code ${code}`;
                 this.sendEvent(new OutputEvent(exitmsg, 'server'));
                 if (!gdbserverStartupResolved) {
@@ -339,6 +346,38 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                 1,
                 err instanceof Error ? err.message : String(err)
             );
+        }
+    }
+    
+    protected async disconnectRequest(
+        response: DebugProtocol.DisconnectResponse,
+        _args: DebugProtocol.DisconnectArguments
+    ): Promise<void> {
+        try {
+            await this.gdb.sendGDBExit();
+            if (!_args.restart) {
+                this.exitGDBServer();
+            }
+            this.sendResponse(response);
+        } catch (err) {
+            this.sendErrorResponse(
+                response,
+                1,
+                err instanceof Error ? err.message : String(err)
+            );
+        }
+    }
+
+    protected async exitGDBServer(): Promise<void> {
+        if (!this.isServerTerminated) {
+            let timeout = setTimeout(() => {
+                this.gdbserver?.kill();
+                return;
+            }, 2000);
+            while (!this.isServerTerminated) {
+                await delay(200);
+            }
+            clearTimeout(timeout);
         }
     }
 }
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -18,6 +18,7 @@ import {
 import * as mi from './mi';
 import { DebugProtocol } from '@vscode/debugprotocol';
 import { spawn, ChildProcess } from 'child_process';
+import * as os from 'os';
 
 export interface TargetAttachArguments {
     // Target type default is "remote"
@@ -76,8 +77,7 @@ export interface TargetLaunchRequestArguments
 
 export class GDBTargetDebugSession extends GDBDebugSession {
     protected gdbserver?: ChildProcess;
-    protected isServerTerminated: boolean = false;
-    protected waitTerminated?: (value?: void | PromiseLike<void>) => void;
+    protected waitTerminated?: (value?: void | PromiseLike<void>) => void; 
 
     protected async attachOrLaunchRequest(
         response: DebugProtocol.Response,
@@ -179,7 +179,6 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         // Wait until gdbserver is started and ready to receive connections.
         await new Promise<void>((resolve, reject) => {
             this.gdbserver = spawn(serverExe, serverParams, { cwd: serverCwd });
-            this.isServerTerminated = false;
             let gdbserverStartupResolved = false;
             let accumulatedStderr = '';
             let checkTargetPort = (_data: any) => {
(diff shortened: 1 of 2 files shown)


bangdang1411:
Hi @jonahgraham-san,
I have read this discussion in #256 and #258 PR and would like to commit my solution.
This commit can fix for Renesas devices, but it needs to be reviewed and confirmed for issue #258.
The fix for issue #256 about "restartRequest" must change some things in RenesasGDBTargetDebugSession and override the restartRequest() function.
Thank you.

jonahgraham:
If you are going to override `exitGDBServer` in Renesas's backend, why not just override `disconnectRequest` entirely then? Then we can merge #258 for the general case and Renesas can do whatever it wants?

PS. If one of the reasons you want to change it here is for the `isServerTerminated`, then I suggest you poll `this.gdbserver.exitCode`, or better add an additional listener on gdbserver's `'exit'` (and have it set a promise so you can remove polling altogether)

bangdang1411:
Hi @jonahgraham-san,
> If you are going to override exitGDBServer in Renesas's backend, why not just override disconnectRequest entirely then? Then we can merge https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/258 for the general case and Renesas can do whatever it wants.

My solution is not to override **exitGDBServer()** and **disconnectRequest** in Renesas's backend. I will override **restartRequest**, please see the commit in renesas-gdb-adapter that I comment in the Jira ticket.
This commit in cdt-gdb-adapter will fix the issue of #258, It supports kill gdb process after 4 seconds timeout (it is not dependent on 2 seconds timeout of **disconnectRequest**).
The bug restart Renesas devices will use this commit to implement **restartRequest**.

> PS. If one of the reasons you want to change it here is for the isServerTerminated, then I suggest you poll this.gdbserver.exitCode, or better add an additional listener on gdbserver's 'exit' (and have it set a promise so you can remove polling altogether)

Thank you for your suggestion. Please help me review the latest commit and give me your idea.
Do we need to write the integration test for it? Thank you.

src/GDBTargetDebugSession.ts
```diff
@@ -235,6 +238,11 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             }
 
             this.gdbserver.on('exit', (code) => {
+                this.isServerTerminated = true;
```
jonahgraham, line 241: Does my earlier suggestion of `this.gdbserver.exitCode` not work?

src/GDBTargetDebugSession.ts
```diff
@@ -341,4 +349,45 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
     }
+    
+    protected async disconnectRequest(
+        response: DebugProtocol.DisconnectResponse,
+        _args: DebugProtocol.DisconnectArguments
+    ): Promise<void> {
+        try {
+            await this.gdb.sendGDBExit();
+            // Kill a process in case of termination, should not in case of restart
+            if (_args.restart !== true) {
```
jonahgraham, line 360: I don't think I understand the `restart` flow. Why do you leave the gdb server running if restart is true?

src/GDBTargetDebugSession.ts
```diff
@@ -341,4 +349,45 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+                // Call async function for checking process exit gdb after send response to client
```
jonahgraham, line 361: Does this work? IIUC VSCode will kill the process tree after the disconnect response? See the VSCode source [here](https://github.com/Microsoft/vscode/blob/e9229a18805fdb9baf7356ddc31bd216da5842ef/src/vs/workbench/contrib/debug/browser/rawDebugSession.ts#L572-L576) and [here](https://github.com/Microsoft/vscode/blob/990a496681948ec274da5503b35cd49a7d06c0a2/src/vs/workbench/contrib/debug/node/debugAdapter.ts#L268-L285)

src/GDBTargetDebugSession.ts
```diff
@@ -341,4 +349,45 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+
+    protected async exitGDBServer(isKillProcess: boolean): Promise<void> {
+        if (this.isServerTerminated) {
+            return;
+        }
+        let timeout;
+        if (isKillProcess) {
+            // Kill process after 4 seconds timeout
+            timeout = setTimeout(() => {
+                this.gdbserver?.kill();
```
jonahgraham, line 382: On Linux this will (only) do a SIGTERM so the gdbserver may still be running which will lead to the await on line 390 to wait forever.

jonahgraham (commented):
> Do we need to write the integration test for it? Thank you.

Yes. If it helps, the answer to "do we need a test" is always yes, or an explanation as to why it is difficult or impractical to write an automated test.

src/GDBTargetDebugSession.ts
```diff
@@ -341,4 +349,45 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+        } catch (err) {
+            this.sendErrorResponse(
+                response,
+                1,
+                err instanceof Error ? err.message : String(err)
+            );
+        }
+    }
+
+    protected async exitGDBServer(isKillProcess: boolean): Promise<void> {
```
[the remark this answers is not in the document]
bangdang1411: I think kill process that should be used in case of **disconnect** or **terminate** the gdb, should not be in case of restart.
If GDB is exited by kill, that means it cannot debug again if user does not reset the voltage in the emulator (Checking by Renesas devices: RL78, RH850,..)
In the case of restart request, isKillProcess is set false (in Renesas backend) to wait for GDB is exited normally.

src/GDBTargetDebugSession.ts
bangdang1411: I think using `isServerTerminated` that can be used in Renesas's backend also.
I do not use `this.gdbserver.exitCode` but I think it is not different.

src/GDBTargetDebugSession.ts
bangdang1411: In the **restart** flow, I think GDB should be exited normally. Kill process that makes it cannot launch again in case of some hardware device (Renesas RL78, RH850,..)

src/GDBTargetDebugSession.ts
bangdang1411: I think yes, the async function monitors the process of GDB in server, I think it is asynchronous with the process tree in VS Code.
Is my understanding correct?

src/GDBTargetDebugSession.ts
bangdang1411: Does Linux not have an event `exit` to resolve this promise and we cannot use kill() gdb process? Is my understanding correct?
Can you share the information about it in Linux? 
Sorry I have not understood in Linux environment.

src/GDBTargetDebugSession.ts
jonahgraham: But renesas backend can use this.gdbserver.exitCode and then it removes this redundant code.

src/GDBTargetDebugSession.ts
jonahgraham: I don't really follow, but I assume I will understand once I see the rest.

src/GDBTargetDebugSession.ts
jonahgraham: See this reasonable explanation of sigterm vs sigkill. https://linuxhandbook.com/sigterm-vs-sigkill/

jonahgraham: Anyways, lets put the Linux support to the side for the moment and get the Windows one working properly.

bangdang1411: Thank you for your information. I will check for both Linux and Window.

src/GDBTargetDebugSession.ts
bangdang1411: >But can you actually see the node process for the adapter running for the 4 seconds after disconnect returns?

I test with RL78 device in some PCs that gdb is killed after 4 seconds and the emulator hangs. It cannot debug again until the voltage in emulator resets.

bangdang1411: I think I should check the node process again. Maybe I misunderstand in some where.

bangdang1411:
Hi @jonahgraham-san,
The latest commit fix both issue `Terminate `and `Restart `GDB server. It does not need to implement in `Renesas's backend`.
I see that cdt-gdb-adapter takes over 4 seconds to send `disconnect response` to vs code has no error returned.
I think the node process is not killed at this time and we can wait for the exiting gdb process completed.
```
[07:45:05.525 UTC] From client: disconnect({"restart":false})
[07:45:05.527 UTC] GDB command: 49 -gdb-exit
[07:45:05.543 UTC] GDB result: 49 exit
[07:45:05.543 UTC] GDB notify async: thread-group-exited,id="i1"
[07:45:08.535 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"server","output":"Disconnected from the Target Debugger.\r\n"}}
[07:45:08.551 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"server","output":"e2-server-gdb has exited with code 0"}}
[07:45:08.551 UTC] To client: {"seq":0,"type":"response","request_seq":10,"command":"disconnect","success":true}
```

src/GDBTargetDebugSession.ts
bangdang1411: My Linux environment has some config error, so I cannot test in the case of Linux.
Could you please help me take a look at my code, is send `kill('SIGTERM')` match your idea in Linux?
Or not I will fix my environment tomorrow and retest it again. Thank you.

src/GDBTargetDebugSession.ts
bangdang1411: I can confirm the node process still existed after 4 seconds. But I am not sure how `stopSession()` in vs code takes much time to kill this process.

src/GDBTargetDebugSession.ts
bangdang1411: I changed it in the latest commit.

src/GDBTargetDebugSession.ts
```diff
@@ -341,4 +347,45 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+            return;
+        }
+        let timeout;
+        if (isKillProcess) {
+            // Kill process after 4 seconds timeout
+            timeout = setTimeout(() => {
+                if (os.platform() === 'win32') {
+                    this.gdbserver?.kill();
+                } else {
+                    this.gdbserver?.kill('SIGTERM');
```
jonahgraham, line 380: Please check [documentation again](https://nodejs.org/api/child_process.html#subprocesskillsignal), on Linux SIGTERM is the default, that is why I was raising that as an issue as kill() on Windows is equivalent to SIGKILL, but kill() on Linux is SIGTERM.

jonahgraham (commented):
I need to spin this up as I want to understand why the code in VSCode is not being observed. That will take me another day.

bangdang1411:
> I need to spin this up as I want to understand why the code in VSCode is not being observed. That will take me another day.

I try to change the code that does not send the response for` disconnect request`. In my environment, after 6 seconds VS Code will kill the node process (I assume the time is depended on the performance of PC). I think we need to extend the timeout on VS Code side. But this code can work for now and the change should be in VSCode repository.

> Please check [documentation again](https://nodejs.org/api/child_process.html#subprocesskillsignal), on Linux SIGTERM is the default, that is why I was raising that as an issue as kill() on Windows is equivalent to SIGKILL, but kill() on Linux is SIGTERM.

Sorry, I misunderstood your question. SIGTERM is only supported in Linux but it is can be ignored and the Promise that I implemented will wait forever. Is it correct? 
I changed both case Windows and Linux to 'SIGKILL'. My understanding is 'SIGKILL' can work in both environments and cannot be ignored.

jonahgraham:
> In my environment, after 6 seconds VS Code will kill the node process (I assume the time is depended on the performance of PC).

I have tested this on a couple of computers and it seems that node is killed in about 2 seconds on each of them.

However, I have a new proposal on how to resolve this problem. Instead of running cdt-gdb-adapter as a separate node process, we can run it in the extension host process. That means that the node process won't be killed at all when disconnect completes and we can do whatever we want. However this moves the problem in part to VSCode extension. Please look up DebugAdapterInlineImplementation in https://code.visualstudio.com/api/extension-guides/debugger-extension

In parallel, please raise an issue with vscode, perhaps with a PR, to remove the 2 second kill.

bangdang1411:
Hi @jonahgraham-san,

> However, I have a new proposal on how to resolve this problem. Instead of running cdt-gdb-adapter as a separate node process, we can run it in the extension host process. That means that the node process won't be killed at all when disconnect completes and we can do whatever we want. However this moves the problem in part to VSCode extension. Please look up DebugAdapterInlineImplementation in https://code.visualstudio.com/api/extension-guides/debugger-extension

I understand your idea but it takes time to investigate and implement. Anyway, It should be implemented in Renesas vs code.
Is that correct? Additionally, when it was implemented, this PR should be merged to fix the issue gdb does not exit normally after 4 seconds.

> In parallel, please raise an issue with vscode, perhaps with a PR, to remove the 2 second kill.

I see the [same requirement](https://github.com/microsoft/vscode/issues/175850), but it is discussing in [another issue](https://github.com/microsoft/vscode/issues/172284)

Could you please share your idea? Thank you.

jonahgraham:
> Hi @jonahgraham-san,
> 
> > However, I have a new proposal on how to resolve this problem. Instead of running cdt-gdb-adapter as a separate node process, we can run it in the extension host process. That means that the node process won't be killed at all when disconnect completes and we can do whatever we want. However this moves the problem in part to VSCode extension. Please look up DebugAdapterInlineImplementation in https://code.visualstudio.com/api/extension-guides/debugger-extension
> 
> I understand your idea but it takes time to investigate and implement. Anyway, It should be implemented in Renesas vs code. Is that correct? 

The main change in cdt-gdb-vscode (and any extensions that use cdt-gdb-adapter too, line renesas-gdb-vscode). It needs review work up and down the entire stack because, for example, cdt-gdb-adapter was written assuming that it would be in its own process. For the most part it probably will work, but the entry points and cleanup especially will need some investigating to be sure.

> Additionally, when it was implemented, this PR should be merged to fix the issue gdb does not exit normally after 4 seconds.

I don't think so, at least not as is. Renesas can do a workaround for the issues that is affecting them in their own adapter. That is to say, this entire PR can just be done in Renesas's own code base as it is a workaround for Renesas's issue that does not work in the general case.

> 
> > In parallel, please raise an issue with vscode, perhaps with a PR, to remove the 2 second kill.
> 
> I see the [same requirement](https://github.com/microsoft/vscode/issues/175850), but it is discussing in [another issue](https://github.com/microsoft/vscode/issues/172284)
> 
> Could you please share your idea? Thank you.

I just read that and it is on point. Of course someone needs to provide a PR to improve the situation there based on what is discussed.

bangdang1411:
> The main change in cdt-gdb-vscode (and any extensions that use cdt-gdb-adapter too, line renesas-gdb-vscode). It needs review work up and down the entire stack because, for example, cdt-gdb-adapter was written assuming that it would be in its own process. For the most part it probably will work, but the entry points and cleanup especially will need some investigating to be sure.

I am investigating the [VS Code Mock Debug](https://github.com/microsoft/vscode-mock-debug) and trying to know how it works. But it seems more difficult than my experience. Could you please help me take a look at it and give me some ideas on how to implement your solution? Thank you.

> I don't think so, at least not as is. Renesas can do a workaround for the issues that is affecting them in their own adapter. That is to say, this entire PR can just be done in Renesas's own code base as it is a workaround for Renesas's issue that does not work in the general case.

I understood. But please consider if we have a short-term solution for fixing the restart bug in Renesas. It should be fixed in the next release (no more time). Thank you.

> I just read that and it is on point. Of course, someone needs to provide a PR to improve the situation there based on what is discussed.

My understanding in that discussion is requested to improve the UX when the termination process is more than 2s. I think removing the 2-second kill is not acceptable.

jonahgraham:
I think the solution properly belongs in the Renesas extension so I am going to close this for now. If there is a generic version please provide it.

Please also see https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/253#issuecomment-1480059599 for the next time you update to newer cdt-gdb-adapter.

Issue #261, opened 2023-04-11
Unable to attach to remote gdbserver


sherrellbc:
When attempting to provide a config that would instruct the adapter to attach to a remote GDB server it seems to fail. I tried the following:

```
cdtDebugTargetAdapter --server=50000 --config-frozen='{"request": "attach", "target" : {"host":"x.x.x.x", "port":"yyyy"}}'
```

As well as

```
cdtDebugTargetAdapter --server=50000 --config-frozen=@config.json
```

With `config.json` having contents of:

```
{
   "request": "attach",
   "target" : {"host":"x.x.x.x", "port":"yyyy"}
}
```

but again, no dice. I am attempting to use the plugin with [Arm's Embedded Debugger](https://github.com/ARM-software/vscode-embedded-debug) for VSCode. When I run with the above configuration I am met with an error dialog in VSCode that simple states "arg is not iterable".

On my test gdbserver instance I do not see any client connections coming in. It is somewhat unclear if the issue is in this plugin or within Arm's VSCode plugin. However, there are no examples in this repo for how to use the `attach` request so my formation of the config is somewhat contrived based on my interpretation of the sources. Other formats I tried:

```
--config-frozen='{"request": "attach", "parameters":"x.x.x.x:yyyy"}'
--config-frozen='{"request": "attach", "host":"x.x.x.x", "port":"yyyy"}'
--config-frozen='{"request": "attach", "connectCommands":"target remote x.x.x.x:yyyy"}'
```

jonahgraham:
@thegecko do you think you can help @sherrellbc on this?

sherrellbc:
Within VSCode, my `launch.json` contains:
```
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Attach WebSocket",
            "type": "arm-debug",
            "request": "attach",
            "address": "ws://localhost:50001",
            "pathMapping": {
                "/home": "${workspaceFolder}"
            }
        }
    ]
```

and the adapter config is:

```
cdtDebugTargetAdapter --server=50000 --config-frozen='{"request": "attach", "program": "/home/hello", "host":"x.x.x.x", "port":"yyyy"}'
```

With a websockify proxy listening on 50001 and forwarding to 50000. With the addition of `program` I now see in VSCode a different failure where it says

"To open a remote debug connection, you need to specify what serial device is attached to the remote system"

sherrellbc:
[...]

PR #262, opened 2023-04-13
Allow attaching (remote and local) with no program

jonahgraham:
There is no strict need to have a program passed to GDB when attaching (remote or local). Either the program can be downloaded from the remote or otherwise read by GDB, or if not possible then GDB can do symbol-free debugging.

Previously in #228 some work was done to better error when program was not specified. This change loosens the restrictions slightly. If the program is not specified, it is still a warning in VSCode because of
[package.json](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/blob/22a9f1048f9030689ed1f5f2b3e5d5e48df9047d/package.json#L205-L207)

See this comment for how it is presented to a user: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/261#issuecomment-1505209833

Part of #261


WyoTwT:
This is a useful PR as I have a need for this functionality. Thank you

jonahgraham:
@WyoTwT if you can review this change and see if anything is missing, please let me know.

WyoTwT:
This PR looks good to me.  Our tree has some minor differences but they're specific to our implementation.

chroberino:
@jonahgraham, @WyoTwT, I am wondering about the status and priority of this PR. I'd really vote for integrating this, since in my remote debugging setup I need to put a dummy.o file into the `program` property.

chroberino:
It seems the feature request from this PR has already been contributed.
At least the documentation for the `gdbtarget` type (https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode?tab=readme-ov-file#gdbtarget-debugger-type) says:
> While `program` is marked as required, the debug adapter launches anyway for remote GDB connections. For example to inspect an embedded target system's memory and other hardware resources without debugging a program.

I have successfully tested it without specifying a `program`.
=> @jonahgraham I think this PR can be closed.

src/GDBTargetDebugSession.ts
```diff
@@ -275,7 +275,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         try {
             this.isAttach = true;
             await this.spawn(args);
-            await this.gdb.sendFileExecAndSymbols(args.program);
+            if (args.program) {
```
jreineckearm, line 278: We could also consider to similarly handle an empty string ('').

jreineckearm:
@jonahgraham , do you see it as realistic to bring this PR to completion in the near future? Or should we take over/"reimplement" and refine based on your solution. We see relaxing this behavior as important for a couple of remote target use cases on our end:
* Multi-core and other connection use cases which require scripting before automated loading of programs.
* Early board bring-up without a loading an application where you as a first step would just like to connect and see something kicking in the target system.
* Post-mortem analysis of a failing program execution where you in a first instance would just like to inspect some system resources (peripheral registers) after a connection without accidentally changing the target state by a program load.

PR #264, opened 2023-04-28
Support disable/enable hardware breakpoint

trongle0504:
Currently, enabling hardwareBreakpoint in launch.json argument allows the debugger to set hardware break-point or software break-point only ("hardwareBreakpoint": true). We need to support enabling/disabling setting SW and HW break-point simultaneously while debugging.


trongle0504:
Hi @jonahgraham ,

Could you please help me review the code?
Please see the document attached to ticket IDE-63527 for details.

I also commit code in the repository cdt-gdb-vscode.
https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/89 

Thank you!

src/GDBBackend.ts
```diff
@@ -165,9 +165,12 @@ export class GDBBackend extends events.EventEmitter {
         return this.gdbNonStop;
     }
 
-    public isUseHWBreakpoint() {
+    public getHWBreakpoint() {
```
jonahgraham, line 168: I don't see a reason to rename this method - boolean results should be generally prefixed with `is` anywhay.

jonahgraham (changes requested):
> Please see the document attached to ticket IDE-63527 for details.

Please provide required information in the open - this is an open source project so should not require access to private information to understand this.

jonahgraham (changes requested):
Please add a test for changing the hw breakpoint too.

src/GDBBackend.ts
trongle0504: I updated it.

src/GDBBackend.ts
```diff
@@ -165,9 +165,12 @@ export class GDBBackend extends events.EventEmitter {
...
         return this.hardwareBreakpoint;
     }
+    public setHWBreakpoint() {
```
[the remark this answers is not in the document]
trongle0504: I changed the solution to use `Toggle Breakpoint` so I changed the name of the function here to `toggleBreakpoint()`

trongle0504:
> Please provide required information in the open - this is an open source project so should not require access to private information to understand this.

Hi @jonahgraham ,
I updated the details information for this PR at https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/89 
Please check it out for details.

Thank you!

src/GDBBackend.ts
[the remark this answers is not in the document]
jonahgraham: No - having a toggle is not an acceptable solution. Making it a toggle means there is no way to programatically ensure that it is in a specific state.

The UI can present to the user as toggle, but the backend needs to be a set.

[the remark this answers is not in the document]
trongle0504: I updated toggle to set breakpoint. 
But now I'm having build failure problem as below.

src/GDBDebugSession.ts
```diff
@@ -289,6 +289,18 @@ export class GDBDebugSession extends LoggingDebugSession {
                             : `Encountered a problem executing ${args.command}`;
                     this.sendErrorResponse(response, 1, message);
                 });
+        } else if (command === 'cdt-gdb-adapter/getHWBreakpoint') {
+            response.body = this.gdb.isUseHWBreakpoint();
+            this.sendResponse(response);
+        } else if (command === 'cdt-gdb-adapter/setHWBreakpoint') {
+            if (typeof args.isSetHWBrekpoint !== 'boolean') {
+                throw new Error(
+                    `Invalid type for 'isSetHWBrekpoint', expected string, got ${typeof args.status}`
```
jonahgraham, line 298:
```suggestion
                    `Invalid type for 'isSetHWBrekpoint', expected boolean, got ${typeof args.status}`
```

src/GDBDebugSession.ts
```diff
@@ -289,6 +289,18 @@ export class GDBDebugSession extends LoggingDebugSession {
                             : `Encountered a problem executing ${args.command}`;
                     this.sendErrorResponse(response, 1, message);
                 });
+        } else if (command === 'cdt-gdb-adapter/getHWBreakpoint') {
+            response.body = this.gdb.isUseHWBreakpoint();
+            this.sendResponse(response);
+        } else if (command === 'cdt-gdb-adapter/setHWBreakpoint') {
+            if (typeof args.isSetHWBrekpoint !== 'boolean') {
+                throw new Error(
```
jonahgraham, line 297: Should this throw an error or do `sendErrorResponse` similar to line 290 above.

jonahgraham (changes requested):
Minor changes + needs testing still.

I have some concerns also regarding how the hardware breakpoints are managed, but much of that is in the front end so I will continue the conversation on that in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/89

trongle0504:
Hi @jonahgraham,
Could you please help me review the solution for save breakpoints after disconnect debug session first? Others, I will update later.
Thank you!

src/GDBDebugSession.ts
```diff
@@ -496,6 +513,8 @@ export class GDBDebugSession extends LoggingDebugSession {
             // Need to get the list of current breakpoints in the file and then make sure
             // that we end up with the requested set of breakpoints for that file
             // deleting ones not requested and inserting new ones.
+            const rootFolder = path.dirname(args.source.path as string);
```
jonahgraham, line 516: Use line 520's `file` here as input to `path.dirname`

src/GDBDebugSession.ts
```diff
@@ -615,13 +634,28 @@ export class GDBDebugSession extends LoggingDebugSession {
                 }
 
                 try {
+                    let newDataIndex = -1;
+                    let existingData: DataObject[] = [];
+                    if (fs.existsSync(filePath)) {
+                        const fileContent = fs.readFileSync(filePath, 'utf-8');
+                        existingData = JSON.parse(fileContent) as DataObject[];
+                        newDataIndex = existingData.findIndex((item) => {
```
jonahgraham, line 642: Instead of using an index here and then having to maintain two variables for the lifetime, have the variable at line 637 be of type `DataObject` and start it `undefined` and in this code set that variable if it is found.

src/GDBDebugSession.ts
```diff
@@ -124,6 +124,11 @@ class ThreadWithStatus implements DebugProtocol.Thread {
         this.running = running;
     }
 }
+interface DataObject {
+    source: string;
+    line: number;
+    type: string;
```
jonahgraham, line 130: Rather than store this as a string, how about make this "hwbreakpoint: boolean"?

PS calling the field `type` is not recommended as type is sometimes special in typescript.

jonahgraham (changes requested):
This is an interesting approach and I look forward to seeing how you evolve it.

Some high level things:

1. You need to provide some tests - this functionality is complicated 
2. `BreakpointSettings.json` appears to being placed in the same folder as the source, but when it is written it contains all the source files, regardless of the folder. Tests need to cover cases where there is > 1 file and > 1 folder those files are in
3. It may be better to have a path to  `BreakpointSettings.json` stored in the launch arguments, with the default being storing it in the same folder as `launch.json`. The cdt-gdb-vscode can be enhanced to provide the correct default.
4. Because the BreakpointSettings file is being stored in a user location, some additional handling on format of the file is needed. e.g. test to make sure that the adapter handles an error in that file's format properly. 
5. Similar handling is needed for function breakpoints
6. Please address the previous comments

trongle0504:
I'm sorry, I would like to close this request because there are some complex cases to cover, and it may cause inconvenience for users. Thank you for taking the time to review the code.

PR #266, opened 2023-04-28
Add some help / info to user so they know about new feature.

jonahgraham:
This is a follow up for #263


jonahgraham:
Folded back into #263

PR #267, opened 2023-05-31
The incorrect state of multicore in Call Stack View

trongle0504:
When starting multicore with debugger type "amalgamator", the state of multicore was displayed in Call Stack View is incorrect.

Due to difference about the information of thread got from threadsRequest, the order of threads to store information for corresponding core at initialization and when got from GDB server is different.

So, the state of multicore was displayed in Call Stack View are incorrectly updated.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -861,19 +861,12 @@ export class GDBDebugSession extends LoggingDebugSession {
     ): Promise<void> {
         try {
             if (!this.isRunning) {
-                const result = await mi.sendThreadInfoRequest(this.gdb, {});
-                const threadMapInfo = result.threads.map((self) =>
-                    this.convertThread(self)
-                );
-                this.threads.map((thread) => {
-                    const threadInfo = threadMapInfo.find(
-                        (self) => self.id === thread.id
-                    );
-                    if (threadInfo !== undefined) {
-                        thread.name = threadInfo.name;
-                        thread.running = threadInfo.running;
-                    }
-                });
+                if (!this.isRunning) {
+                    const result = await mi.sendThreadInfoRequest(this.gdb, {});
+                    this.threads = result.threads.map((thread) =>
+                        this.convertThread(thread)
+                    ).sort((a, b) => a.id - b.id);
+                }
             }
 
             response.body = {
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -863,7 +863,7 @@ export class GDBDebugSession extends LoggingDebugSession {
             const result = await mi.sendThreadInfoRequest(this.gdb, {});
             this.threads = result.threads.map((thread) =>
                 this.convertThread(thread)
-            ).sort((a, b) => a.id - b.id);
+            ).sort((a, b) => a.id - b.id); 
 
             response.body = {
                 threads: this.threads,
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -860,12 +860,10 @@ export class GDBDebugSession extends LoggingDebugSession {
         response: DebugProtocol.ThreadsResponse
     ): Promise<void> {
         try {
-            if (!this.isRunning) {
-                const result = await mi.sendThreadInfoRequest(this.gdb, {});
-                this.threads = result.threads.map((thread) =>
-                    this.convertThread(thread)
-                ).sort((a, b) => a.id - b.id);
-            }
+            const result = await mi.sendThreadInfoRequest(this.gdb, {});
+            this.threads = result.threads.map((thread) =>
+                this.convertThread(thread)
+            ).sort((a, b) => a.id - b.id);
 
             response.body = {
                 threads: this.threads,
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -860,10 +860,12 @@ export class GDBDebugSession extends LoggingDebugSession {
         response: DebugProtocol.ThreadsResponse
     ): Promise<void> {
         try {
-            const result = await mi.sendThreadInfoRequest(this.gdb, {});
-            this.threads = result.threads.map((thread) =>
-                this.convertThread(thread)
-            ).sort((a, b) => a.id - b.id); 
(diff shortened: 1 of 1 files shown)


jonahgraham (changes requested):
I think we should be keeping the thread order the same as GDB provides as to match the `info threads` output.

It sounds like you have a case where GDB is returning threads in a random(ish)/changing order?

If this is strictly related to the amalgamator, perhaps there is a bug in the amalgamator where it is assuming that the order of threads returned is stable. If so,it should be using the thread's id to match them on subsequent calls.

Another issue with this code is that it discards threads that `-thread-info` returns if they weren't in the threads map already. This is a problem for targets that don't send thread created events. Unfortunately we don't have a test for that.

If we really need to keep a consistent order here them we should simply sort the return value (`result.threads`) by thread id. But I need an explanation why we should be overriding GDB's ordering.

trongle0504:
> It sounds like you have a case where GDB is returning threads in a random(ish)/changing order?

We have a case where GDB is returning threads in a changing order compared to the original order.
MicrosoftTeams-image (5)


> If we really need to keep a consistent order here them we should simply sort the return value (`result.threads`) by thread id. But I need an explanation why we should be overriding GDB's ordering.

The thread order after initializing when starting multicore is set with the corresponding frameID, and when GDB is returning value with the new thread order, It causes confusion about frameID for the respective threadID in the new order of the return value by GDB.

src/GDBDebugSession.ts
```diff
@@ -861,10 +861,12 @@ export class GDBDebugSession extends LoggingDebugSession {
...
-                const result = await mi.sendThreadInfoRequest(this.gdb, {});
-                this.threads = result.threads.map((thread) =>
-                    this.convertThread(thread)
-                );
+                if (!this.isRunning) {
+                    const result = await mi.sendThreadInfoRequest(this.gdb, {});
+                    this.threads = result.threads.map((thread) =>
+                        this.convertThread(thread)
+                    ).sort((a, b) => a.id - b.id);
+                }
```
jonahgraham, line 869: Looks like you have a small error with a double-check of `this.isRunning`
```suggestion
                const result = await mi.sendThreadInfoRequest(this.gdb, {});
                this.threads = result.threads.map((thread) =>
                    this.convertThread(thread)
                ).sort((a, b) => a.id - b.id);
```

jonahgraham:
> > It sounds like you have a case where GDB is returning threads in a random(ish)/changing order?
> 
> We have a case where GDB is returning threads in a changing order compared to the original order. MicrosoftTeams-image (5)

Do you know *why* GDB is returning in the different order? Normally I would assume the order of threads as returned by GDB is the **correct** order to display to users as that will match what is displayed by `info threads`.

> 
> > If we really need to keep a consistent order here them we should simply sort the return value (`result.threads`) by thread id. But I need an explanation why we should be overriding GDB's ordering.
> 
> The thread order after initializing when starting multicore is set with the corresponding frameID, and when GDB is returning value with the new thread order, It causes confusion about frameID for the respective threadID in the new order of the return value by GDB.

This is the part I don't understand - what is the confusion caused? Who is confused here? If it is amalgamator it sounds like bug in amalgamator. Is VSCode fine with the changing order?

jonahgraham (changes requested):

src/GDBDebugSession.ts
trongle0504: Hi @jonahgraham ,

I changed some code based on your suggestion, and with this change, it worked fine.
But when I commit the code, I get the build fails error, apparently due to a server problem. 
Could you please help me review it? And also if this code is merged, can you please build the latest version? we need it to test for an upcoming release.

Thank you so much!

jonahgraham: You seem to have not applied my suggestion properly. In commit [17f6704](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/267/commits/17f67042eef237cfd2676644f000f92b2a55a1c3) you applied my suggestion, but then in the subsequent commit you went further and removed `if (!this.isRunning) {`

This was the original comment:

> Looks like you have a small error with a double-check of this.isRunning

It was because you had this as the code:

```typescript
            if (!this.isRunning) {
                if (!this.isRunning) {
                    const result = await mi.sendThreadInfoRequest(this.gdb, {});
                    this.threads = result.threads.map((thread) =>
                        this.convertThread(thread)
                    ).sort((a, b) => a.id - b.id);
                }
            }
```

The tests are probably failing because you are trying to call thread info while target is running.

jonahgraham:
In addition to restoring one of the `if (!this.isRunning) {`:

> Please answers questions in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/267#issuecomment-1572533406

trongle0504:
> This is the part I don't understand - what is the confusion caused? Who is confused here? If it is amalgamator it sounds like bug in amalgamator. Is VSCode fine with the changing order?

I was a bit confused. The cause of confusion here is that the FrameID has not changed the order according to the threadID in the new order of the value returned by GDB. It's true that this bug should be fixed in amalgamator, but I still couldn't find a way to change it in amalgamator, so I used cdt-gdb-adapter's threadsRequest() to change the order. 

I checked with this change that there will be a change in the display order in VSCode's Call Stack View, however it's fine.

jonahgraham:
Where in the amalgamator is it using the array index of the threads instead of the threadId of the thread. I'm guessing in `collectChildTheads`? 

I don't think the simple sorting by thread id in the adapter is a big problem, I still disagree with adding it as it leaves open holes related to any adapter that dynamically creates and deletes threads. However most of the time (all the time?) standard GDB returns threads in thread id order. Therefore I am going to merge this change as, compared to the first iteration of fixing this issue it seems straightforward enough.

The tests are all passing now that you have restored the `if (!this.isRunning) {`

jonahgraham (approved):

jonahgraham:
Latest versions now published:

- https://www.npmjs.com/package/cdt-gdb-adapter/v/0.0.23
- https://www.npmjs.com/package/cdt-amalgamator/v/0.0.11
- https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/tree/v0.0.99

PR #271, opened 2023-06-29
Adding the ability to configure serial port output to the debug console

AdhamRagabMCHP:
This PR introduces a set of parameters that allow us to configure a serial line in the adapter to print UART output from on-board/emulation debug to the debug console. These parameters include the TCP port to hook onto (for emulation programs outputting UART output), or the serial port on the host machine to capture UART output from. For the serial line, we also configure the baud rate, stop bits, handshaking method, etc.

The source code looks to take these parameters to then configure serial ports and sockets to read UART data and output it to the debug console. It will also set up necessary event handlers, and changes the log level in the non-verbose case to allow for "log" messages to be printed in a manner such that UART output isn't designated as a "warning".

~Lastly, GDBDebugSession.ts features an addition to remove a warning about an "unhandled notify" for the cmd-param-changed notification.~ (this part moved to #272)

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,9 @@
   "dependencies": {
     "@vscode/debugadapter": "^1.59.0",
     "@vscode/debugprotocol": "^1.59.0",
-    "node-addon-api": "^4.3.0"
+    "node-addon-api": "^4.3.0",
+    "serialport": "11.0.0",
+    "@types/serialport": "8.0.2"
   },
   "devDependencies": {
     "@types/chai": "^4.3.4",
--- a/package.json
+++ b/package.json
@@ -61,8 +61,7 @@
     "@vscode/debugadapter": "^1.59.0",
     "@vscode/debugprotocol": "^1.59.0",
     "node-addon-api": "^4.3.0",
-    "serialport": "11.0.0",
-    "@types/serialport": "8.0.2"
+    "serialport": "11.0.0"
   },
   "devDependencies": {
     "@types/chai": "^4.3.4",
@@ -70,6 +69,7 @@
     "@types/mocha": "^9.1.0",
     "@types/node": "^14.18.17",
     "@types/tmp": "^0.2.3",
+    "@types/serialport": "8.0.2",
     "@typescript-eslint/eslint-plugin": "^5.54.0",
     "@typescript-eslint/parser": "^5.54.0",
     "@vscode/debugadapter-testsupport": "^1.59.0",
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -495,7 +495,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
     ): Promise<void> {
         try {
             try {
-                SerialPort.close();
+                if (SerialPort.isOpen) SerialPort.close();
             } catch (err) {
                 logger.error((err as Error).message);
             }
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -21,6 +21,25 @@ import { spawn, ChildProcess } from 'child_process';
 import{ SerialPort, ReadlineParser } from "serialport";
 import { Socket } from "net";
 
+interface UARTArguments {
+    // Path to the serial port connected to the UART on the board.
+    serialPort?: string;
+    // Target TCP port on the host machine to attach socket to print UART output (defaults to 3456)
+    socketPort?: string;
+    // Baud Rate (in bits/s) of the serial port to be opened (defaults to 115200).
+    baudRate?: number;
+    // The number of bits in each character of data sent across the serial line (defaults to 8).
+    characterSize?: (5 | 6 | 7 | 8);
+    // The type of parity check enabled with the transmitted data (defaults to "none" - no parity bit sent)
+    parity?: ("none" | "even" | "odd" | "mark" | "space");
+    // The number of stop bits sent to allow the receiver to detect the end of characters and resynchronize with the character stream (defaults to 1).
+    stopBits?: (1 | 1.5 | 2);
+    // The handshaking method used for flow control across the serial line (defaults to "none" - no handshaking)
+    handshakingMethod?: ("none" | "XON/XOFF" | "RTS/CTS");
+    // The EOL character used to parse the UART output line-by-line.
+    eolCharacter?: ("LF" | "CRLF")
+}
+
 export interface TargetAttachArguments {
     // Target type default is "remote"
     type?: string;
@@ -34,22 +53,7 @@ export interface TargetAttachArguments {
     // Target connect commands - if specified used in preference of type, parameters, host, target
     connectCommands?: string[];
     // Settings related to displaying UART output in the debug console
-    uart?: {
-        // Path to the serial port connected to the UART on the board.
-        serialPort?: string;
-        // Target TCP port on the host machine to attach socket to print UART output (defaults to 3456)
-        socketPort?: string;
-        // Baud Rate (in bits/s) of the serial port to be opened (defaults to 115200).
-        baudRate?: number;
-        // The number of bits in each character of data sent across the serial line (defaults to 8).
-        characterSize?: (5 | 6 | 7 | 8);
-        // The type of parity check enabled with the transmitted data (defaults to "none" - no parity bit sent)
-        parity?: ("none" | "even" | "odd" | "mark" | "space");
-        // The number of stop bits sent to allow the receiver to detect the end of characters and resynchronize with the character stream (defaults to 1).
-        stopBits?: (1 | 1.5 | 2);
-        // The handshaking method used for flow control across the serial line (defaults to "none" - no handshaking)
-        handshakingMethod?: ("none" | "XON/XOFF" | "RTS/CTS");
-    }
+    uart?: UARTArguments;
 }
 
 export interface TargetLaunchArguments extends TargetAttachArguments {
@@ -288,6 +292,122 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         });
     }
 
+    protected initializeUARTConnection(
+        uart: UARTArguments,
+        host: string | undefined
+    ): void {
+        if (uart.serialPort !== undefined) {
+            // Set the path to the serial port
+            this.serialPort = new SerialPort({
+                path: uart.serialPort,
+                // If the serial port path is defined, then so will the baud rate.
+                baudRate: uart.baudRate ?? 115200,
+                // If the serial port path is deifned, then so will the number of data bits.
+                dataBits: uart.characterSize ?? 8,
+                // If the serial port path is defined, then so will the number of stop bits.
+                stopBits: uart.stopBits ?? 1,
+                // If the serial port path is defined, then so will the parity check type.
+                parity: uart.parity ?? 'none',
+                // If the serial port path is defined, then so will the type of handshaking method.
+                rtscts:
+                    uart.handshakingMethod === 'RTS/CTS'
+                        ? true
+                        : false,
+                xon:
+                    uart.handshakingMethod === 'XON/XOFF'
+                        ? true
+                        : false,
+                xoff:
+                    uart.handshakingMethod === 'XON/XOFF'
+                        ? true
+                        : false,
+                autoOpen: false,
+            });
+
+            this.serialPort.on("open", () => {
+                this.sendEvent(
+                    new OutputEvent(
+                        `listening on serial port ${this.serialPort?.path}`,
+                        "Serial Port"
+                    )
+                );
+            });
+
+            const SerialUartParser = new ReadlineParser({
+                delimiter: uart.eolCharacter === "LF" ? "\n" : "\r\n",
+                encoding: "utf8"
+            });
+
+            this.serialPort.pipe(SerialUartParser).on("data", (line: string) => {
+                this.sendEvent(
+                    new OutputEvent(line, "Serial Port")
+                );
+            });
+
+            this.serialPort.on("close", () => {
+                this.sendEvent(
+                    new OutputEvent(
+                        "closing serial port connection",
+                        "Serial Port"
+                    )
+                );
+            });
+
+            this.serialPort.open();
+        } else if (uart.socketPort !== undefined) {
+            this.socket = new Socket();
+            this.socket.setEncoding("utf-8");
+
+            let eolChar: string = uart.eolCharacter === "LF" ? "\n" : "\r\n";
+
+            let tcpUartData = "";
+            this.socket.on("data", (data: string) => {
+                for (const char of data) {
+                    if (char === eolChar) {
+                        this.sendEvent(
+                            new OutputEvent(
+                                tcpUartData,
+                                "Socket"
+                            )
+                        )
+                        tcpUartData = "";
+                    } else{
+                        tcpUartData += char;
+                    }
+                }
+            });
+            this.socket.on("close", () => {
+                this.sendEvent(
+                    new OutputEvent(
+                        tcpUartData,
+                        "Socket"
+                    )
+                );
+                this.sendEvent(
+                    new OutputEvent(
+                        "closing socket connection",
+                        "Socket"
+                    )
+                );
+            });
+
+            this.socket.connect(
+                // Putting a + (unary plus operator) infront of the string converts it to a number.
+                +uart.socketPort,
+                // Default to localhost if target.host is undefined.
+                host ?? "localhost",
+                () => {
+                    this.sendEvent(
+                        new OutputEvent(
+                            `listening on tcp port ${uart?.socketPort}`,
+                            "Socket"
+                        )
+                    );
+                }
+            );
+        }
+    }
+
     protected async startGDBAndAttachToTarget(
         response: DebugProtocol.AttachResponse | DebugProtocol.LaunchResponse,
         args: TargetAttachRequestArguments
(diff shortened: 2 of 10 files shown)


jonahgraham:
Thanks @AdhamRagabMCHP this looks like an interesting new feature. Thank you for taking the time to create a PR.

I have done a preliminary review of the code and here are my first thoughts:

- Please separate unrelated changes (e.g. cmd-param-changed and logging changes) into their own PRs. If they are independent we can get the reviewed and merged quickly and that reduces the diff size of this PR making it faster to review too. There is an additional benefit, once your first PR gets merged then GitHub considers you trusted and will run the GitHub actions checks automatically, so getting a trivial PR approved and merged is a great way to start
- Some of the new fields have been introduced in the global scope. Can they be moved into the debug session (e.g. SerialPort and socket)?
- Can the new parameters be grouped into their own interface? e.g. uartSerialPort to handshakingMethod be placed in a new structure and add a field called uart to the existing structure(s). This will allow users in the launch.json to logically group related entries together. 
- Can you explain the reason to use the logger to get messages from serial to the user? I think it may be better to create `OutputEvent`s with a custom destination (such as `"uart"` perhaps? or maybe `"stdout"`)
- We are going to need some tests for this code so that it can be exercised properly and to ensure it does not regress in the future.

Feature wise this provides one way comms from target to IDE. Is there a plan on how to support bidirectional communications? Bidirectional comms are not a pre-requisite, but I want to understand what the future path may look like.

AdhamRagabMCHP:
Thanks for your prompt review! This is really refreshing to witness.

I'll try to address your comments one-by-one here, and of course the changes requested will be incoming:
1. Sure, I can do that. I think logically, separating the changes in GDBDebugSession.ts and GDBTargetDebugSession.ts makes sense - they are unrelated changes. Changing the logging level is directly related to how I wanted to print the serial output (by using logger.log as opposed to logger.warn) coming from a user's application.
2. The socket has an event handler for closing, so that can be moved. However, we need to declare the SerialPort globally so that we can attempt to close the port on a disconnect request.
3. Sure, I can do something like that.
4. We're using the CDT-GDB-VSCode/Adapter to debug user applications on-board and/or through an emulation platform. In our setup, we have three debugging flows:
 - On-host debugging - We use a separate plugin for this (we were working on this flow before needed to support debugging on-board and through an emulation platform, where we came across this extension that was very helpful in remote debugging). In this case, output from the program prints directly to the debug console.
 - On-board debugging - In our case, a lot of our user applications feature some sort of output printed on the UART, and so as to prevent users from having to open up an external screen/gtkterm/Putty window to view UART output, we found it helpful to print it within their IDE. In this flow, we would be needing to read UART output across a serial line, necessitating the use of the NodeJS serialport library.
 - Debugging through an emulation platform - If you have no CPU for an application developed to work on-board, we also have the option to debug an application on an emulation platform that mimics the CPU you are compiling for. In this case, the flow would work just like the on-board debugging flow, except that we would try to read off of a TCP port that the emulation platform redirects UART output to.
 We want to be able to switch between all 3 flows without having to change how we debug/how we view debug results drastically. In our case, that means that the way output is printed should be the same across all three (hence printing output to the debug console). Doing so also preserves UART output in the generated log file for the debug session, which can be helpful for users debugging an application where they may expect specific output/output in a specific order.
 5. Sure, I'll think about what test cases I can add
 6. I don't think there is a concrete plan on our end to support bidirectional communication. We don't really have a use case right now where we'd need bidirectional communication, although it would be an interesting exercise to think about.

jonahgraham:
> Thanks @jonahgraham for your prompt review! This is really refreshing to witness.

:) I try to be quick when possible, but sometimes it can be very slow. Anytime feel free to ping me!

> 2. The socket has an event handler for closing, so that can be moved. However, we need to declare the SerialPort globally so that we can attempt to close the port on a disconnect request.

Not sure I undertand why here - if serialport is a field in the session you should be able to access in disconnect in the same way that we do `await this.gdb.sendGDBExit();` on the next line.

> 4. We're using the CDT-GDB-VSCode/Adapter to debug user applications on-board and/or through an emulation platform. In our setup, we have three debugging flows [...]

The output appearing the debug console isn't an issue for me. I'm just not terribly fond of user data going through the logger. Using an OutputEvent will still deliver the output to the debug console without all the extra indirection. However if you have a use case where making it go through the logger is important to you I won't block the PR on that issue.

> 5. Sure, I'll think about what test cases I can add

Please let me know if you need insight or help on writing test cases for this. The serial port one may be difficult, especially on Windows, but having at least something that exercises the code is useful.

AdhamRagabMCHP:
Thank you for your comments!
On 2: I see what you mean now - sure, I'll do something like that.
On 4: I'll try to make use of an OutputEvent and see how that changes the look of the output - perhaps it does what we want all the same without having to go through the logger.
On 5: Thanks so much! Yes, it will be difficult, so I may ping you from time to time :)

AdhamRagabMCHP:
I forgot to mention this in the commit message, but the latest commit also feature a small fix to properly disconnect remote targets when exiting a GDB session.

jonahgraham:
> I forgot to mention this in the commit message, but the latest commit also features a small fix to properly disconnect remote targets when exiting a GDB session.

Can you do that in a different PR please. The disconnect sequence has been a fairly big talking point, both on here and with some of the other adopters so I want to ensure that change is very visible.

AdhamRagabMCHP:
> > I forgot to mention this in the commit message, but the latest commit also features a small fix to properly disconnect remote targets when exiting a GDB session.
> 
> Can you do that in a different PR please. The disconnect sequence has been a fairly big talking point, both on here and with some of the other adopters so I want to ensure that change is very visible.

Sure that can work....I'll do that now

src/GDBTargetDebugSession.ts
```diff
@@ -343,6 +505,11 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             this.sendEvent(new InitializedEvent());
             this.sendResponse(response);
             this.isInitialized = true;
+            this.sendEvent(
```
jonahgraham, line 508: This looks like debugging the debug adapter message. I assume it isn't related to this change? The extra messages are ok in general if you think they help you and your users, but I find this one a little terse to be useful.

src/GDBTargetDebugSession.ts
```diff
@@ -331,6 +362,137 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+                            this.sendEvent(
+                                new OutputEvent(
+                                    `listening on serial port ${this.serialPort?.path}`,
+                                    "Serial Port"
+                                )
+                            );
+                        });
+
+                        const SerialUartParser = new ReadlineParser({
+                            delimiter: "\n",
```
jonahgraham, line 405: Should either of these items be parameters for the user? Often in UARTs `\r\n` is used a delimiter. Also, while UTF-8 is most often ok, i.e. don't know if we should support other types.

jonahgraham (changes requested):
This code looks good and nearly ready. Other than some small comments that I hope are clear enough and some test code I think this is ready and a great new feature for the adapter.

src/GDBTargetDebugSession.ts
```diff
@@ -18,6 +18,46 @@ import {
...
+
+/**
+ * This is a placeholder for the serial port. We assign the correct serial
+ * port path below in "startGDBAndAttachToTarget" if it is specified.
+ * Defining this variable here allows us to properly close the serial port
+ * on debug finish.
+ * Additionally, set autoOpen to false to prevent the serial port from being
+ * opened automatically.
+ */
+let SerialPort: serialport.SerialPort = new serialport.SerialPort({
```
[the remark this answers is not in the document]
AdhamRagabMCHP: Resolved.

src/GDBTargetDebugSession.ts
AdhamRagabMCHP: Removed.

src/GDBTargetDebugSession.ts
```diff
@@ -373,6 +540,22 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+                )
+                    this.serialPort.close();
+            } catch (err) {
+                this.sendErrorResponse(
+                    response,
+                    1,
+                    err instanceof Error ? err.message : String(err)
+                );
+            }
+            if (this.targetType === "remote") {
```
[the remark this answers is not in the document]
AdhamRagabMCHP: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/273 - Moved to a new PR.

src/GDBTargetDebugSession.ts
```diff
@@ -331,6 +362,137 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+                        this.sendErrorResponse(
+                            response,
+                            1,
+                            err instanceof Error ? err.message : String(err)
+                        );
+                    }
+                } else if (target.uart.socketPort !== undefined) {
+                    try {
+                        /**
+                         * This is a placeholder for the socket. We assign the correct socket
```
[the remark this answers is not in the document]
AdhamRagabMCHP: Addressed.

src/GDBTargetDebugSession.ts
```diff
@@ -331,6 +362,137 @@ export class GDBTargetDebugSession extends GDBDebugSession {
 
             await this.gdb.sendCommands(args.initCommands);
 
+            if (target.uart !== undefined) {
```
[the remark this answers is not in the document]
jonahgraham: > Although you the starting point for you for this method (startGDBAndAttachToTarget) was very large and complex, I think having the uart creation split out is fairly logical place to split it.

I just reread this comment and am sad by the mess it is. I hope you understood what I meant, but here is a clearer version:

Although this method (startGDBAndAttachToTarget) was already very large and complex when you started adding to it, I think we should try not to have too much additional complication. Having the uart creation split out at this time is a logical place to do it.

src/GDBTargetDebugSession.ts
jonahgraham: The biggest issue I tend to see with (my client Renesas) is that sometimes reading ShiftJIS (Japanese default) encoded data as UTF-8 causes corruption. If we are only interested in ASCII then UTF-8 should indeed be ok.

src/GDBTargetDebugSession.ts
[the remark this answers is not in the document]
AdhamRagabMCHP: No worries! I understood the initial comment :)

jonahgraham:
> On 5: Thanks so much! Yes, it will be difficult, so I may ping you from time to time :)

Any luck with how to test the code? I want to make sure that we don't spend an unreasonable time on this part of it.

AdhamRagabMCHP:
> > On 5: Thanks so much! Yes, it will be difficult, so I may ping you from time to time :)
> 
> Any luck with how to test the code? I want to make sure that we don't spend an unreasonable time on this part of it.

Hey! I'm still trying to write the test-code - we had a long weekend here, so I'm just getting back to it now. I'll keep you posted on the test code - I'll start off writing a testcase for the socket case.

AdhamRagabMCHP:
> > On 5: Thanks so much! Yes, it will be difficult, so I may ping you from time to time :)
> 
> Any luck with how to test the code? I want to make sure that we don't spend an unreasonable time on this part of it.

On this - what's your policy of including binaries as part of the test suite? I think the easiest option for me would be to recreate a simple emulator setup and go from there, but perhaps you'd like to avoid including the emulator binary as part of the testsuite, in which case I would have to keep exploring different options.

jonahgraham:
We can't have binaries checked in, but we can build additional binaries as part of the build.

AdhamRagabMCHP:
> We can't have binaries checked in, but we can build additional binaries as part of the build.

Understandable. I currently have a working testcase (I think) - it requires netcat, which is pre-installed by default on most, if not all, Linux distros but will need to be downloaded manually for Windows - what's the policy in that case? I believe there's instructions to build it from source, but it would still require something like Cygwin.

AdhamRagabMCHP:
@jonahgraham Just wanted to double-check that the way I wrote the test is valid - seems that it passes on both Windows and Linux.

src/integration-tests/debugClient.ts
```diff
@@ -371,6 +371,35 @@ export class CdtDebugClient extends DebugClient {
...
+                .then((_event) => {
+                    return this.waitForOutputEvent(
+                        category,
+                        output
+                    );
+                })
+                .then((response) => {
+                    expect(response.body.category).to.equal(category);
+                    expect(response.body.category).to.equal(output);
+                }),
```
jonahgraham, line 396: You don't need to `then` the `waitForOutputEvent` because `waitForOutputEvent` will already fail if an output event without that category/output comes through.
```suggestion
                }),
```

jonahgraham (changes requested):
This looks very promising and in the right direction. I have offered some feedback and I hope it is useful. 

I do try to maintain a high bar for making sure new code is at least somewhat tested. However as you are the one doing this work I would appreciate your feedback if you find that the overhead of writing the test for this code is disproportionate to the perceived value.

jonahgraham:
> Understandable. I currently have a working testcase (I think) - it requires netcat, which is pre-installed by default on most, if not all, Linux distros but will need to be downloaded manually for Windows - what's the policy in that case? I believe there's instructions to build it from source, but it would still require something like Cygwin.

I think if the most straightforward way to write the tests is to use netcat or any tool that is Linux specific, I would simply mark the tests as skip on Windows. Getting the code coverage on Linux would be great and will help ensure that a future code change doesn't break your new feature.

If we later find a regression or error that is Windows specific we may then want to write a test that covers that case.

I hope that helps!

src/integration-tests/launchRemote.spec.ts
```diff
@@ -43,4 +43,31 @@ describe('launch remote', function () {
             }
         );
     });
+
+    it('can print a message to the debug console sent from a socket server', async function () {
+        dc.getSocketOutput(
```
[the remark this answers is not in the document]
AdhamRagabMCHP: > This code is missing an `await` here, so the tests aren't actually running (they may be running, but they end up running in the background and the results aren't captured).
> 
> Below is some information you may find useful to track down test results:
> 
> I can see this is in the GitHub actions output [(link to the part quoted below)](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/5479172154/jobs/9980650432?pr=271#step:8:99):
> 
> ```
>   launch remote
>       ✔ /defaults/can launch remote and hit a breakpoint: 1125ms
>       ✔ /defaults/can print a message to the debug console sent from a socket server: 0ms
> ```
> 
> A `0ms` runtime is the indication of a problem as it means that the test hasn't actually run long enough to start/stop a gdb + gdbserver pair.
> 
> The other thing you can do is look at the logs (as too briefly mentioned [here](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/README.md#testing-on-github-actions)). For example [this is the link](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/suites/14115793439/artifacts/789958902) to the Ubuntu logs for the run in question. In it the tests are organized by folder, so the logs for this test are in "launch remote" - each test is run multiple times, which is the next level in the hierarchy, followed by logs for the individual tests.
> 
> For this test I can see `/launch remote/defaults/can print a message to the debug console sent from a socket server.log` for the default test arguments. In that test you can see the `disconnect` is being sent by the client (aka the test class) in the same millisecond as the log file is created.
> 
> The Windows version of that file doesn't exist in the zip file. My best guess is that is because the adapter doesn't even have time to start before the test is over (process creation time on Windows being much slower than Linux generally). The log files are written by the adapter, so if the adapter doesn't even start the log file doesn't exist.

That's what I suspected as well - thank you for this feedback! It seemed odd that it worked on Windows, so just wanted to confirm that it hadn't actually ran.

src/integration-tests/launchRemote.spec.ts
```diff
@@ -43,4 +43,31 @@ describe('launch remote', function () {
...
+                    "host": "localhost",
+                    "port": "9899",
+                    "server": "gdbserver",
+                    "serverParameters": [
+                        "--once",
+                        "localhost:9899",
+                        emptyProgram
+                    ],
+                    "uart": {
+                        "socketPort": "3456",
```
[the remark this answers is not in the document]
AdhamRagabMCHP: I missed this - thank you for the feedback. I was playing around with these ports while developing. I'll fix this in my next commit.

src/integration-tests/debugClient.ts
AdhamRagabMCHP: Thank you! I'll apply this change.

AdhamRagabMCHP:
> > Understandable. I currently have a working testcase (I think) - it requires netcat, which is pre-installed by default on most, if not all, Linux distros but will need to be downloaded manually for Windows - what's the policy in that case? I believe there's instructions to build it from source, but it would still require something like Cygwin.
> 
> I think if the most straightforward way to write the tests is to use netcat or any tool that is Linux specific, I would simply mark the tests as skip on Windows. Getting the code coverage on Linux would be great and will help ensure that a future code change doesn't break your new feature.
> 
> If we later find a regression or error that is Windows specific we may then want to write a test that covers that case.
> 
> I hope that helps!

I think NodeJS hopefully is a way forward - other than the serial port (which we'll look into for now), I think there's probably some promise there. I want to avoid having to download any tools, so I'll keep looking.

Thank you so much for all your helpful feedback on this - I've learnt a lot!

src/integration-tests/launchRemote.spec.ts
```diff
@@ -43,4 +43,31 @@ describe('launch remote', function () {
             }
         );
     });
+
+    it('can print a message to the debug console sent from a socket server', async function () {
+        dc.getSocketOutput(
+            fillDefaults(this.test, {
+                program: emptyProgram,
+                openGdbConsole: false,
+                initCommands: ["shell echo \"Hello World!\" | nc -l 3456 &"],
```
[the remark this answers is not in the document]
AdhamRagabMCHP: I was playing around with socat yesteday actually - it seems promising on Linux like you said, and just like you mentioned - the roadblock is a cross-platform solution. I've found that simulating a socket server from the terminal was probably my best solution, so the NodeJS way should hopefully work. I'll push a commit for this.

AdhamRagabMCHP:
@jonahgraham is there a way I can get access to the logfile from the most recent run, or does it get cleaned up after the tests run? I don' want to keep force pushing blindly - I can't run these tests locally so I'm relying on the PR run to test them.

jonahgraham:
> @jonahgraham is there a way I can get access to the logfile from the most recent run, or does it get cleaned up after the tests run?

Yes - they should all be there: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions - the logs are preserved indefinitely.

> I don't want to keep force pushing blindly - I can't run these tests locally so I'm relying on the PR run to test them.

The other thing you can do is enable github actions on your fork and then all pushes you do will run the actions on your GHA: https://github.com/AdhamRagabMCHP/cdt-gdb-adapter/actions

jonahgraham:
Specifically the most recent run seems to be https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/5489485644

AdhamRagabMCHP:
> Specifically the most recent run seems to be https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/5489485644

I was referring more to the adapter log files - where we can see the verbose adapter logs. I haven't been able to find those logs at all.

jonahgraham:
Click on the run link, then Summary in the top left:

image

Then scroll down a little and download the zipped log files per platform:

image

AdhamRagabMCHP:
> Click on the run link, then Summary in the top left:
> 
> image
> 
> Then scroll down a little and download the zipped log files per platform:
> 
> image

Thank you so much!

AdhamRagabMCHP:
The socket test passes on both Ubuntu and Windows now - just formatting the source code now.

I'll look to add the test for the serial port.

AdhamRagabMCHP:
I added the test for the serial line output - I skip the test on Windows, and it passes in Linux.

AdhamRagabMCHP:
> I added the test for the serial line output - I skip the test on Windows, and it passes in Linux.

Actually - I noticed that sometimes, it fails to register as a success for this last one - from seeing the logs, it seems that the intended behaviour occurs and in fact, some of the runs show the test as having passed:

image


Is there anything I'm missing?

src/integration-tests/debugClient.ts
```diff
@@ -371,6 +371,24 @@ export class CdtDebugClient extends DebugClient {
...
+        launchArgs: any,
+        category: string,
+        output: string
+    ): Promise<any> {
+        return Promise.all([
+            this.waitForEvent('initialized'),
+            this.launch(launchArgs).then((_event) => {
+                return this.waitForOutputEvent(category, output);
+            }),
+        ]);
```
jonahgraham, line 390: > Is there anything I'm missing?

The reason the tests are unstable is because you only start waiting for the output after launch returns - but if the output comes too quickly after the launch before you have a chance to start `waitForOutputEvent` the output you are looking for is missed. Therefore you have a race condition here.

Here is a suggested change. 

```suggestion
        return Promise.all([
            this.waitForEvent('initialized'),
            this.launch(launchArgs),
            this.waitForOutputEvent(category, output),
        ]);
```

package.json
```diff
@@ -60,7 +60,9 @@
   "dependencies": {
     "@vscode/debugadapter": "^1.59.0",
     "@vscode/debugprotocol": "^1.59.0",
-    "node-addon-api": "^4.3.0"
+    "node-addon-api": "^4.3.0",
+    "serialport": "11.0.0",
+    "@types/serialport": "8.0.2"
```
jonahgraham, line 65: I think `@types/serialport` should be a dev dependency and both of these dependency changes should also appear in the `yarn.lock` file

src/integration-tests/launchRemote.spec.ts
```diff
@@ -43,4 +46,71 @@ describe('launch remote', function () {
             }
         );
     });
+
+    it('can print a message to the debug console sent from a socket server', async function () {
+        const socketServer = cp.spawn(
+            'node',
+            [`${path.join(testProgramsDir, 'socketServer.js')}`],
```
jonahgraham, line 53: Can you run this code in the same process (and write it in typescript)? If it needs to be in another process, you need to ensure the process is terminated as part of the test.

See also #277 - we don't want to make this problem worse :)

src/integration-tests/launchRemote.spec.ts
```diff
@@ -43,4 +46,71 @@ describe('launch remote', function () {
...
+            'Hello World!'
+        );
+    });
+
+    it('can print a message to the debug console sent from across a serial line', async function () {
+        // Skip this test on Windows - socat utility only available on Linux.
+        if (os.platform() === 'win32') this.skip();
+
+        // Start a virtual serial line. Use /tmp/ttyV0 and /tmp/ttyV1 to refer to the two ends.
+        cp.spawn('socat', [
```
jonahgraham, line 91: You need to terminate `socat` - see [this](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/5498351729/jobs/10043412201#step:23:1):

image

jonahgraham (changes requested):
This is very close to done. Looks very good to me. Please see the individual comments for what is left.

src/integration-tests/launchRemote.spec.ts
AdhamRagabMCHP: Resolved - I'll ensure the process is terminated as part of the test.

src/integration-tests/launchRemote.spec.ts
AdhamRagabMCHP: Resolved - I'll ensure the process is terminated as part of the test.

src/integration-tests/launchRemote.spec.ts
AdhamRagabMCHP: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/5510587354/jobs/10045008025#step:23:1 - All processes are terminated.

package.json
AdhamRagabMCHP: Resolved.

jonahgraham:
@AdhamRagabMCHP it looks like you have resolved all my outstanding review comments. Please let me know if you have anything else pending or if you are ready for me to merge.

AdhamRagab25:
I think it's good to merge - I don't think I'll be adding anything else to the PR.

jonahgraham (approved):

jonahgraham:
Thank you @AdhamRagab25 for this new functionality!

AdhamRagabMCHP:
@jonahgraham Thank you for all your help and feedback! It was incredibly helpful!

PR #272, opened 2023-06-29
Resolve Unhandled Notify Warning

AdhamRagabMCHP:
This change resolves an "unhandled notify" warning coming when the cmd-param-changed notification is received.

To recreate this warning:
- In your initCommands, you can put something like "set mem inaccessible-by-default off" - this warning should appear in the debug log.

To check that this change resolves the warning:
- Re-run a debug session with the same initCommand above after applying the change.


jonahgraham (approved):

jonahgraham:
Thank you @AdhamRagabMCHP and congratulations on your first merged contribution.

I build and publish this to npm when requested, so if you want the version with this change published let me know, or else we can publish once #271 is resolved.

PR #273, opened 2023-06-30
Disconnect before exit for remote targets

AdhamRagabMCHP:
This changes fixes an issue whereby when debugging on a remote target, the exit sequence is not entirely correct. We should send a "disconnect" GDB command before "exit". This prevents a messages such as "Remote doesn't know how to detach" and "Cannot execute this command while the target is running" from appearing, as they can be somewhat misleading.

To recreate this:
- Exit a remote debug session before this change.

To check that this works:
- Exit a remote debug session after this change. In place of the messages described above, you should see "disconnect" in the debug console if the logging level is at warn ("verbose" = false in the launch.json)/

### DIFF
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -80,6 +80,13 @@ export class GDBTargetDebugSession extends GDBDebugSession {
     protected gdbserver?: ChildProcess;
     protected killGdbServer = true;
 
+    /**
+     * Define the target type here such that we can run the "disconnect"
+     * command when servicing the disconnect request if the target type
+     * is remote.
+     */
+    protected targetType?: string;
+
     protected async attachOrLaunchRequest(
         response: DebugProtocol.Response,
         request: 'launch' | 'attach',
@@ -293,7 +300,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             }
 
             if (target.connectCommands === undefined) {
-                const targetType =
+                this.targetType =
                     target.type !== undefined ? target.type : 'remote';
                 let defaultTarget: string[];
                 if (target.port !== undefined) {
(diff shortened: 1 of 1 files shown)


jonahgraham (approved):
I am fine with this change. The disconnect sequence of different hardware gets pretty tricky and the number of ways to do it (disconnect, detach, do nothing) means there are potentially as many opinions on the right way to do it.

Therefore before I merge this change as is I will let the change sit for a couple of days in case other people who may be affected feel strongly enough to comment.

cc: @thamho0902 @thegecko @marcdumais-work @sdirix  - do any of you have a strong opinion on this? I have cc'ed each of you because AFAIK you have at some point been involved in the behaviour around terminating debug sessions and I don't want this change to come as too much of a surprise to you.

jonahgraham:
@AdhamRagabMCHP We'll give it a few days to give others a chance to comment. Feel free to ping me next week to merge it if there has been no further comment.

AdhamRagabMCHP:
Sounds good!

thegecko:
lgtm

AdhamRagabMCHP:
> @AdhamRagabMCHP We'll give it a few days to give others a chance to comment. Feel free to ping me next week to merge it if there has been no further comment.

Hello Jonah! I'm not sure if this PR is good to merge now - we can wait an additional day or so, but just curious if now's a good time to merge.

sdirix:
I did not test the PR but from my understanding it should work fine for the Blueprint use cases. Therefore I'm fine with merging :+1:

jonahgraham:
Thanks @AdhamRagabMCHP for the fix. I can do a new version now, or wait until #271 is done. Let me know if you want it now.

AdhamRagabMCHP:
> Thanks @AdhamRagabMCHP for the fix. I can do a new version now, or wait until #271 is done. Let me know if you want it now.

I think we'll wait until #271 is done - it makes sense to have all the changes in first.

PR #276, opened 2023-07-05
Add support for escaped unicode characters

### LINKED ISSUE #275: Dose not support chinese character?
Hi,
   I have a simple C++ file, and I want to debug it in vscode, the path of the C++ file has some Chinese characters, and when I debug I get an error
image
my project folder is as follows
image
It seems that cdt-gdb-vscode does not recognize ‘测试’, and give a series of number which is not in my project
when I change ‘测试’ to English character it is ok… . how to fix it?

jonahgraham:
GDB is escaping unicode characters leading to the messages coming back from GDB not being understood correctly.

For example, hitting a breakpoint on issue-275-测试.c comes back from GDB as `file="issue-275-\346\265\213\350\257\225.c"` but by the time we send it to DAP client we are sending
`"name":"issue-275-346265213350257225.c"`

This patchset supports properly escaping these strings.

Fixes #275

TODO:

- [x] This is just a test so far that demonstrates the problem in #275, so the fix is needed
- [x] On Windows this doesn't build the example programs yet which have the unicode characters. That isn't an adapter problem, but a test problem.
- [x] Tom provided diagnosis of https://sourceware.org/bugzilla/show_bug.cgi?id=30618 and it looks like I can at least workaround the issue in the test

### DIFF
--- a/package.json
+++ b/package.json
@@ -61,7 +61,8 @@
     "@vscode/debugadapter": "^1.59.0",
     "@vscode/debugprotocol": "^1.59.0",
     "node-addon-api": "^4.3.0",
-    "serialport": "11.0.0"
+    "serialport": "11.0.0",
+    "utf8": "^3.0.0"
   },
   "devDependencies": {
     "@types/chai": "^4.3.4",
@@ -70,6 +71,7 @@
     "@types/node": "^14.18.17",
     "@types/serialport": "8.0.2",
     "@types/tmp": "^0.2.3",
+    "@types/utf8": "^3.0.1",
     "@typescript-eslint/eslint-plugin": "^5.54.0",
     "@typescript-eslint/parser": "^5.54.0",
     "@vscode/debugadapter-testsupport": "^1.59.0",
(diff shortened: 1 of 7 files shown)


jonahgraham:
I have added a fix which is to decode all cstrings as UTF-8 and handling of the escaped octal numbers. It doesn't work on Windows yet.

I am not sure what else may be needed here, nor if there is any negative impacts of this change. Interestingly I see other encoding errors reported for cpp tools and cortex debug as unfixable in the adapters. See https://github.com/microsoft/vscode-cpptools/issues/1998, https://github.com/Marus/cortex-debug/issues/207 and https://github.com/Marus/cortex-debug/issues/297. Therefore while I think we fix the specific issue reported in #275 I am convinced other issues may still prevent proper support.

tromey:
MI does a weird thing where all strings are just ASCII, so any non-ASCII characters are emitted byte-wise using C-like escapes.  The MI client has to know which encoding to use.  This is awful of course, but fixing it would require bumping the MI version and there hasn't been any real pressure to deal with it.  Some other escape sequences are used here (e.g., "\n").  I am not sure if this is documented.

jonahgraham:
Thank for having a look. We're handling n, t, r and now octal 0-7. I'll add more as I have specific cases.

tromey:
> Thank for having a look. We're handling n, t, r and now octal 0-7. I'll add more as I have specific cases.

According to the source I think `\b`, `\f`, `\e`, and `\a` can be emitted as well.

PR #279, opened 2023-07-10
Make LF the default for UART eol characters

AdhamRagabMCHP:
This fixes a bug introduced by #271 where the EOL character for UART output should have been LF, but was instead CRLF. See https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/91 for comments on this.


jonahgraham (approved):

AdhamRagabMCHP:
Just updated it to resolve the Prettier formatting error.

PR #280, opened 2023-07-11
Add newline characters to every OutputEvent emitted regarding UART output

AdhamRagabMCHP:
This PR adds newlines to every OutputEvent emitted for UART output to the debug console. It matches the OS it is on + the eolCharacter specified in the launch.json. See https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/91 for an initial discussion on why this came to be.

### DIFF
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -16,6 +16,7 @@ import {
     OutputEvent,
 } from '@vscode/debugadapter';
 import * as mi from './mi';
+import * as os from 'os';
 import { DebugProtocol } from '@vscode/debugprotocol';
 import { spawn, ChildProcess } from 'child_process';
 import { SerialPort, ReadlineParser } from 'serialport';
@@ -325,7 +326,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             this.serialPort.on('open', () => {
                 this.sendEvent(
                     new OutputEvent(
-                        `listening on serial port ${this.serialPort?.path}`,
+                        `listening on serial port ${this.serialPort?.path}${
+                            os.platform() === 'win32' ? '\r\n' : '\n'
+                        }`,
                         'Serial Port'
                     )
                 );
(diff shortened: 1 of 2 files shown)


src/integration-tests/launchRemote.spec.ts
```diff
@@ -79,7 +79,7 @@ describe('launch remote', function () {
                 } as TargetLaunchArguments,
             } as TargetLaunchRequestArguments),
             'Socket',
-            'Hello World!'
+            'Hello World!\n'
```
jonahgraham, line 82: This too should be `os.EOL` if you agree about the above changes.

jonahgraham (changes requested):

src/GDBTargetDebugSession.ts
```diff
@@ -325,7 +326,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             this.serialPort.on('open', () => {
                 this.sendEvent(
                     new OutputEvent(
-                        `listening on serial port ${this.serialPort?.path}`,
+                        `listening on serial port ${this.serialPort?.path}${
+                            os.platform() === 'win32' ? '\r\n' : '\n'
```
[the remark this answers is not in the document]
AdhamRagabMCHP: I used this based on a different bit of code I saw, but I'll push this sure. Resolved.

src/integration-tests/launchRemote.spec.ts
AdhamRagabMCHP: I will change that indeed yes - will be part of the next push.

AdhamRagabMCHP:
Seems that using os.EOL causes issues on Windows...let me see why that is

AdhamRagabMCHP:
To client: {"seq":0,"type":"event","event":"output","body":{"category":"Socket","output":"Hello World!\r\r\n"}}

It has two carriage return characters. Let me see how to ensure how this doesn't happen.

AdhamRagabMCHP:
Ah - It's because of the test. Let me fix it.

AdhamRagabMCHP:
Test should be fixed on Windows now.

jonahgraham (approved):
LGTM - once tests pass I will run it up in VSCode as part of the extension to make sure this all works. I can then do a new tagged version.

AdhamRagabMCHP:
Tests now have passed.

PR #282, opened 2023-07-11
Add missing entries to yarn.lock

jonahgraham:
This is a fixup for e39e7bc4f44aee4ddbb0c6c21b71ca74067dc33e


jonahgraham:
I missed that these entries should have been added in #271. These missing entries were simply created by running `yarn` and then reviewed to make sure they should be present.

@AdhamRagabMCHP - FYI.

AdhamRagabMCHP:
I ignored these entries since I thought they were duplicates - my bad for not adding them.

PR #283, opened 2023-07-11
Verify yarn.lock is up to date

jonahgraham:


jonahgraham:
This is to prevent the error that required #282 from being required in the future.

jonahgraham:
@AdhamRagabMCHP - this new check will make sure I don't accidentally merge without an update yarn.lock in the future.

AdhamRagabMCHP:
Sounds good - sorry about that! Some of the dependencies just looked like duplicates, but I'll be careful in the future!

jonahgraham:
No worries - the dependency structures of yarn are very complex so it was a reasonable assumption you made.

PR #288, opened 2023-08-30
Remove unnecessary `process` imports

colin-grant-work:
When experimenting with bundling the CDT-GDB VSCode extension using ESBuild, I ran into a problem similar to [this one](https://github.com/evanw/esbuild/issues/587) that turned out to be due to the style of `process` imports used here. Since `process` is a Node global, the imports should be unnecessary, so this PR removes them.

### DIFF
--- a/src/debugAdapter.ts
+++ b/src/debugAdapter.ts
@@ -9,7 +9,6 @@
  *
  * SPDX-License-Identifier: EPL-2.0
  *********************************************************************/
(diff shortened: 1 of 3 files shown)


jonahgraham:
@colin-grant-work - feel free to merge at your convenience, or ask me and I can merge. I assume it would be best to publish a new version to npm too for consumption in cdt-gdb-vscode?

PR #296, opened 2023-09-19
Improved Error Handling for sendCommand

WyoTwT:
Improve debugging of sendCommand by capturing the stack where the request originated instead of the stream read/parsing of message.


jonahgraham:
Thanks @WyoTwT!

PR #297, opened 2023-09-21
Improve Call Stack Window Display

WyoTwT:
Use the function name if available before falling back to fullname. This improves the use of space in the call stack in most cases since the function name is known.

This change:
Ericsson_improvement_stackTraceRequest

Upstream with long form:
Upstream_stackTraceRequest

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -939,7 +939,10 @@ export class GDBDebugSession extends LoggingDebugSession {
                     threadId: args.threadId,
                     frameId: parseInt(frame.level, 10),
                 });
-                const name = frame.func || frame.fullname || '';
+                const name =
+                    frame.func && frame.func != '??'
+                        ? frame.func
+                        : frame.fullname || frame.addr || '';
                 const sf = new StackFrame(
                     frameHandle,
                     name,


jonahgraham:
This change looks ok to me - but for reasons I can't understand yet there are test failures with it. I have restarted tests for a second time just now as it may just be weird Windows issues as the tests work fine on Linux.

jonahgraham:
Hmm - tests keep failing and I haven't seen this problem before. I created #298 to test the test runs without this code change to find out what is going on.

jonahgraham:
I found what changed - GitHub updated chocolatey's preinstall mingw from version 11 to 12 in the [last few days](https://github.com/actions/runner-images/commit/fa9f2bc4a36943bb45e61cd1544ac280d95df2f5) and that includes gdb going from 10.2 -> 11.2. So it looks like there is some instability in the new version causing us grief. 

That isn't related to this change so I won't hold this change up for that problem

jonahgraham (commented):
> Use the function name if available before falling back to fullname. 

I don't understand how the change in the code does this - the old code seemed to prefer frame.func if it was available:

```typescript
name = frame.func || frame.fullname || ''
```

I do like the change that you are excluding `??` and falling back to `addr` before using blank string, but I don't understand how this change leads to the difference in screenshots provided.

WyoTwT:
Hmm. Yes, I agree with your analysis and I can't reproduce the full address problem that I used in the screenshot. Now I'm uncertain what the state was for the full address issue - it probably was a different section of code that changed. 

Thanks, I'll close this PR.

PR #298, opened 2023-09-22
Test the reliability of GHA

jonahgraham:
Runs of #297 keep failing but for reasons that don't *seem* to be part of the change, so this PR is to allow me to test GHA runs.


jonahgraham:
See https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/297#issuecomment-1731786254 and #299

PR #300, opened 2023-09-22
Accumulate gdbserver output before checking for target port

### LINKED ISSUE #299: GDB version update on GitHub Actions on Windows causes test failures
GitHub updated chocolatey's preinstall mingw from version 11 to 12 in the [last few days](https://github.com/actions/runner-images/commit/fa9f2bc4a36943bb45e61cd1544ac280d95df2f5) and that includes gdb going from 10.2 -> 11.2. So it looks like there is some instability in the new version causing us grief.

jonahgraham:
Because the needed output from gdbserver may not arrive as a single 'data' event on stdout we need to accumulate the output before passing it to the regex checker.

Fixes #299

### DIFF
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -219,6 +219,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         await new Promise<void>((resolve, reject) => {
             this.gdbserver = spawn(serverExe, serverParams, { cwd: serverCwd });
             let gdbserverStartupResolved = false;
+            let accumulatedStdout = '';
             let accumulatedStderr = '';
             let checkTargetPort = (_data: any) => {
                 // do nothing by default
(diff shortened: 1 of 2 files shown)


src/GDBTargetDebugSession.ts
```diff
@@ -235,14 +236,22 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
+                    const regexStr = target.serverPortRegExp
+                        ? target.serverPortRegExp
+                        : 'Listening on port ([0-9]+)';
+                    const regex = new RegExp(regexStr);
+                    this.sendEvent(
+                        new OutputEvent(
+                            `checking '${data}' for /${regexStr}/`,
+                            'server'
+                        )
                     );
```
jonahgraham, line 248: I had a failure with the earlier version of this change, so I added this extra output event to understand what is happening and now the test passes - still a bit of experimentation to make sure that this accumulation is implemented correctly.

jonahgraham:
This change updates the regex from `Listening on port ([0-9]+)` to `Listening on port ([0-9]+)\r?\n`, all the docs for default value of `serverPortRegExp` say "defaults to matching a string like 'Listening on port 41551'":

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/3e91a6732cb2f6740edf15889e25dd320602c590/src/GDBTargetDebugSession.ts#L67

which means I don't think it should be a problem to change the default implementation here.

PR #303, opened 2023-10-04
Injecting Environment Variables to GDB and GDBSERVER processes

asimgunes:
Hi, 

I like to propose a new feature for cdt-gdb-adapter which enables injecting environment variables to gdb and gdbserver processes.

The new variables are located as `environment` and `target.environment`. The root `environment` injects environment variables to `gdb` process, where `target.environment` injects environment variables to gdbserver process.

Similar to `cwd` and `target.cwd` behaviour, gdbserver process also includes `environment` definition. (Which means gdbserver is injecting both environment and target.environment).

In this implementation, we also support construction of new values with using the old values. This is a common scenario for PATH environment variable in most cases. The following configuration will append a new path to the PATH variable:

PATH: '%PATH%;C:\some\new\path'

or

PATH: '$PATH:/some/new/path'

New value construction is not limited to the PATH variable, the logic could be used in any
variable and the following formats are supported:

%VAR_NAME% format:
TEST_VAR: "%TEST_VAR%;Some other text"

$VAR_NAME format:
TEST_VAR: "$TEST_VAR;Some other text"

${env.VAR_NAME} format:
TEST_VAR: "${env.TEST_VAR};Some other text"

I believe this could be a useful scenario in some cases where developers need to inject some environment variables seamlessly and improve the user experience.

To sum up, I would be happy if you can review this request, and happy to hear any feedback.

Kind regards.
Asim Gunes


asimgunes:
Some additional update notes:

1. I observe that there is a problem in the Windows integration tests. As far as I see, new tests are working fine but one old test is failing in the report.

2. I implemented new tests for utility functions, which checks the inject logic and text building formats, however, I couldn't find the proper way to test if the `environment` variables is injected/passed in to the spawn function. (Since CdtDebugClient is already creating a new process while running the tests, libraries like sinon could not create stub functions to check the parameters.), I do my tests manually and locally, out of scope of this update. I am open for suggestions on this point.

jonahgraham:
For windows tests are broken, see #299

For tests I think we can use getenv C function in a new test class to see what environment ends up in the inferior. See https://www.tutorialspoint.com/c_standard_library/c_function_getenv.htm
Eg you can call setenv on a few env vars and assign them to c variables, insert a breakpoint after those calls and ensure the C vars have expected values. 

The other thing you can do is use `cdt-gdb-tests/executeCommand` to send arbitrary gdb commands to check the effect has worked as expected. 

I'll do a review of the code itself soon.

src/util.ts
```diff
@@ -105,3 +113,92 @@ export function compareVersions(v1: string, v2: string): number {
...
+ * 		Regex rule definitions to capture the key information from the string.
+ * @return
+ * 		String build with provided data collection and key rules.
+ */
+export function buildString(
+    str: string,
+    data: any,
+    ...keyRegexs: RegExp[]
+): string {
+    const _resolveFromSourceHandler =
```
jonahgraham, line 134: I expected `_` to be unreferenced variables:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/33b2f35173f67563ebb1d42a6cd4aaf7484cb90c/.eslintrc.js#L16-L22

Is this a different style?

src/integration-tests/util.spec.ts
```diff
@@ -58,3 +63,231 @@ describe('util', async () => {
...
+            'TEST VALUE TEST1_VALUE, also $VAR2, and also ${env.VAR3}'
+        );
+    });
+});
+
+describe('createEnvValues', () => {
+    const initialENV = {
+        VAR1: 'TEST1',
+        VAR2: 'TEST2',
+        var3: 'TEST3',
```
jonahgraham, line 171: I see use of lowercase here, but afaict there is no test for either of these cases:

- Linux/osx: case sensitive variables, i.e. `var` != `VAR`
- Windows: case insensitive variables, i.e. `var` == `VAR`

In particular, `PATH` on Windows is typically written `Path`, but can be any capitalization of `path`.

So a `valuesToInject` of `{Path: 'C:\\mytools\\bin;%Path%'}` should update if the `initialENV` has `PATH:` or `Path:` properly.

src/integration-tests/util.spec.ts
```diff
@@ -58,3 +63,231 @@ describe('util', async () => {
...
+        const valuesToInject = {
+            VAR5: 'TEST5',
+        };
+        const result = createEnvValues(initialENV, valuesToInject);
+
+        expect(result).to.deep.equals({ ...initialENV, ...valuesToInject });
+    });
+    it('should injection formats', () => {
+        const valuesToInject = {
+            VARFORMATTEST1: '$VAR1',
```
jonahgraham, line 200: I think there should be a test where there is more than one environment variable in the `valuesToInject`, e.g: `valuesToInject = { NEWTEST1: '$VAR1$VAR1', NEWTEST2: '$VAR1VAR2' }`

jonahgraham (changes requested):
I tried adding `${env:PATH}` to my launch.json and it was expanded as I expected with no special handling. Can you look again at that? I would prefer just to have `${env:XXXX}` from VSCode and do no special handling.

jonahgraham:
> I tried adding `${env:PATH}` to my launch.json and it was expanded as I expected with no special handling.

I experimented with case sensitivity and multiple vars in a single expansion and it all worked as I would expect according to https://code.visualstudio.com/docs/editor/variables-reference#_environment-variables

src/util.ts
asimgunes: In this implementation, setting the variable to null is unsetting the environment variable. I add it to the tests, but I see that I missed in documentation. 

My personal view is since `_` is also a string I will vote for keep using `null` to unset the variables and update the documentation, how this sounds to you? 

By the way, `buildString` is only used for merging the key/values into a given string (some kind of string interpolation). The variables handled in `createEnvValues` function.

asimgunes (commented):
Thanks for the feedback, let me check on them.

src/util.ts
jonahgraham: I think we are talking cross purposes. 

My comment was the minor issue of `_resolveFromSourceHandler` starting with a `_` because of the quoted rules in eslint

jonahgraham:
@asimgunes please let me know when you are ready for me to re-review this, with a comment on pressing the "re-request review" button.

asimgunes:
Hi @jonahgraham,

I think this pull request is ready to re-check right now. 

To summarize, following changes performed:

1. Custom variable injection is removed so we leave the VSCode to handle ${env:VARNAME} format and we do not perform any string editing operation.
2. We cover the Windows environment variables handling scenario with the following rules:
2.1. We looked for case-insensitive environment variables key, if we found any, we update the value of this variable while keeping the existing casing.
2.2. If we do not found the variable than we insert the variable with the provided definition.
2.3. We ignore this check/control if operating system is not Windows.
3. Update the `cdt-gdb-tests/executeCommand` command and inject console output as well as the command result. I need to update this implementation due to executing command and getting results. (The previous implementation was neither getting the result nor getting the console output. I added both of them.)
4. Tests are updated to check environment and target.environment settings are working properly. We get the gdb process environment variables via gdb commands and inferior variables from the application itself. 

Note: Both commands tried: `show environment <VARNAME>` and `call (char *) getenv("<VARNAME>")` to get the variables and both of them return the result in console output. Thats the reason for `cdt-gdb-tests/executeCommand` command update.

src/integration-tests/launchWithEnvironment.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+        test?: Mocha.Runnable | undefined,
+        environment?: Record<string, string | null> | undefined,
+        targetEnvironment?: Record<string, string | null> | undefined
+    ) => {
+        dc = await standardBeforeEach(adapter || debugAdapter);
+        await dc.launchRequest(
+            fillDefaults(test, {
+                program: path.join(
+                    testProgramsDir,
+                    platform() === 'win32' ? 'vars_env.exe' : 'vars_env'
```
jonahgraham, line 42: It doesn't matter much, but do you know why this code needs the `.exe` where the rest of the tests don't?

src/integration-tests/launchWithEnvironment.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+    };
+
+    afterEach(async function () {
+        // dc could be undefined if test is skipped.
+        await dc?.stop();
+        dc = undefined;
+    });
+
+    it('sets environment variables passed to the process', async function () {
+        if (hardwareBreakpoint) {
```
jonahgraham, line 151: Why skip on hardwareBreakpoint?

src/integration-tests/launchWithEnvironment.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+                'Path does not include appended folder'
+            ).to.includes(pathToAppend);
+        } else {
+            const entriesInPath = APP_PROC.ENV_PATH!.split(path.delimiter);
+            expect(entriesInPath[0]).to.equals(pathToAppend);
+        }
+    });
+
+    it('check setting null will delete the variable', async function () {
+        if (platform() === 'win32' || hardwareBreakpoint) {
```
jonahgraham, line 330: Why skipped on win32? Can we not delete on win32?

src/integration-tests/launchWithEnvironment.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+        expect(APP_PROC.ENV_VARTEST4).to.equals('TEST4');
+
+        expect(GDB_PROC.SHOW_VARTEST1).to.equals('VARTEST1 = TEST1');
+        expect(GDB_PROC.SHOW_VARTEST2).to.equals('VARTEST2 = TEST2');
+        expect(GDB_PROC.SHOW_VARTEST3).to.equals('VARTEST3 = TEST3');
+        expect(GDB_PROC.SHOW_VARTEST4).to.equals('VARTEST4 = TEST4');
+    });
+
+    it('checks setting environment variables with debugAdapter', async function () {
+        if (hardwareBreakpoint || (platform() === 'win32' && !isRemoteTest)) {
```
jonahgraham, line 179: Why this skip condition?

src/integration-tests/launchWithEnvironment.spec.ts
```diff
@@ -0,0 +1,345 @@
...
+
+        expect(APP_PROC.ENV_PATH).not.to.equals(null);
+    });
+
+    it('ensures that new entries could be injected to path', async function () {
+        if (hardwareBreakpoint) {
+            this.skip();
+        }
+        const pathToAppend = __dirname;
+        const currentPathValue = process.env.PATH || process.env.Path;
```
jonahgraham, line 301: This is probably fine, but path can be anycase (i.e. `process.env.path`). In practice it almost never is anything other than PATH or Path on the test machines.

src/integration-tests/utils.ts
```diff
@@ -111,6 +111,21 @@ export function verifyVariable(
...
+/**
+ * Parses and returns a given char* variable value
+ */
+export function getCharStringVariableValue(
+    variable: DebugProtocol.Variable
+): string | null | undefined {
+    if (variable.value === '0x0') {
+        return null;
+    }
+    // IMPORTANT: When value of the string is long, it returns dots `...` at the end,
```
jonahgraham, line 123: I don't really like this :-(

Instead I recommend sending a command to gdb to disable this behaviour for the test with `set print elements 0` - see https://sourceware.org/gdb/onlinedocs/gdb/Print-Settings.html#index-number-of-array-elements-to-print

Alternatively use `printf "%s", name` as the gdb command to get the value with the custom command.

The printf option has the advantage that you don't need to do regex to get the string out.

jonahgraham (changes requested):
This LGTM, some small issues with the tests to look at and it should be ready to go.

src/integration-tests/launchWithEnvironment.spec.ts
asimgunes: Getting error in hardwareBreakpoint condition: 
"No hardware breakpoint support in the target." 
I skipped the test since the target of the test is not related to the hardwareBreakpoint condition.

src/integration-tests/launchWithEnvironment.spec.ts
asimgunes: Win32 test environment is injecting another folder into the path so it never became null.

src/util.ts
asimgunes: This code is no longer exists in the update.

src/integration-tests/util.spec.ts
asimgunes: The functinality removed and no longer exists in the update.

src/integration-tests/util.spec.ts
asimgunes: The functinality removed and no longer exists in the update.

src/integration-tests/launchWithEnvironment.spec.ts
asimgunes: I fixed this issue and remoed the platform() condition.

src/integration-tests/launchWithEnvironment.spec.ts
asimgunes: I changed the skip condition for hardwareBreakpoint only. I realise that win32 is failing because of the recent test problem in Windows tests.

src/integration-tests/utils.ts
asimgunes: With print I got other complications in NULL checks (throwing error). I used `set print addr off` option and continue with that. Could you please check the current implementation?

src/integration-tests/launchWithEnvironment.spec.ts
asimgunes: Not sure to update this one, because I am returning checking the same casing in C code too. "ensures that path is not null" test is implicitly controlling this. I can still update this if you think it could cause a problem in the future.

asimgunes:
Hi @jonahgraham,

I made some updates about your comments, Could you please re-check the updates when you are available?

Kind regards.

src/integration-tests/launchWithEnvironment.spec.ts
asimgunes: As I mentioned, the problem is related to the test environment, I added an arbitrary environment variable named "ENV_TEST_VAR" to control the deleting variable functionality and updated this code to also check it in Windows.

src/integration-tests/launchWithEnvironment.spec.ts
jonahgraham: The reason you were getting those errors is you were using `debugAdapter.js` when `--test-remote` was in use. The reason it fails is that hardware breakpoints can only be used in remote (target) debugging. See my commit where I fix this up consistent with how other tests force using the target adapter.

src/integration-tests/test-programs/vars_env.c
```diff
@@ -0,0 +1,23 @@
+#include <stdio.h>
+#include <stdlib.h>
+
+int main(int argc, char *argv[])
+{
+    char *path, *test1, *test2, *test3, *test4, *envtest;
+    path = getenv("PATH");
+    if(path == NULL) {
+        path = getenv("Path");
+    }
```
jonahgraham, line 10: This extra check isnt' necessary - envvars are [case insensitive in Windows' getenv](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/getenv-wgetenv?view=msvc-170).

```suggestion
    path = getenv("PATH");
```

src/integration-tests/launchWithEnvironment.spec.ts
jonahgraham: I just tested the assumption here that KEY in `process.env.KEY` is case sensitive, it is not case sensitive on Windows, which matches the [documentation](https://nodejs.org/api/process.html#processenv):

> On Windows operating systems, environment variables are case-insensitive.


```suggestion
        const currentPathValue = process.env.PATH;
```

jonahgraham (changes requested):

src/integration-tests/launchWithEnvironment.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+
+        const bpResp = await dc.setBreakpointsRequest({
+            source: {
+                name: 'vars_env.c',
+                path: path.join(testProgramsDir, 'vars_env.c'),
+            },
+            breakpoints: [
+                {
+                    column: 1,
+                    line: 22,
```
jonahgraham: Lets remove hardcoded line numbers, at least from new tests. Some old tests still use them. See latest commit.

jonahgraham:
I have now fixed the Windows issue, rebasing this on main will pull in the fix for #299

jonahgraham (approved):

jonahgraham:
@asimgunes this is now merged - are there other changes incoming, or should I do a publish to npm and version bump?

Will you provide the corresponding change to cdt-gdb-vscode?

asimgunes:
Hi @jonahgraham,

Thanks for the merge, there is no upcoming change from my side, and sure I can send an update the cdt-gdb-vscode references once you publish the recent cdt-gdb-adapter to npm.

jonahgraham:
@asimgunes - I have published new version - https://www.npmjs.com/package/cdt-gdb-adapter/v/0.0.29 - can you consume that update when you make the cdt-gdb-vscode changes?

jonahgraham:
> can you consume that update when you make the cdt-gdb-vscode changes?

I am making the change to cdt-gdb-vscode now to pull in the fix in #306

PR #309, opened 2023-10-26
Support SteppingGranularity & instruction stepping

colin-grant-work:
This PR adds support for `SteppingGranularity` to the `next` and `stepIn` requests, e.g. from the VSCode disassembly view.

To do:
 - [x] I'll add some tests ~tomorrow~ next week.

### DIFF
--- a/src/GDBDebugSession.ts
+++ b/src/GDBDebugSession.ts
@@ -335,6 +335,7 @@ export class GDBDebugSession extends LoggingDebugSession {
         response.body.supportsDisassembleRequest = true;
         response.body.supportsReadMemoryRequest = true;
         response.body.supportsWriteMemoryRequest = true;
+        response.body.supportsSteppingGranularity = true;
         this.sendResponse(response);
     }
 
@@ -1003,7 +1004,9 @@ export class GDBDebugSession extends LoggingDebugSession {
         args: DebugProtocol.NextArguments
     ): Promise<void> {
         try {
-            await mi.sendExecNext(this.gdb, args.threadId);
+            await (args.granularity === 'instruction'
+                ? mi.sendExecNextInstruction(this.gdb, args.threadId)
+                : mi.sendExecNext(this.gdb, args.threadId));
             this.sendResponse(response);
         } catch (err) {
             this.sendErrorResponse(
@@ -1019,7 +1022,9 @@ export class GDBDebugSession extends LoggingDebugSession {
         args: DebugProtocol.StepInArguments
     ): Promise<void> {
         try {
-            await mi.sendExecStep(this.gdb, args.threadId);
+            await (args.granularity === 'instruction'
+                ? mi.sendExecStepInstruction(this.gdb, args.threadId)
+                : mi.sendExecStep(this.gdb, args.threadId));
             this.sendResponse(response);
         } catch (err) {
             this.sendErrorResponse(
--- a/src/integration-tests/stepping-granularity.spec.ts
+++ b/src/integration-tests/stepping-granularity.spec.ts
@@ -1,0 +1,200 @@
+/*********************************************************************
+ * Copyright (c) 2023 Ericsson and others
+ *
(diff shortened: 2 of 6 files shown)


src/GDBDebugSession.ts
```diff
@@ -1003,7 +1004,7 @@ export class GDBDebugSession extends LoggingDebugSession {
         args: DebugProtocol.NextArguments
     ): Promise<void> {
         try {
-            await mi.sendExecNext(this.gdb, args.threadId);
+            await (args.granularity === 'instruction' ? mi.sendExecStepInstruction(this.gdb, args.threadId) : mi.sendExecNext(this.gdb, args.threadId));
```
jonahgraham, line 1007: For `next` I think you should issue `-exec-next-instruction` instead.

jonahgraham (commented):
I am delighted to see you brining this to the adapter.

jonahgraham:
> [Battle of the formatters :-/](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/309/commits/8f9ff803ee3e97f251edf0d9002b847a8fdb5fbe)

Who is battling? "yarn format" uses prettier, I have prettier installed in the IDE so it formats on save. Maybe we should add it to `recommendations` in [extension.json](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/92bb15046fea82256742a69f0b240129a1949a76/.vscode/extensions.json)?

colin-grant-work:
> Who is battling? "yarn format" uses prettier, I have prettier installed in the IDE so it formats on save. Maybe we should add it to recommendations in [extension.json](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/92bb15046fea82256742a69f0b240129a1949a76/.vscode/extensions.json)?

Yeah, it's the battle between VSCode's default formatter and Prettier. It would likely be helpful to add Prettier to the recommendations & add 

```
editor.defaultFormatter: `esbenp.prettier-vscode`
```
to the `settings.json`. Then the only thing missing for a perfect set-up is for contributors to set other things as the default formatter _elsewhere_ so having the Prettier plugin installed doesn't make formatting harder in other repos. Or to set Prettier as disabled in other workspaces and enabled here. I had things more nicely set up for a while, but I jump around a lot and use a lot of ad-hoc multi-root workspaces, so it got messy again.

jonahgraham:
Your change looks good to me, I have made a couple of commits to make the test code more generic. Please see commits for details as to why I suggest such changes.

- 510fcf7 Fix warning in C code:
- 8e964ea Use line tags instead of hard coded line numbers
- 61bc8e5 make the line of code the same as opening brace to account for different gdb/gcc combinations

colin-grant-work:
Thanks for the fixes! Could you explain the significance of this part?

> make the line of code the same as opening brace to account for different gdb/gcc combinations

I saw that comment / commit message in other test cases, but I didn't and still don't understand why that particular change is important.

jonahgraham:
For a function like this:

```c
int method(void) {
  call_something();
}
```

some combinations of GCC/GDB versions consider line 1 (line with `{`) the first line of the method and some consider it line 2 (line with first instructions). So to avoid cases where we want to detect we are on the first source line of a method we place `{` and first code on same source line so we don't have to identify the differences in the tests.

If you want I can dig through my history for which combinations do which thing.

Is that change ok to you?

colin-grant-work:
Yeah, definitely OK - just wanted to understand better why it was needed.

jonahgraham (approved):
Can you squash and rebase (and then merge if CI approves)?

colin-grant-work:
Sorry for the multiple pushes - was confused about where your commits lived, but I believe I've got them now.

jonahgraham:
> Sorry for the multiple pushes - was confused about where your commits lived, but I believe I've got them now.

No worries - it is sort of weird that reviewers can push to contributor branches :-)

Issue #311, opened 2023-10-27
MIParser hangs on nested objects


jonahgraham:
_This issue came out of #308_

The MI parser hangs when some types of nested objects are encountered. e.g. breakpoint messages with `script` such as this simplified example:

```
=breakpoint-modified,bkpt={number="3",type="breakpoint",script={"commands here"}}
```

jonahgraham:
What is missing is that an MI Tuple can have _just_ a MI Value, [in CDT it is handled like this](https://github.com/eclipse-cdt/cdt/blob/caf2292768deccd885b5b6989b731742e2e5edf4/dsf-gdb/org.eclipse.cdt.dsf.gdb/src/org/eclipse/cdt/dsf/mi/service/command/output/MIParser.java#L347-L349):

```java
			// Try for the DsfMIValue first
			MIValue value = processMIValue(buffer);
			if (value != null) {
```

While in cdt-gdb-adapter it is assumed that MI Tuple (called Object in this code) assumes that it is only key value pairs:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/edb0d46effb1297e916458fa7f7cfc91dc8ec8e1/src/MIParser.ts#L164-L173

XingMicrochip:
Thanks @jonahgraham for explaining the issue. I can look at this issue.

jonahgraham:
I've been writing some tests for the miparser which should make it easier to develop. See #312, but I have a bigger change pending which I expect I'll push today.

XingMicrochip:
Are you going to push a change that solves this MIParser bug, or is that change to add test cases for this issue? - i.e. Should I work on fixing this bug?

jonahgraham:
Just the tests for now.

XingMicrochip:
I see where the issue is but I do not fully understand now why it happens in `handleObject()`. Say the example looks like below, can you walk through the code to help me understand which part hangs, to help me develop a fix?
```
{number="1",type="breakpoint",disp="keep",enabled="y",addr="0x8000150c",func="main",file="C:/Microchip/...",fullname="C:\\Microchip\\...",line="2029",thread-groups=["i1"],times="0",script={"p 123","p 456","p 789"},original-location="main"}
```

jonahgraham:
[...]

PR #313, opened 2023-11-01
Add special case for MIParser to parse 'objects' with values but no keys

XingMicrochip:
@jonahgraham This PR is for [issue #311](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/311). I add a special case in `MIParser` to parse objects that are not formatted as name-value pairs. I have tested this change from the IDE in our application on both Windows and Linux. I added a test case.


jonahgraham:
This looks promising. It looks like one of the tests failed, I expect it was an intermittent failure so I restarted the tests. Once tests pass I will have a closer look.

jonahgraham:
_sorry, I didn't get to looking at this today, I'll try tomorrow._

jonahgraham (changes requested):
This change works for the inputs provided, but falls over fairly easily if anything more complicated is put in the "script" block. This is because "handleString" can't be used to parse a "c-string" which is a C-style string.

Consider an input where the quoted string contains control characters, like `}`:

```
            '+message,bkpt={number="1",type="breakpoint",thread-groups=["i1"],script={"p }123","p 321","p 789"}}'
```

Also, it is generally bad form to drop the leading and trailing `"` without verifying that it is a `"`-quoted string.

This is why in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/311#issuecomment-1783547806 I recommended checking if there a a value before doing a key=value parse. e.g. call handleValue() first to see if the next token is a value. Note that the handleValue has a `default` case that isn't suitable for this use case so some small refactoring will be needed around that.

XingMicrochip:
You are right. I will work on a fix.

src/MIParser.ts
```diff
@@ -177,15 +177,35 @@ export class MIParser {
...
+                        result[name] = this.handleValue();
+                    }
+                    c = this.next();
                 }
+            } else {
+                // "object" contains just values
                 c = this.next();
+                let key = 0;
+                while (c !== '}') {
+                    let value = '';
```
jonahgraham, line 197: Please consider using handleCString to parse a c-string here as this code still will fail if there is an embedded escaped `"` in the value. Also handleCString will also unescape the string.

jonahgraham (approved):
While it is theoretically possible for other types of messages to show up, in practice I don't know if it happens, so we can defer fixing those cases until we have a concrete example using `handleValue` instead of just checking if it is a `c-string`

e.g these cases:

```
script={["v1","v2"}}
script={{key="value"}}
script={"cstring",key="value"}
```

jonahgraham:
Thank you @XingMicrochip - I will publish this and an updated cdt-gdb-vscode soon.

jonahgraham:
I have tagged [cdt-gdb-vscode](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/releases/tag/v0.0.106) and tested the vsix with this:

script:
```
b main
commands
print "Arrived at main"
end
```

and this in the launch.json:
```
      "initCommands": ["source ${workspaceFolder}/script"],
```

I can see when running that I get the following in the Debug Console:

```
$1 = "Arrived at main"
```

XingMicrochip:
I think `v0.0.106` is broken. I just downloaded it from `https://open-vsx.org/api/eclipse-cdt/cdt-gdb-vscode/0.0.106/file/eclipse-cdt.cdt-gdb-vscode-0.0.106.vsix` and integrated it into our IDE. We couldn't launch the debugger with the same launch configuration. The last version we used was `v0.0.103`. I also noticed the adapter source files are now bundled. Could there be something in between `v0.0.103` and `v0.0.106` that broke the adapter?

jonahgraham:
Yes, the packaging of cdt-gdb-vscode has changed since 0.0.103 - see https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/96

Can you open a bug in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/ with what you see. You are not the only one to experience problems with this change, but I cannot reproduce the current problem (I fixed an [earlier one](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/105))

> It looks like this now, while it used to be that all the source files are seperated:

BTW That change is the main part of https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/96

Issue #317, opened 2023-11-15
On Windows, "\r\n" sends an extra new line


XingMicrochip:
Hi @jonahgraham,

When we debug on Windows through the IDE using the adapter, we see an extra new line if we send a `\r\n` character to the UART. This only happens on Windows, but not Linux. If we send `\n`, then it works on both Windows and Linux (there's no extra new line). We suspect it's an issue with the adapter because `\r\n` works fine if we debug through the command line (using a manual flow) without the adapter. Do you by any chance know why this happens and if it's an issue with the adapter?

Relevant part of the application code:
```
UART_polled_tx_string(&g_core_uart_0, "CoreTimer timer interrupt.\r\n");
```
Windows IDE debug console:
```
CoreTimer timer interrupt.

CoreTimer timer interrupt.

CoreTimer timer interrupt.

CoreTimer timer interrupt.

CoreTimer timer interrupt.

```
Linux IDE debug console (also the expected output):
```
CoreTimer timer interrupt.
CoreTimer timer interrupt.
CoreTimer timer interrupt.
CoreTimer timer interrupt.
CoreTimer timer interrupt.
```

jonahgraham:
@XingMicrochip can you look at what @AdhamRagabMCHP did in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/280 as it seems to have been trying to solve the opposite issue of missing newline. It certainly sounds like something the adapter does.

XingMicrochip:
It seems @AdhamRagabMCHP faced the same issue here: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/280#issuecomment-1630899743. My debug log looks like this on Windows, which I think causes the extra new line issue (on Linux there's no extra `\r`):
```
[14:32:05.909 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"Socket","output":"CoreTimer timer interrupt.\r\r\n"}}
[14:32:06.552 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"Socket","output":"CoreTimer timer interrupt.\r\r\n"}}
[14:32:07.161 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"Socket","output":"CoreTimer timer interrupt.\r\r\n"}}
[...]

PR #318, opened 2023-11-16
Fix bug with parsing UART output

XingMicrochip:
This PR fixes _[issue #317](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/317)_.

Problematic code snippet in `GDBTargetDebugSession.ts`:
```
            const eolChar: string =
                uart.eolCharacter === 'CRLF' ? '\r\n' : '\n';

            let tcpUartData = '';
            this.socket.on('data', (data: string) => {
                for (const char of data) {
                    if (char === eolChar) {
                        this.sendEvent(
                            new OutputEvent(tcpUartData + os.EOL, 'Socket')
                        );
                        tcpUartData = '';
                    } else {
                        tcpUartData += char;
                    }
                }
            });
```
If `data` is `CoreTimer timer interrupt.\r\n` - for example - without setting `eolCharacter` in launch configuration, the adapter uses default `\n` to parse `data`. It accumulates `CoreTimer timer interrupt.\r` in variable `tcpUartData` and then appends `os.EOL` at the end. On Windows, the final `tcpUartData` becomes `CoreTimer timer interrupt.\r\r\n`. The extra new line comes from here. To fix this, the adapter appends `\n` instead of `\r\n` because `\r` is already in `tcpUartData`.

I tried setting `eolCharacter` to `CRLF` in launch configuration. This didn't work. `char` is one character in length, but `eolChar` is two in length, so `char === eolChar` is never `true`. The adapter also can't parse two characters at a time because one burst of `data` might be `CoreTimer timer interrupt.\r` and the next one `\n`. If this happens, `\r\n` will never be detected together.

I have tested this change fixes the problem I described in _[issue #317](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/317)_.

### DIFF
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -391,15 +391,12 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             this.socket = new Socket();
             this.socket.setEncoding('utf-8');
 
-            const eolChar: string =
-                uart.eolCharacter === 'CRLF' ? '\r\n' : '\n';
-
             let tcpUartData = '';
             this.socket.on('data', (data: string) => {
                 for (const char of data) {
-                    if (char === eolChar) {
+                    if (char === '\n') {
                         this.sendEvent(
-                            new OutputEvent(tcpUartData + os.EOL, 'Socket')
+                            new OutputEvent(tcpUartData + '\n', 'Socket')
                         );
                         tcpUartData = '';
                     } else {
--- a/src/integration-tests/test-programs/socketServer.js
+++ b/src/integration-tests/test-programs/socketServer.js
@@ -1,5 +1,6 @@
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const net = require('net');
+const os = require('os');
 
 // Create socket echo server.
 const socketServer = new net.Server();
@@ -12,7 +13,7 @@ socketServer.on('connection', (connection) => {
     });
 
     // Echo "Hello World!"
-    connection.write('Hello World!\n');
+    connection.write(`Hello World!${os.EOL}`);
 });
 
 socketServer.on('close', () => {


src/GDBTargetDebugSession.ts
```diff
@@ -391,15 +391,12 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
             this.socket.setEncoding('utf-8');
 
-            const eolChar: string =
-                uart.eolCharacter === 'CRLF' ? '\r\n' : '\n';
-
             let tcpUartData = '';
             this.socket.on('data', (data: string) => {
                 for (const char of data) {
-                    if (char === eolChar) {
+                    if (char === '\n') {
```
jonahgraham, line 397: This looks fine - indeed I don't know how it matched anything when `eolCharacter` was `\r\n` as char would presumably be a single character.

src/GDBTargetDebugSession.ts
```diff
@@ -391,15 +391,12 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             this.socket = new Socket();
             this.socket.setEncoding('utf-8');
 
-            const eolChar: string =
-                uart.eolCharacter === 'CRLF' ? '\r\n' : '\n';
```
jonahgraham, line 395: Is `eolCharacter` used elsewhere?

jonahgraham (commented):
This change looks fine to me, but it would be nice to have a test that demonstrates it and prevents it regressing.

src/GDBTargetDebugSession.ts
XingMicrochip: It's used here:
```
            const SerialUartParser = new ReadlineParser({
                delimiter: uart.eolCharacter === 'CRLF' ? '\r\n' : '\n',
                encoding: 'utf8',
            });
```

XingMicrochip:
I will add a test.

jonahgraham:
> I will add a test.

Great! Seeing as there are already some failing tests with the change you provided it may be worth looking into whether there is already a test that just needs a modification.

XingMicrochip:
Can you re-run the tests? @jonahgraham I don't know what the one that's failing is about.

XingMicrochip:
All the tests are passing now. Can you release a new version on Open VSX for this change?

jonahgraham:
> All the tests are passing now. Can you release a new version on Open VSX for this change? Thanks.

I'll try to look at it soon, but it may have to wait until Monday.

XingMicrochip:
That works. Thank you.

jonahgraham (approved):

PR #321, opened 2023-12-05
Enhancing ImageAndSymbols to load several binaries

FlorentVSTM:
Currently, the "imageAndSymbols" option can be used to load one additional file. To load multiple files you can manually load them via the "initCommands" option.
This PR enhances the ImageAndSymbols property to manage this use case.
It is linked to the PR: https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/117

### DIFF
--- a/src/GDBTargetDebugSession.ts
+++ b/src/GDBTargetDebugSession.ts
@@ -79,18 +79,26 @@ export interface TargetLaunchArguments extends TargetAttachArguments {
     cwd?: string;
 }
 
-export interface ImageAndSymbolArguments {
+export interface SymbolFileAndOffset {
     // If specified, a symbol file to load at the given (optional) offset
-    symbolFileName?: string;
-    symbolOffset?: string;
+    file: string;
+    offset?: string;
+}
+
+export interface ImageFileAndOffset {
     // If specified, an image file to load at the given (optional) offset
-    imageFileName?: string;
-    imageOffset?: string;
+    file: string;
+    offset?: string;
+}
+
+export interface ImageAndSymbolsArguments {
+    symbolFilesAndOffset?: SymbolFileAndOffset[];
+    imageFilesAndOffset?: ImageFileAndOffset[]
 }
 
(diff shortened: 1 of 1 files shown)


jonahgraham (commented):
LGTM - needs to pass builds and then I can publish it and do the same thing for the corresponding vscode part.

Any thoughts on how to test this one?

jonahgraham:
I haven't been able to come up with a way to test the proper automation of this within our existing test suite in a reasonable way.

Closest I have come is a bunch of extra steps to take a program like this (`example.c` compiled to `example`):

```c
#include <stdio.h>

char data[1000]; 

int main(int argc, char *argv[])
{
    puts(data);
    return 0;
}
````

and then load an srec into the data (srec created from a text file like this: `objcopy -I binary -O srec data.txt data.srec`). The problem is what the addresses should be, the addresses need to be calculated by gdb (and assume address randomization is off). To give an idea of how, something like this perhaps:

```
(gdb) # print out the data address
(gdb) p &data # print out the data address
$1 = (char (*)[1000]) 0x555555558040 <data>
(gdb) # print out the PC so we can reset it after the load
(gdb) p $pc
$1 = (void (*)()) 0x7ffff7fe3290 <_start>
(gdb) load data.srec 0x555555558040
Loading section .sec1, size 0x5c lma 0x555555558040
Start address 0x0000000000000000, load size 92
Transfer rate: 736 bits in <1 sec, 92 bytes/write.
(gdb) # reset the PC because the load messed it up
(gdb)  set $pc = 0x7ffff7fe3290
(gdb) # continue and observe that data has the correct value (e.g. read memory, or see the output in the terminal)
```

For the above, the server side command was:

```sh
$ gdbserver --once :1234 example
```

and gdb was started like this:

```sh
gdb -ex 'target remote :1234' empty
```

So we could launch gdb once, print the values of `$pc` and `&data`, then construct a new launch configuration with those values to test it out. But for now that seems like too much effort for the value it brings.

PR #339, opened 2024-11-07
Add API to pause target and wait for debugger to be suspended

### LINKED ISSUE #295: Add API to easily pause target and wait for it to be ready

Luke-zhang-mchp:
Closes #295.

Code to suspend/continue the debug session when setting breakpoints are moved to a reusable function `pause` and `continue` methods. The `pause` method `await`s a promise which gets resolves when the target is actually suspended.

This `pause` method is also used in `GDBTargetDebugSession::disconnectRequest`.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -328,28 +328,47 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         return this.gdb?.spawn(args);
     }
 
+    /**
+     * Sends a pause command to GDBBackend, and resolves when the debugger is
+     * actually paused
+     */
+    protected async pause(): Promise<void> {
+        const waitPromise = new Promise<void>((resolve) => {
+            this.waitPaused = resolve;
+        });
+        if (this.gdb.isNonStopMode()) {
+            const threadInfo = await mi.sendThreadInfoRequest(this.gdb, {});
+
+            this.waitPausedThreadId = parseInt(
+                threadInfo['current-thread-id'],
+                10
+            );
+            this.gdb.pause(this.waitPausedThreadId);
+        } else {
+            this.gdb.pause();
+        }
+
+        // This promise resolves when handling GDBAsync for the "stopped"
+        // result class, which indicates that the call to `GDBBackend::pause`
+        // is actually finished.
+        await waitPromise;
+    }
+
+    protected async continue(): Promise<void> {
+        if (this.gdb.isNonStopMode()) {
+            await mi.sendExecContinue(this.gdb, this.waitPausedThreadId);
+        } else {
+            await mi.sendExecContinue(this.gdb);
+        }
+    }
+
     protected async setBreakPointsRequest(
         response: DebugProtocol.SetBreakpointsResponse,
         args: DebugProtocol.SetBreakpointsArguments
     ): Promise<void> {
         this.waitPausedNeeded = this.isRunning;
         if (this.waitPausedNeeded) {
-            // Need to pause first
-            const waitPromise = new Promise<void>((resolve) => {
-                this.waitPaused = resolve;
-            });
-            if (this.gdb.isNonStopMode()) {
-                const threadInfo = await mi.sendThreadInfoRequest(this.gdb, {});
-
-                this.waitPausedThreadId = parseInt(
-                    threadInfo['current-thread-id'],
-                    10
-                );
-                this.gdb.pause(this.waitPausedThreadId);
-            } else {
-                this.gdb.pause();
-            }
-            await waitPromise;
+            await this.pause();
         }
 
         try {
@@ -519,11 +538,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         }
 
         if (this.waitPausedNeeded) {
-            if (this.gdb.isNonStopMode()) {
-                mi.sendExecContinue(this.gdb, this.waitPausedThreadId);
-            } else {
-                mi.sendExecContinue(this.gdb);
-            }
+            await this.continue();
         }
     }
 
@@ -534,21 +549,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.waitPausedNeeded = this.isRunning;
         if (this.waitPausedNeeded) {
             // Need to pause first
-            const waitPromise = new Promise<void>((resolve) => {
-                this.waitPaused = resolve;
-            });
-            if (this.gdb.isNonStopMode()) {
-                const threadInfo = await mi.sendThreadInfoRequest(this.gdb, {});
-
-                this.waitPausedThreadId = parseInt(
-                    threadInfo['current-thread-id'],
-                    10
-                );
-                this.gdb.pause(this.waitPausedThreadId);
-            } else {
-                this.gdb.pause();
-            }
-            await waitPromise;
+            await this.pause();
         }
 
         try {
@@ -646,11 +647,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         }
 
         if (this.waitPausedNeeded) {
-            if (this.gdb.isNonStopMode()) {
-                mi.sendExecContinue(this.gdb, this.waitPausedThreadId);
-            } else {
-                mi.sendExecContinue(this.gdb);
-            }
+            await this.continue();
         }
     }
 
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -331,9 +331,21 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     /**
      * Sends a pause command to GDBBackend, and resolves when the debugger is
      * actually paused. The paused thread ID is saved to `this.waitPausedThreadId`.
-     *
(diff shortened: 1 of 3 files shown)


Luke-zhang-mchp:
Hey @jonahgraham, this is what I assume the issue was alluding to, and it works fine on our IDE. Let me know how the changes look.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -328,28 +328,47 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         return this.gdb?.spawn(args);
     }
 
+    /**
+     * Sends a pause command to GDBBackend, and resolves when the debugger is
+     * actually paused
+     */
+    protected async pause(): Promise<void> {
```
Luke-zhang-mchp, line 335: I think this name is questionable, since there's also `pauseRequest` and `continueRequest`, but I'm not sure what a good name would be. Something like `pauseAndWait`?

jonahgraham (commented):
This looks quite close to what I was talking about - but I was hoping the logic for "waitPauseNeeded" could be in the new method.

i.e. have a single call that is "pauseIfNeeded" that can be awaited, and then at the end of the block "resumeIfNeeded" that is perhaps passed a token pauseIfNeeded returned (or elsewhere stored in the state).

---
sort of related topic brain dump:

This is a place I really like Python's `with` statement, because in Python I would write something like:

```python
    with this.targetPaused():
        // do some operations on a paused target
```

which will pause if needed at the beginning and resume automatically when leaving the block.

Luke-zhang-mchp:
Yeah that with statement would be really nice. What about [method decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators)? It's not really the same cause it would be the beginning and end of the function. Also probably not worth the trouble, but it would be something like:

```ts
const paused = <T extends []>(
    target: GDBDebugSessionBase,
    propertyKey: string,
    descriptor: TypedPropertyDescriptor<(...args: T[]) => void>
) => {
    const originalMethod = descriptor.value; // Original function

    descriptor.value = function (...args: T[]) {
        target.pause();
        originalMethod?.apply(this, args);
        target.continue();
    };
};

export abstract class GDBDebugSessionBase extends LoggingDebugSession {
    @paused
    protected async setBreakPointsRequest(
        response: DebugProtocol.SetBreakpointsResponse,
        args: DebugProtocol.SetBreakpointsArguments
    ): Promise<void> {}
}
```

jonahgraham:
> What about [method decorators](https://www.typescriptlang.org/docs/handbook/decorators.html#method-decorators)?

My lack of experience with typescript means I don't know if this is a good idea or not in this case. 

> Also probably not worth the trouble

That is my instinct for now.

jonahgraham (commented):
This change looks correct - but it also looks like it exposes some issues in the tests when async is off.

For example I can see in `logpoints/remote gdb-async-off/hits a logpoint.log` that at the moment that the SIGINT is sent to gdb, inferior happens to be paused, leading to this output:

```
[21:00:09.996 UTC] GDB signal: SIGINT to pid 13380
[21:00:09.996 UTC] GDB notify async: breakpoint-modified,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0x000055555555513f",func="main",file="count.c",fullname="/home/runner/work/cdt-gdb-adapter/cdt-gdb-adapter/src/integration-tests/test-programs/count.c",line="4",thread-groups=["i1"],times="2",original-location="-source /home/runner/work/cdt-gdb-adapter/cdt-gdb-adapter/src/integration-tests/test-programs/count.c -line 4"}
[21:00:09.996 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"log","output":"Quit\n"}}
```

What this means is that the adapter thought the inferior was running, so sent a SIGINT, but in the time between the test and SIGINT being delivered, the inferior hit a breakpoint, so the SIGINT was interpreted as Quit instead of pause.

This is a fundamental problem with async mode because this race condition can not be detected or avoided.  What I don't know is if/why this change exposes this problem more, or if it was just bad luck. I have restarted the tests in the first instance.

With async on by default for the last couple of years I don't think this matters too much. But we do need to have  a clean run. If the above isn't enough information to resolve the issue, let me know and I can try to fix up these problematic test cases.

src/web/GDBTargetDebugSession.ts
```diff
@@ -326,15 +326,8 @@ export class GDBTargetDebugSession extends GDBDebugSession {
...
-                if (this.gdb.getAsyncMode() && this.isRunning) {
-                    // See #295 - this use of "await" is to try to slightly delay the
-                    // call to disconnect. A proper solution that waits for the
-                    // interrupt to be successful is needed to avoid future
-                    // "Cannot execute this command while the target is running"
-                    // errors
-                    await this.gdb.sendCommand('interrupt');
-                }
-
+                // Need to pause first, then disconnect and exit
```
jonahgraham, line 329: The test failures in async mode are due to this change which expose the tests to the previously mentioned race condition more easily. For example continues.spec.ts runs code that hits a breakpoint in a tight loop, so in async mode the `disconnect` will be processed after hitting the next breakpoint without this change.

With this change a SIGINT is sent.

I think this is a test problem in async code and should be fixed in the tests rather than changing the code.

Let me know if you understand enough of what I have said to change the tests, if not I will fix the tests and we can rebase this change.

jonahgraham (changes requested):

src/web/GDBTargetDebugSession.ts
Luke-zhang-04: Oh sorry, just saw this. I did get it working with the commit I just pushed (https://github.com/Luke-zhang-mchp/cdt-gdb-adapter/actions/runs/11788461303), but I will look at fixing the tests instead.

jonahgraham: The commit you just pushed maintains the status quo of having the same disconnect bug (in sync mode). We can just go with that for now and fix the disconnect bug in sync mode (or more likely leave it documented as known issue in a mode we don't recommend anyway)

PS. The last commit was pushed as a different author email, can you force push a new commit fixing the author please:

image

Luke-zhang-04: Ah shoot, I didn't think that would be much of an issue, I don't have my work laptop with me. I'll take a closer look tomorrow and see if I can fix it in the tests.

jonahgraham: > Just to make sure, you meant to say sync mode right? It seems to make more sense that way.

Not sure which line ^^^ was in response to, so....

> The commit you just pushed maintains the status quo of having the same disconnect bug (in sync mode). We can just go with that for now and fix the disconnect bug in sync mode (or more likely leave it documented as known issue in a mode we don't recommend anyway)

... I'll rephrase ^^^ because it wasn't super clear

The change you have added with `requireAsync = false` parameter added to `pauseIfNeeded` is acceptable for now. It means that in sync mode `disconnectRequest` doesn't work reliably, but that is ok as that is the status quo.

I would like API to not recommend that `requireAsync` be used at all and that it will be removed in the future. We can state that `pauseIfNeeded` and `disconnectRequest` don't work reliably in sync mode and recommend users (extenders really) only use async mode.

If someone has a need for reliable sync mode, then they can contribute fixes/improvements.

Luke-zhang-mchp: Ok so I just spent the day trying to fix the tests as you suggested earlier and... I think it's time for me to give up. The multithread one is particularly annoying to debug, since it's only an issue every 1 in 20 runs on my machine. I did manage to fix the logpoint failure consistently (312d977d06c4305e9007fd166b9162e491d985d3). I think the last straw was when the test from `continues` failed (https://github.com/Luke-zhang-mchp/cdt-gdb-adapter/actions/runs/11826261580/job/32951836487#step:8:1024), and I realized that the issues in sync mode are probably deeper than I had thought.

Just for the record, here are all the changes I ended up making: https://github.com/Luke-zhang-mchp/cdt-gdb-adapter/compare/8b3628ec62e1e3cf851fc189b5f5db4cdae4f643...Luke-zhang-mchp:cdt-gdb-adapter:61e9ff046c09528c5b8020acf8eca0475ef7b5de.

I'll make the changes for recommending against `requireAsync` in the form of deprecated overloads.

Luke-zhang-mchp: I added warning comments, but they don't do a great job at warning about sync mode specifically. Are you also looking for runtime warnings?

jonahgraham: What you have done here is what I had hoped for. Thanks for spending time looking at the other tests, it is great to have more people familiar with how all this works to be able to maintain this long term.

Thanks @Luke-zhang-mchp - I am going to merge this version of the change.

jonahgraham (approved):

Issue #340, opened 2024-11-18
Incorrect address (pc) resolution in DisassembleRequest


asimgunes:
During the debug session, when user navigates and opens the `Disassembly View` in the VSCode, the Disassembly View focuses on an incorrect line when it is compared program counter (pc) register as shown above:

![image](https://github.com/user-attachments/assets/fc84e46c-a25b-494c-a777-0081c87c93dd)

After a review, I found that the current implementation of the `disassembleRequest` method at the `GDBDebugSessionBase` class are not handling the `instructionOffset` parameters. It seems the `disassembleRequest` implementation requires to be updated to match the DAP reference. 
https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disassemble 

@jonahgraham and I already discussed the issue previously, planning to propose an update to solve the issue.

PR #341, opened 2024-11-19
Improving the DisassembleRequest logic

asimgunes:
Hi @jonahgraham ,

Here is an update for the `disassembleRequest` logic in the debug adapter. This implementation is covering the scenarios discussed at the #340 . 

Could you please help me on reviewing this update?

**Notes:**
There was one issue that I was in doubt. The previous implementation was covering a scenario for the `endMemoryReference` argument. I couldn't find the argument in the DAP specifications. Since this argument doesn't exists on the current DAP protocol, I avoid to make an implementation for this argument. Please let me know if the argument still expected for any clients used by the cdt-gdb-adapter.

### DIFF
--- a/src/integration-tests/diassemble.spec.ts
+++ b/src/integration-tests/diassemble.spec.ts
@@ -13,7 +13,6 @@ import * as path from 'path';
 import { DebugProtocol } from '@vscode/debugprotocol/lib/debugProtocol';
 import { CdtDebugClient } from './debugClient';
 import { fillDefaults, standardBeforeEach, testProgramsDir } from './utils';
-import { calculateMemoryOffset } from '../util/calculateMemoryOffset';
 import { assert } from 'sinon';
 
 describe('Disassembly Test Suite', function () {
@@ -180,28 +179,17 @@ describe('Disassembly Test Suite', function () {
         );
     });
 
-    it('can handle disassemble at bad address', async function () {
-        const disassemble = (await dc.send('disassemble', {
-            memoryReference: '0x0',
-            instructionCount: 10,
-        })) as DebugProtocol.DisassembleResponse;
-
-        expect(disassemble).not.eq(undefined);
-        expect(disassemble.body).not.eq(undefined);
-        if (disassemble.body) {
-            const instructions = disassemble.body.instructions;
-            expect(instructions).to.have.lengthOf(10);
-            // Checking the invalid instructions content
-            instructions.forEach((inst, ix) => {
-                expect(inst.address).to.eq(
-                    calculateMemoryOffset('0x0', ix * 2)
-                );
-                expect(inst.address).to.have.lengthOf.greaterThan(0);
-                expect(inst.instruction).to.eq(
-                    'failed to retrieve instruction'
-                );
-                expect(inst.presentationHint).to.eq('invalid');
-            });
+    it('return error at bad address', async function () {
+        try {
+            await dc.send('disassemble', {
+                memoryReference: '0x0',
+                instructionCount: 10,
+            } as DebugProtocol.DisassembleArguments);
+            assert.fail('Should throw error!');
+        } catch (e) {
+            expect(e).to.be.deep.equal(
+                new Error('Cannot access memory at address 0x0')
+            );
         }
     });
 });
--- a/src/integration-tests/diassemble.spec.ts
+++ b/src/integration-tests/diassemble.spec.ts
@@ -13,12 +13,35 @@ import * as path from 'path';
 import { DebugProtocol } from '@vscode/debugprotocol/lib/debugProtocol';
 import { CdtDebugClient } from './debugClient';
 import { fillDefaults, standardBeforeEach, testProgramsDir } from './utils';
+import { calculateMemoryOffset } from '../util/calculateMemoryOffset';
 
 describe('Disassembly Test Suite', function () {
     let dc: CdtDebugClient;
     let frame: DebugProtocol.StackFrame;
     const disProgram = path.join(testProgramsDir, 'disassemble');
     const disSrc = path.join(testProgramsDir, 'disassemble.c');
+
+    const expectsGeneralDisassemble = (
+        disassemble: DebugProtocol.DisassembleResponse,
+        length: number,
+        ignoreEmptyInstructions?: boolean
+    ) => {
+        expect(disassemble).not.eq(undefined);
+        expect(disassemble.body).not.eq(undefined);
+        if (disassemble.body) {
+            const instructions = disassemble.body.instructions;
+            expect(instructions).to.have.lengthOf(length);
+            // the contents of the instructions are platform dependent, so instead
+            // make sure we have read fully
+            for (const i of instructions) {
+                expect(i.address).to.have.lengthOf.greaterThan(0);
+                expect(i.instruction).to.have.lengthOf.greaterThan(0);
+                if (!ignoreEmptyInstructions) {
+                    expect(i.instructionBytes).to.have.lengthOf.greaterThan(0);
+                }
+            }
+        }
+    };
 
     beforeEach(async function () {
         dc = await standardBeforeEach();
@@ -60,19 +83,8 @@ describe('Disassembly Test Suite', function () {
             memoryReference: 'main',
             instructionCount: 100,
         })) as DebugProtocol.DisassembleResponse;
-        expect(disassemble).not.eq(undefined);
-        expect(disassemble.body).not.eq(undefined);
-        if (disassemble.body) {
-            const instructions = disassemble.body.instructions;
-            expect(instructions).to.have.lengthOf(100);
-            // the contents of the instructions are platform dependent, so instead
-            // make sure we have read fully
-            for (const i of instructions) {
-                expect(i.address).to.have.lengthOf.greaterThan(0);
-                expect(i.instructionBytes).to.have.lengthOf.greaterThan(0);
-                expect(i.instruction).to.have.lengthOf.greaterThan(0);
-            }
-        }
+
+        expectsGeneralDisassemble(disassemble, 100);
     });
 
     it('can disassemble with no source references', async function () {
@@ -82,19 +94,75 @@ describe('Disassembly Test Suite', function () {
             memoryReference: 'main+1000',
             instructionCount: 100,
         })) as DebugProtocol.DisassembleResponse;
-        expect(disassemble).not.eq(undefined);
-        expect(disassemble.body).not.eq(undefined);
-        if (disassemble.body) {
-            const instructions = disassemble.body.instructions;
-            expect(instructions).to.have.lengthOf(100);
-            // the contents of the instructions are platform dependent, so instead
-            // make sure we have read fully
-            for (const i of instructions) {
-                expect(i.address).to.have.lengthOf.greaterThan(0);
-                expect(i.instructionBytes).to.have.lengthOf.greaterThan(0);
-                expect(i.instruction).to.have.lengthOf.greaterThan(0);
-            }
-        }
+
+        expectsGeneralDisassemble(disassemble, 100);
+    });
+
+    it('can disassemble with negative offsets', async function () {
+        // In this case we attempt to read from where there is no source,
+        // GDB returns data in a different format in that case
+        const disassemble = (await dc.send('disassemble', {
+            memoryReference: 'main',
+            instructionOffset: -20,
+            instructionCount: 20,
+        } as DebugProtocol.DisassembleArguments)) as DebugProtocol.DisassembleResponse;
+
+        expectsGeneralDisassemble(disassemble, 20, true);
+    });
+
+    it('can disassemble with correct boundries', async function () {
+        const get = (
+            disassemble: DebugProtocol.DisassembleResponse,
+            offset: number
+        ) => {
+            const instruction = disassemble.body?.instructions[offset];
+            expect(instruction).not.eq(undefined);
+            // Instruction undefined already checked.
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+            return instruction!;
+        };
+
+        const expectsInstructionEquals = (
+            instruction1: DebugProtocol.DisassembledInstruction,
+            instruction2: DebugProtocol.DisassembledInstruction,
+            message?: string
+        ) => {
+            expect(instruction1.address).to.eq(instruction2.address, message);
+        };
+
+        // In this case we attempt to read from where there is no source,
+        // GDB returns data in a different format in that case
+        const disassembleLower = (await dc.send('disassemble', {
+            memoryReference: 'main',
+            instructionOffset: -20,
+            instructionCount: 20,
+        } as DebugProtocol.DisassembleArguments)) as DebugProtocol.DisassembleResponse;
+        const disassembleMiddle = (await dc.send('disassemble', {
+            memoryReference: 'main',
+            instructionOffset: -10,
+            instructionCount: 20,
+        } as DebugProtocol.DisassembleArguments)) as DebugProtocol.DisassembleResponse;
+        const disassembleHigher = (await dc.send('disassemble', {
+            memoryReference: 'main',
+            instructionOffset: 0,
+            instructionCount: 20,
+        } as DebugProtocol.DisassembleArguments)) as DebugProtocol.DisassembleResponse;
+
+        expectsGeneralDisassemble(disassembleLower, 20, true);
+        expectsGeneralDisassemble(disassembleMiddle, 20, true);
+        expectsGeneralDisassemble(disassembleHigher, 20, true);
+
+        expectsInstructionEquals(
+            get(disassembleLower, 15),
+            get(disassembleMiddle, 5),
+            'lower[15] should be same with middle[5]'
+        );
+
+        expectsInstructionEquals(
+            get(disassembleMiddle, 15),
+            get(disassembleHigher, 5),
+            'middle[15] should be same with higher[5]'
+        );
     });
 
     it('can handle disassemble at bad address', async function () {
@@ -102,18 +170,23 @@ describe('Disassembly Test Suite', function () {
             memoryReference: '0x0',
             instructionCount: 10,
         })) as DebugProtocol.DisassembleResponse;
+
         expect(disassemble).not.eq(undefined);
         expect(disassemble.body).not.eq(undefined);
         if (disassemble.body) {
             const instructions = disassemble.body.instructions;
             expect(instructions).to.have.lengthOf(10);
-            // the contens of the instructions are platform dependent, so instead
-            // make sure we have read fully
-            for (const i of instructions) {
-                expect(i.address).to.have.lengthOf.greaterThan(0);
-                expect(i.instruction).to.have.lengthOf.greaterThan(0);
-                expect(i.instructionBytes).eq(undefined);
-            }
+            // Checking the invalid instructions content
+            instructions.forEach((inst, ix) => {
+                expect(inst.address).to.eq(
+                    calculateMemoryOffset('0x0', ix * 2)
+                );
+                expect(inst.address).to.have.lengthOf.greaterThan(0);
+                expect(inst.instruction).to.eq(
+                    'failed to retrieve instruction'
+                );
+                expect(inst.presentationHint).to.eq('invalid');
+            });
         }
     });
 });
--- a/src/integration-tests/diassemble.spec.ts
+++ b/src/integration-tests/diassemble.spec.ts
@@ -14,6 +14,7 @@ import { DebugProtocol } from '@vscode/debugprotocol/lib/debugProtocol';
 import { CdtDebugClient } from './debugClient';
 import { fillDefaults, standardBeforeEach, testProgramsDir } from './utils';
 import { calculateMemoryOffset } from '../util/calculateMemoryOffset';
+import { assert } from 'sinon';
 
 describe('Disassembly Test Suite', function () {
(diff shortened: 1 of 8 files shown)


src/integration-tests/util.spec.ts
```diff
@@ -140,3 +147,208 @@ describe('createEnvValues', () => {
         expect(result).to.deep.equals(expectedResult);
     });
 });
+
+describe('calculateMemoryOffset', () => {
```
jonahgraham, line 151: I think when the disassembly view is open at the beginning of the debug session, vscode sometimes sends an effectively invalid command like:

```
From client: disassemble({"memoryReference":"","offset":0,"instructionOffset":-50,"instructionCount":50,"resolveSymbols":true})
```

i.e. no real memoryReference - this code translates it to:

```
GDB command: 11 -data-disassemble -s "(+0)-100" -e "(+0)+0" -- 5
```

the `(+0)` part comes from `calculateMemoryOffset` AFIACT, but we don't have a test for this case.

That said, not sure how we should handle it - perhaps an empty string should be handled by failing the whole command rather than trying something. Especially because vscode is doing caching with the return value on success based on what address it thinks it requested.

src/integration-tests/diassemble.spec.ts
```diff
@@ -82,38 +94,99 @@ describe('Disassembly Test Suite', function () {
...
-                expect(i.instruction).to.have.lengthOf.greaterThan(0);
-            }
-        }
+
+        expectsGeneralDisassemble(disassemble, 100);
+    });
+
+    it('can disassemble with negative offsets', async function () {
+        // In this case we attempt to read from where there is no source,
+        // GDB returns data in a different format in that case
```
jonahgraham, line 103: This comment appears to be copy-paste error? Similar copy-paste below too

src/integration-tests/diassemble.spec.ts
```diff
@@ -82,38 +94,99 @@ describe('Disassembly Test Suite', function () {
...
+        expectsInstructionEquals(
+            get(disassembleLower, 15),
+            get(disassembleMiddle, 5),
+            'lower[15] should be same with middle[5]'
+        );
+
+        expectsInstructionEquals(
+            get(disassembleMiddle, 15),
+            get(disassembleHigher, 5),
+            'middle[15] should be same with higher[5]'
```
jonahgraham, line 164: This test happens to work at the moment - but it can be made to fail with some tweaking/changes to instructions being decoded. This may be more of a problem on x86 with its CISC arch than on ARM with RISC arch.

For example, depending on the start address of what you ask gdb to disassembly you may get different results for disassembly. Here is a simple example in a screenshot.

In the top half is the result of using this implementation where the command passed to gdb was `-data-disassemble -s "(0x0000555555555149)-100" -e "(0x0000555555555149)+0" -- 5` - i.e. start at address `50E5` (last 4 hex digits)

But the bottom half is a CLI command `>disass /mr 0x0000555555555000, 0x000055555555513b` which starts a number of bytes higher and leads to a more likely alignment for real instructions.

As you can see in the red rounded box the instruction address and instruction are the same. But in the green square corner box the misalignment is exposed. The instruction in this case actually started at `50E4`, but because we requested `50E5` the disassembler interpreted the bytes it saw `3d 25 2f 00 00 00` as a different instruction.

image

The general problem is unsolvable with the tools we have now. It is one of the challenges of doing negative offsets from a known instruction address. Unfortunately I don't have a good suggestion for how to fix this now, but I did want you to be aware of it.

PS I don't know if the ARM disassembler will realign if given a non-aligned address to disassemble from. I don't think this is a problem we have to deal with as long as we always are returning valid addresses in `instructionPointerReference`

jonahgraham (changes requested):
Overall the design here looks good - please see individual comments for more details

src/integration-tests/diassemble.spec.ts
asimgunes: Hi @jonahgraham ,

I tried to run some experiments to check the details, like you mentioned above, the starting address is an issue here and according to the start address especially the instructions which are adjacent to the borders may not be correct. Do you think if we provide a larger margin, and dropping the instructions at the beginning and the end of the list could achieve a better result? I am not sure if it can be generalising the behaviour here, but I observe if we skip a couple of instructions, the results seems stable same instruction addresses and instruction bytes.

If we do not have any other approach, possibly, I am going to skip this issue since you also mentioned that we dont have a proper tool to resolve this.

image

src/util/disassembly.ts
```diff
@@ -0,0 +1,201 @@
...
+        }
+    }
+
+    if (length < 0) {
+        // Remove the heading, if necessary
+        if (absLength < list.length) {
+            list.splice(0, list.length - absLength);
+        }
+
+        // Add empty instructions, if necessary
```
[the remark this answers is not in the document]
asimgunes: Noted that, I implemented a similar loop back in the read logic. Planning to push after a some tests.

jonahgraham:
@asimgunes FYI on https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/120#issuecomment-2504354399 - if we can land this soon, I will push the update out in a new vscode version unless you recommend against or we run into any roadblocks.

src/integration-tests/diassemble.spec.ts
asimgunes: Sorry, I removed it.

src/integration-tests/util.spec.ts
```diff
@@ -140,3 +147,208 @@ describe('createEnvValues', () => {
...
+    it('should expect to calculate basic operations', () => {
+        expect(calculateMemoryOffset('0x0000ff00', 2)).to.eq('0x0000ff02');
+        expect(calculateMemoryOffset('0x0000ff00', 8)).to.eq('0x0000ff08');
+        expect(calculateMemoryOffset('0x0000ff00', 64)).to.eq('0x0000ff40');
+        expect(calculateMemoryOffset('0x0000ff00', -2)).to.eq('0x0000fefe');
+        expect(calculateMemoryOffset('0x0000ff00', -8)).to.eq('0x0000fef8');
+        expect(calculateMemoryOffset('0x0000ff00', -64)).to.eq('0x0000fec0');
+    });
+
+    it('should expect to handle 64bit address operations ', () => {
```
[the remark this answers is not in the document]
asimgunes: I added a couple of calculation operation which is handled in 8 bytes long number to which normally not expected to be handled correctly by the `number` data type, requires `bigint`. 

I can update the test, but could you please help me what do you prefer to see in this test?

src/util/disassembly.ts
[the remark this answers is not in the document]
asimgunes: I updated the logic.

src/integration-tests/diassemble.spec.ts
asimgunes: Added the note and skipped the issue for now.

asimgunes:
Hi @jonahgraham , I made some updates related to the feeback, could you please check it when you are suitable?

PS: tests seems failed due to timeout issues, however all passed in my test pr, I am expecting all of the tests will pass if they are triggered again.

src/util/disassembly.ts
```diff
@@ -115,85 +111,86 @@ export const getInstructions = async (
...
-            const startAddress = list[0].address;
-            const filling = getEmptyInstructions(
-                startAddress,
-                absLength - list.length,
-                -2
+    };
+    const remainingLength = () =>
+        Math.sign(length) * Math.max(absLength - list.length, 0);
+    try {
+        // Do we need an explicit guard to prevent infinite loop
```
jonahgraham, line 163: Yes - I think so. I think `sendDataDisassemble` can return a 0 length list. So I would add a break on the loop if result below is 0 length. See the previous loop break early logic here:

https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/3204f5d6ff1a55d8738c525f99e204ce8c819db4/src/gdb/GDBDebugSessionBase.ts#L1356-L1359

jonahgraham (changes requested):

jonahgraham (changes requested):
I was experimenting with the change and I scrolled backwards (lower memory) for a while, then I got to unreadable memory and VSCode won't scroll backwards any further.

I see this in the log:

```
From client: disassemble({"memoryReference":"0x000055555555534e","offset":0,"instructionOffset":-1200,"instructionCount":50,"resolveSymbols":true})
GDB command: 38 -data-disassemble -s "(0x000055555555534e)-4800" -e "(0x000055555555534e)+0" -- 5
GDB result: 38 done,asm_insns=[{address="0x000055555555408e",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x0000555555554090",opcodes="18 03",inst="sbb    %al,(%rbx)"},{address="0x0000555555554092",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x0000555555554094",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x0000555555554096",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x0000555555554098",opcodes="1c 00",inst="sbb    $0x0,%al"},{address="0x000055555555409a",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x000055555555409c",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x000055555555409e",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540a0",opcodes="1c 00",inst="sbb    $0x0,%al"},{address="0x00005555555540a2",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540a4",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540a6",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540a8",opcodes="01 00",inst="
GDB -cont-: 38 add    %eax,(%rax)"},{address="0x00005555555540aa",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540ac",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540ae",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540b0",opcodes="01 00",inst="add    %eax,(%rax)"},{address="0x00005555555540b2",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540b4",opcodes="04 00",inst="add    $0x0,%al"},{address="0x00005555555540b6",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540b8",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540ba",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540bc",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540be",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540c0",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540c2",opcodes="00 00",inst="add    %al,(%rax)"},{address="0x00005555555540c4",opcodes="00 00
-- elided many more GDB -cont-: 38 lines --
GDB -cont-: 38 2eb",func-name="test_arr_md",offset="243",opcodes="8b 45 fc",inst="mov    -0x4(%rbp),%eax"},{address="0x00005555555552ee",func-name="test_arr_md",offset="246",opcodes="48 98",inst="cltq"},{address="0x00005555555552f0",func-name="test_arr_md",offset="248",opcodes="8b 55 f8",inst="mov    -0x8(%rbp),%edx"},{address="0x00005555555552f3",func-name="test_arr_md",offset="251",opcodes="48 63 d2",inst="movslq %edx,%rdx"},{address="0x00005555555552f6",func-name="test_arr_md",offset="254",opcodes="48 c1 e2 05",inst="shl    $0x5,%rdx"},{address="0x00005555555552fa",func-name="test_arr_md",offset="258",opcodes="48 01 d0",inst="add    %rdx,%rax"},{address="0x00005555555552fd",func-name="test_arr_md",offset="261",opcodes="48 8d 14 85 00 00 00 00",inst="lea    0x0(,%rax,4),%rdx"},{address="0x0000555555555305",func-name="test_arr_md",offset="269",opcodes="48 8d 05 34 cd 00 00",inst="lea    0xcd34(%rip),%rax        # 0x555555562040 <arr_md>"},{address="0x000055555555530c",func-name="test_arr_md",offset=
GDB -cont-: 38 "276",opcodes="c7 04 02 00 00 00 00",inst="movl   $0x0,(%rdx,%rax,1)"},{address="0x0000555555555313",func-name="test_arr_md",offset="283",opcodes="83 45 fc 01",inst="addl   $0x1,-0x4(%rbp)"},{address="0x0000555555555317",func-name="test_arr_md",offset="287",opcodes="83 7d fc 1f",inst="cmpl   $0x1f,-0x4(%rbp)"},{address="0x000055555555531b",func-name="test_arr_md",offset="291",opcodes="0f 8e 54 ff ff ff",inst="jle    0x555555555275 <test_arr_md+125>"},{address="0x0000555555555321",func-name="test_arr_md",offset="297",opcodes="83 45 f8 01",inst="addl   $0x1,-0x8(%rbp)"},{address="0x0000555555555325",func-name="test_arr_md",offset="301",opcodes="81 7d f8 ff 03 00 00",inst="cmpl   $0x3ff,-0x8(%rbp)"},{address="0x000055555555532c",func-name="test_arr_md",offset="308",opcodes="0f 8e 37 ff ff ff",inst="jle    0x555555555269 <test_arr_md+113>"},{address="0x0000555555555332",func-name="test_arr_md",offset="314",opcodes="8b 05 08 cd 00 00",inst="mov    0xcd08(%rip),%eax        # 0x555555562040 <
GDB -cont-: 38 arr_md>"},{address="0x0000555555555338",func-name="test_arr_md",offset="320",opcodes="83 c0 01",inst="add    $0x1,%eax"},{address="0x000055555555533b",func-name="test_arr_md",offset="323",opcodes="89 05 ff cc 00 00",inst="mov    %eax,0xccff(%rip)        # 0x555555562040 <arr_md>"},{address="0x0000555555555341",func-name="test_arr_md",offset="329",opcodes="e9 17 ff ff ff",inst="jmp    0x55555555525d <test_arr_md+101>"},{address="0x0000555555555346",func-name="main",offset="0",opcodes="f3 0f 1e fa",inst="endbr64"},{address="0x000055555555534a",func-name="main",offset="4",opcodes="55",inst="push   %rbp"},{address="0x000055555555534b",func-name="main",offset="5",opcodes="48 89 e5",inst="mov    %rsp,%rbp"}]
To client: {"seq":0,"type":"response","request_seq":32,"command":"disassemble","success":true,"body":{"instructions":[{"address":"0x000055555555483f","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554841","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554843","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554845","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554847","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554849","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x000055555555484b","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x000055555555484d","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x000055555555484f","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554851","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554853","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554855","instructionBytes":"00 00","instruction":"add    %al,(%rax)","location":{}},{"address":"0x0000555555554857","instructionBytes":"00 00","instruction":"add    %al,([...]
From client: disassemble({"memoryReference":"0x000055555555534e","offset":0,"instructionOffset":-1250,"instructionCount":50,"resolveSymbols":true})
GDB command: 39 -data-disassemble -s "(0x000055555555534e)-5000" -e "(0x000055555555534e)+0" -- 5
GDB result: 39 error,msg="Cannot access memory at address 0x555555553fc6"
To client: {"seq":0,"type":"response","request_seq":33,"command":"disassemble","success":true,"body":{"instructions":[]}}
From client: disassemble({"memoryReference":"0x000055555555534e","offset":0,"instructionOffset":-1250,"instructionCount":50,"resolveSymbols":true})
GDB command: 40 -data-disassemble -s "(0x000055555555534e)-5000" -e "(0x000055555555534e)+0" -- 5
GDB result: 40 error,msg="Cannot access memory at address 0x555555553fc6"
To client: {"seq":0,"type":"response","request_seq":34,"command":"disassemble","success":true,"body":{"instructions":[]}}
```

In here I see two issues:

1. The `-data-disassemble` command keeps getting longer, it goes from ref - 4800 to ref - 0, even though `instructionCount` is 50.
2. On failing to read memory, the reply to the client contains `{"instructions":[]}` - i.e. no instructions, rather than a bunch of invalid instructions. You have a comment `Fill with empty instructions in case of memory error.` so it looks like there is a special case not being handled here.

asimgunes:
Hi @jonahgraham ,

> In here I see two issues:
> 
> 1. The `-data-disassemble` command keeps getting longer, it goes from ref - 4800 to ref - 0, even though `instructionCount` is 50.

This is related to instruction conversion, the request came as: 

disassemble({"memoryReference":"0x000055555555534e","offset":0,"instructionOffset":-1200,"instructionCount":50,"resolveSymbols":true})

We need to have all the instructions until the memoryReference point to calculate the instructionOffset and instructionCount. Here you can say we need smaller chunks to query if thats the case, I might need to re-visit the implementation.

> 2. On failing to read memory, the reply to the client contains `{"instructions":[]}` - i.e. no instructions, rather than a bunch of invalid instructions. You have a comment `Fill with empty instructions in case of memory error.` so it looks like there is a special case not being handled here.

It shouldn't happen, let me check whats missing here.

jonahgraham:
> This is related to instruction conversion, the request came as:
> 
> disassemble({"memoryReference":"0x000055555555534e","offset":0,"instructionOffset":-1200,"instructionCount":50,"resolveSymbols":true})
> 
> We need to have all the instructions until the memoryReference point to calculate the instructionOffset and instructionCount. Here you can say we need smaller chunks to query if thats the case, I might need to re-visit the implementation.

Thanks for explaining it - the distinction between byte offset and instruction offset will probably keep tripping me (and future developers) up, but your implementation looks correct. I don't think we should bother with chunking unless/until we see some performance hit of doing these large reads.

jonahgraham:
> It shouldn't happen, let me check whats missing here.

:+1:

asimgunes:
Hi @jonahgraham ,

I updated the logic and sent as a separate commit so you can easily track the changes. You can find the change related to the memory read and empty instruction filling in the last commit. Could you please review this update when you are available? 

Additionally, it seems the Linux tests are broken but I believe this is a general issue rather than related to this update. I am leaving the tests to you but please let me know if it is not a general error and related to this update.

**More details about this update:**

You can spot a different approach than we discussed previously, so let me explain the new approach and its reason here: 

When we read the memory,  and go to the lower memory values, at some point, we came accross the invalid memory areas where we cannot disassemble the intructions. at this point if we return empty instructions, this will override previously read correct instructions. Because of this purpose, I returned back the error instead of correct number of the empty instructions. With this behaviour VSCode stops the scrolling and shows the error without killing the debug session.

I believe this is a better approach, otherwise I observed suddenly the disassembled instructions turned to empty instructions during my test session.

On the other hand, approach still contains edge cases. Since there is a conversion between the instruction offsets and memory offsets via a lineer conversion logic, in most of the times we are fetching more than we need. This is causing to hit the memory read error much more earlier than it needs. For example, DAP request came for instructionOffset: -1000, however we read memoryOffset: -4000 which contains much more instructions, for instance we read ~1500 instructions. When DAP request came for instructionOffset: -1200 we encounter memory read error while reading memoryOffset: -4800, however, as you can imagine, if we can handle the boundries better, we can still return back the 1200 instructions. 

I tried some additional approaches but every different approach that I tried came with other issues. So, I left the implementation as simple as possible. Since our current appoach is a "Greedy" approach we are not keeping track of the areas which are disassembled before, in the future we may think about changing our approach with a "Dynamic Programming" approach where backtracking the successfully disassembled range and split/chunk the memory with a better approach and do not override/erase the already diassembled instructions with empty ones.

jonahgraham:
> Additionally, it seems the Linux tests are broken

They are because GHA changed version of Ubuntu we are using. See #345 which is close to ready, but not quite there yet.

I'll review the code change once GHA issue is resolved - your description sounds good to me, but I want to run it up to see before I approve it.

jonahgraham (approved):
I'm going to approve this now - I think there are still some corner cases, but seeing as many of them are probably related to vscode and the design to use negative offsets in this way,

Three particular issues I see that I think all need to be solved in vscode side, so if there is interest, someone needs to invest there.

- vscode often seems to stop fetching instructions when scrolling, it gets worse when the adapter sends an (legitimate) error, or when source code is interleaved
- refreshing of the view is inconsistent. I can really see it when toggling source code in the disassembly
- the way vscode handles negative offsets is really frustrating - I think vscode should make subsequent negative offsets using the memory reference that the adapter returned, rather than the current IP.

I have merged the test changes from #345 and if I get green tick I will merge and we can discuss if anything else is wanted in future versions.

jonahgraham:
I published this change as 1.0.3 on npm

PR #343, opened 2024-12-04
Array optimisation for variablesRequest

asimgunes:
Hi @jonahgraham ,

As we discussed the issue previously, this is an update to optimising the array read operation in the `variablesRequest`. 

Recently, there are some redundant operations performed during the `variablesRequest` in the adapter, causing a performance issue while retreiving larger arrays. The operation, which is removed in this update, creates O(n) compexity in the read operation in terms of gdb communication. With this update, by removing this redundant logic, gdb communication complexity is going to decrease to O(1). 

Update tested manually in 1-dimensional int array, 2-dimensional int array, 1 dimensional struct array and 1-dimensional struct contains 1-dimensional array element. We observe the increase in the performance without having issue on the read operation. 

Could you please review this update when you are suitable?

**PS:** Besides the removal of the array logic, the `this.getFullPathExpression` is also replaced since still causing an expensive gdb request which affects the performance.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2008,56 +2008,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     // update the display name for array elements to have square brackets
                     name = `[${child.exp}]`;
                 }
-                if (isArrayParent || isArrayChild) {
-                    // can't use a relative varname (eg. var1.a.b.c) to create/update a new var so fetch and track these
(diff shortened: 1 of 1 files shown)


jonahgraham (approved):

jonahgraham:
@asimgunes do you want a release built with this improvement, or do you want to wait for #341 to be completed too?

asimgunes:
@jonahgraham , I made an update to #341. If we are going to merge it soon, we can have a release build together.

jonahgraham:
v1.0.2 has been published that contains this fix.

PR #345, opened 2024-12-10
Update versions of node and workflow items in GitHub Actions

jonahgraham:
This change also includes some test changes that are as a result of the changes in versions of tools.


jonahgraham:
Parts of this update is needed because of https://github.com/actions/runner-images/issues/10636 - the main branch is failing because of running build and test on latest ubuntu.

PR #348, opened 2025-01-22
Disassembly address handling improvement

asimgunes:
Hi @jonahgraham ,

We observe some issues in while using the disassembly in embedded devices due to sending disassemble request for negative memory regions. We implemented a code-update to limit the lower bound of the query to the address 0x0. 

If we still need to fill more instructions depending on the DAP request, you may realise that we used relative addresses in the empty instructions (e.g. (0x00)-2, (0x00)-4 etc). This is arguable and I am happy to hear your thoughts. Simply the VSCode is ignoring these lines and not showing in the disassembly window (maybe because they are not convertable to number or bigint).

It was hard to check this behaviour in unit test, thus I implemented a test for the internal function and checked the behaviour over that internal function. 

Could you please review the update when you are suitable? I am happy to hear your feedback.

Kind regards

Asim

### DIFF
--- a/src/integration-tests/util.spec.ts
+++ b/src/integration-tests/util.spec.ts
@@ -14,11 +14,16 @@ import { expect } from 'chai';
 import * as os from 'os';
 import { calculateMemoryOffset } from '../util/calculateMemoryOffset';
 import { MIDataDisassembleAsmInsn } from '../mi';
+import * as midata from '../mi/data';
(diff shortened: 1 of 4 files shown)


jonahgraham (approved):
LGTM once CI approves I can merge and tag.

jonahgraham:
@asimgunes published to npm. Please let me know if anything looks amiss.

PR #350, opened 2025-02-03
Hardware/Software Breakpoint Modes (Breakpoint Types)

asimgunes:
Hi @jonahgraham ,

We like to send an update for new DAP feature "Breakpoint Modes". Here in this update I added hardware and software breakpoint types where the debug adapter sends this capability to the IDE, and uses this information during the breakpoint operations.

I would appreciate it if you can check the pull request when you are suitable. 

By the way, there could be some arguable decisions here, let me emphasis the important points:

1. Enabling the breakpointModes as Hardware and Software:

I implemented the idea over 'hardware' and 'software' modes. By default, I enabled both of the modes, however I implemented a separate function `getBreakpointModes` to let user to override the function if they needed. Thus, user can override the behaviour and return any desired combination.

In this new implementation, I priotize the `mode` information over `hardware` flag, in other words, if `mode` is defined I considered the `mode` parameter, otherwise I checked the `hardware` flag.

2. `hardware` flag became redundant after having the `mode` property:

Previously we have a `hardware` flag on `MIBreakpointInsertOptions` class, this flag became redundant after the having the `mode` property. I still left the `hardware` flag for backward compatibility. Moreover, the `hardware` controlled by launch arguments to force debug adapter to send the breakpoints as hardware breakpoints. Maybe we can consider to introduce a `defaultBreakpointMode` parameter and remove the `hardwareBreakpoint` parameter. I didnot make a change to avoid breaking the current users.

3. Windows tests:

I couldn't managed to run the hardware breakpoint tests in the Windows, conditions like isRemoteTest or !isRemoteTest seems failing too, please let me know, if there is a better Windows test conditions.

Kind regards,

Asim

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -195,6 +195,21 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         }
     }
 
+    protected getBreakpointModes(): DebugProtocol.BreakpointMode[] | undefined {
+        return [
+            {
+                label: 'Hardware Breakpoint',
+                mode: 'hardware',
+                appliesTo: ['source'],
+            },
+            {
+                label: 'Software Breakpoint',
+                mode: 'software',
+                appliesTo: ['source'],
+            },
+        ];
+    }
+
     protected initializeRequest(
         response: DebugProtocol.InitializeResponse,
         args: DebugProtocol.InitializeRequestArguments
@@ -216,6 +231,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         response.body.supportsReadMemoryRequest = true;
         response.body.supportsWriteMemoryRequest = true;
         response.body.supportsSteppingGranularity = true;
+        response.body.breakpointModes = this.getBreakpointModes();
         this.sendResponse(response);
     }
 
@@ -438,11 +454,19 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     const vsbpCond = vsbp.condition || undefined;
                     const gdbbpCond = gdbbp.cond || undefined;
 
+                    const vsbpIsBreakpointTypeHardware = vsbp.mode
+                        ? vsbp.mode === 'hardware'
+                        : this.gdb.isUseHWBreakpoint();
+                    const gdbbpIsBreakpointTypeHardware =
+                        gdbbp.type === 'hw breakpoint';
+
                     // Check with original-location so that relocated breakpoints are properly matched
                     const gdbOriginalLocation = `${gdbOriginalLocationPrefix}${vsbp.line}`;
                     return !!(
                         gdbbp['original-location'] === gdbOriginalLocation &&
-                        vsbpCond === gdbbpCond
+                        vsbpCond === gdbbpCond &&
+                        vsbpIsBreakpointTypeHardware ===
+                            gdbbpIsBreakpointTypeHardware
                     );
                 }
             );
@@ -516,6 +540,9 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
 
                 try {
                     const line = vsbp.line.toString();
+                    const breakpointMode = vsbp.mode as
+                        | mi.MIBreakpointMode
+                        | undefined;
                     const options = await this.gdb.getBreakpointOptions(
                         {
                             locationType: 'source',
@@ -526,7 +553,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                             condition: vsbp.condition,
                             temporary,
                             ignoreCount,
-                            hardware: this.gdb.isUseHWBreakpoint(),
+                            mode: breakpointMode,
+                            hardware: breakpointMode
+                                ? breakpointMode === 'hardware'
+                                : this.gdb.isUseHWBreakpoint(),
                         }
                     );
                     const gdbbp = await mi.sendSourceBreakpointInsert(
--- a/src/integration-tests/breakpointModes.spec.ts
+++ b/src/integration-tests/breakpointModes.spec.ts
@@ -1,0 +1,272 @@
+/*********************************************************************
+ * Copyright (c) 2025 Renesas Electronics Corporation and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *********************************************************************/
+import * as path from 'path';
+import * as os from 'os';
+import { expect } from 'chai';
+import { CdtDebugClient } from './debugClient';
+import { standardBeforeEach, testProgramsDir, fillDefaults } from './utils';
+import { DebugProtocol } from '@vscode/debugprotocol';
+import { MIBreakpointMode } from '../mi';
+
+// This mock adapter is overriding the getBreakpointOptions method.
+const adapter =
+    'integration-tests/mocks/debugAdapters/dynamicBreakpointOptions.js';
+const argHardwareBreakpointTrue = '--hardware-breakpoint-true';
+const argHardwareBreakpointFalse = '--hardware-breakpoint-false';
+const argBreakpointModeHardware = ['--breakpoint-mode', 'hardware'];
+const argBreakpointModeSoftware = ['--breakpoint-mode', 'software'];
+
+const program = 'count';
+
+const startDebugClientWithArgs = async (
+    test: Mocha.Test | undefined,
+    ...args: string[]
+) => {
+    const dc = await standardBeforeEach(adapter, args);
+    await dc.launchRequest(
+        fillDefaults(test, {
+            program: path.join(testProgramsDir, 'count'),
+        })
+    );
+    return dc;
+};
+
+const sendBreakpointRequest = (
+    dc: CdtDebugClient,
+    options?: Partial<DebugProtocol.SourceBreakpoint>
+) => {
+    return dc.setBreakpointsRequest({
+        source: {
+            name: `${program}.c`,
+            path: path.join(testProgramsDir, `${program}.c`),
+        },
+        breakpoints: [
+            {
+                column: 1,
+                line: 4,
+                ...(options ?? {}),
+            },
+        ],
+    });
+};
+
+const expectBreakpoint = async (
+    dc: CdtDebugClient,
+    response: DebugProtocol.SetBreakpointsResponse,
+    breakpointMode: MIBreakpointMode
+) => {
+    expect(response.body.breakpoints.length).eq(1);
+    expect(response.body.breakpoints[0].verified).eq(true);
+    expect(response.body.breakpoints[0].message).eq(undefined);
+    await dc.configurationDoneRequest();
+
+    let isCorrect;
+    let outputs;
+    while (!isCorrect) {
+        // Cover the case of getting event in Linux environment.
+        // If cannot get correct event, program timeout and test case failed.
+        outputs = await dc.waitForEvent('output');
+        isCorrect = outputs.body.output.includes('breakpoint-modified');
+    }
+    expect(outputs?.body.output).includes(
+        `type="${
+            breakpointMode === 'hardware' ? 'hw breakpoint' : 'breakpoint'
+        }"`
+    );
+    const stoppedOutput = await dc.waitForEvent('stopped');
+    expect(stoppedOutput.body?.reason).eq('breakpoint');
+};
+
+// Hardware breakpoint tests are not working, so skipped for Windows.
+// For further information, please check the discussion here:
+// https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/350
+const skipHardwareBreakpointTest = os.platform() === 'win32';
+
+describe('breakpoint mode', async () => {
+    let dc: CdtDebugClient;
+
+    afterEach(async () => {
+        await dc.stop();
+    });
+
+    describe('with no override', async () => {
+        beforeEach(async function () {
+            dc = await startDebugClientWithArgs(this.currentTest);
+        });
+
+        it('insert breakpoint as hardware breakpoint', async function () {
+            if (skipHardwareBreakpointTest) {
+                this.skip();
+            }
+
+            const response = await sendBreakpointRequest(dc, {
+                mode: 'hardware',
+            });
+
+            await expectBreakpoint(dc, response, 'hardware');
+        });
+
+        it('insert breakpoint as software breakpoint', async function () {
+            const response = await sendBreakpointRequest(dc, {
+                mode: 'software',
+            });
+
+            await expectBreakpoint(dc, response, 'software');
+        });
+    });
+
+    describe('with options hardware is true', async () => {
+        beforeEach(async function () {
+            dc = await startDebugClientWithArgs(
+                this.currentTest,
+                argHardwareBreakpointTrue
+            );
+        });
+
+        it('when no mode - insert breakpoint as hardware breakpoint', async function () {
+            if (skipHardwareBreakpointTest) {
+                this.skip();
+            }
+
+            const response = await sendBreakpointRequest(dc);
+
+            await expectBreakpoint(dc, response, 'hardware');
+        });
+
+        it('when mode is hardware - insert breakpoint as hardware breakpoint', async function () {
+            if (skipHardwareBreakpointTest) {
+                this.skip();
+            }
+
+            const response = await sendBreakpointRequest(dc, {
+                mode: 'hardware',
+            });
+
+            await expectBreakpoint(dc, response, 'hardware');
+        });
+
+        it('when mode is software - insert breakpoint as software breakpoint', async function () {
+            const response = await sendBreakpointRequest(dc, {
+                mode: 'software',
+            });
+
+            await expectBreakpoint(dc, response, 'software');
+        });
+    });
+
+    describe('with options hardware is false', async () => {
+        beforeEach(async function () {
+            dc = await startDebugClientWithArgs(
+                this.currentTest,
+                argHardwareBreakpointFalse
+            );
+        });
+
+        it('when no mode - insert breakpoint as software breakpoint', async function () {
+            const response = await sendBreakpointRequest(dc);
+
+            await expectBreakpoint(dc, response, 'software');
+        });
+
+        it('when mode is hardware - insert breakpoint as hardware breakpoint', async function () {
+            if (skipHardwareBreakpointTest) {
+                this.skip();
+            }
+
+            const response = await sendBreakpointRequest(dc, {
+                mode: 'hardware',
+            });
+
+            await expectBreakpoint(dc, response, 'hardware');
+        });
+
+        it('when mode is software - insert breakpoint as software breakpoint', async function () {
+            const response = await sendBreakpointRequest(dc, {
+                mode: 'software',
+            });
+
+            await expectBreakpoint(dc, response, 'software');
+        });
+    });
+
+    describe('with options mode overriden to hardware', async () => {
+        beforeEach(async function () {
+            dc = await startDebugClientWithArgs(
+                this.currentTest,
+                ...argBreakpointModeHardware
+            );
+        });
+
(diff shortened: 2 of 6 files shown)


jonahgraham (commented):
> I couldn't managed to run the hardware breakpoint tests in the Windows, conditions like isRemoteTest or !isRemoteTest seems failing too, please let me know, if there is a better Windows test conditions.

I don't know what support Windows native has for hardware breakpoints. If it does support it I am sure we can find a way to enable the tests. What happens if do do "hbreak main" in a manually launched GDB?

For example, here is a simple test on Linux to compare to windows:

```
$ gdb ./empty1.exe 
GNU gdb (Ubuntu 15.0.50.20240403-0ubuntu1) 15.0.50.20240403-git
(gdb) hbreak main
No hardware breakpoint support in the target.
(gdb) break main
Breakpoint 1 at 0x115e: file ./empty1.c, line 17.
(gdb) run
Starting program: /scratch/debug/git/cdt-amalgamator/sampleWorkspace/empty1.exe 
Breakpoint 1, main () at ./empty1.c:17
17          int local_in_main = function1();    
(gdb) hbreak main
Note: breakpoint 1 also set at pc 0x55555555515e.
Hardware assisted breakpoint 2 at 0x55555555515e: file ./empty1.c, line 17.
(gdb) 
```

As shown above, hbreak may not be available before details of the target are known, which happens here when I run.

jonahgraham:
PS Please include a more complete commit message in your next commit so that suitable history can end up in git.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -438,11 +454,21 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     const vsbpCond = vsbp.condition || undefined;
                     const gdbbpCond = gdbbp.cond || undefined;
 
+                    // If mode is not defined, ignore it
+                    const vsbpIsBreakpointTypeHardware = vsbp.mode
+                        ? vsbp.mode === 'hardware'
+                        : undefined;
```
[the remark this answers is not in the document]
asimgunes: Yes, it is more reasonable. I updated it.

src/mi/breakpoint.ts
```diff
@@ -49,8 +49,11 @@ export interface MIBreakListResponse extends MIResponse {
     };
 }
 
+export type MIBreakpointMode = 'hardware' | 'software';
+
 export interface MIBreakpointInsertOptions {
     temporary?: boolean;
+    mode?: MIBreakpointMode;
```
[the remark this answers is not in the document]
asimgunes: I added some describtion to the fields. I hope they are clear.

asimgunes:
> I don't know what support Windows native has for hardware breakpoints. If it does support it I am sure we can find a way to enable the tests. What happens if do do "hbreak main" in a manually launched GDB?

I have a couple of findings.

Firstly, I observe that hardware breakpoints only available in remote debugging in Windows, as you can find gdb logs here. Additionally, there is a stop condition issue in gdb async off, so, I am planning to use the skip condition as `skipHardwareBreakpointTest = os.platform() === 'win32' && (!isRemoteTest || !gdbAsync)`.

Local debugging with gdb
```
> gdb count.exe
GNU gdb (GDB) 15.1
...
(gdb) hbreak main
No hardware breakpoint support in the target.
(gdb) break main
Breakpoint 1 at 0x40146e: file count.c, line 2.
```

Remote debugging using gdb+gdbserver:

```
> gdb
GNU gdb (GDB) 15.1
...
(gdb) target remote localhost:2345
Remote debugging using localhost:2345
...
(gdb) hbreak main
Hardware assisted breakpoint 1 at 0x40146e: file count.c, line 2.
```

Secondly, it seems the `while` loop in the test (which I copied from 'breakpoints.spec.ts') is causing timeout. In Gitlab tests, we are not receiving the 'breakpoint-modified' output for 'hardware' breakpoints, on the other hand, I am not receiving the output for both 'hardware'&'software' breakpoints in my local environment (local environment uses GDB 15.1, github tests uses GDB 11). 

```
let outputs;
while (!isCorrect) {
	// Cover the case of getting event in Linux environment.
	// If cannot get correct event, program timeout and test case failed.
	outputs = await dc.waitForEvent('output');
	isCorrect = outputs.body.output.includes('breakpoint-modified');
}
```

asimgunes:
Hi @jonahgraham ,

I updated the pull-request, could you please review when you are suitable?

Kind regards.

asimgunes:
> PS Please include a more complete commit message in your next commit so that suitable history can end up in git.

Tried to made a more detailled commit message in recent update.

src/mi/breakpoint.ts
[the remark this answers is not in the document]
jonahgraham: It looks good, just a couple of small changes. See suggestions.

src/mi/breakpoint.ts
```diff
@@ -49,8 +49,25 @@ export interface MIBreakListResponse extends MIResponse {
     };
 }
 
+export type MIBreakpointMode = 'hardware' | 'software';
+
 export interface MIBreakpointInsertOptions {
     temporary?: boolean;
+
+    /**
+     * The `mode` property is priotized over the `hardware` property. If `mode` is defined, then the information in the `hardware` flag is ignored during the insert breakpoint operation.
```
jonahgraham, line 58: prettier doesn't wrap/format comments, so my suggestion is just to wrap this very long line.

```suggestion
     * The `mode` property is priotized over the `hardware` property.
     * If `mode` is defined, then the information in the `hardware` flag is
     * ignored during the insert breakpoint operation.
```

src/mi/breakpoint.ts
```diff
@@ -49,8 +49,25 @@ export interface MIBreakListResponse extends MIResponse {
...
+     *
+     * The value of the mode wil be:
+     *
+     * - `'hardware'`: If user explicitly selects the breakpoint mode as 'Hardware Breakpoint' at the user interface.
+     * - `'software'`: If user explicitly selects the breakpoint mode as 'Software Breakpoint' at the user interface.
+     * - `undefined`: If user didn't make an explicitly breakpoint mode selection.
```
jonahgraham, line 64:
```suggestion
     * - `undefined`: If user didn't make an explicitly breakpoint mode selection, in this case the `hardware` flag will be used.
```

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -195,6 +195,21 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         }
     }
 
+    protected getBreakpointModes(): DebugProtocol.BreakpointMode[] | undefined {
+        return [
+            {
+                label: 'Hardware Breakpoint',
+                mode: 'hardware',
+                appliesTo: ['source'],
```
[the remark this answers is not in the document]
jonahgraham: > I also like to consider handling the instruction breakpoints in a separate conversation.

+1

jonahgraham:
> > PS Please include a more complete commit message in your next commit so that suitable history can end up in git.
> 
> Tried to made a more detailled commit message in recent update.

Thank you.

jonahgraham:
> Firstly, I observe that hardware breakpoints only available in remote debugging in Windows, as you can find gdb logs here. Additionally, there is a stop condition issue in gdb async off, so, I am planning to use the skip condition as `skipHardwareBreakpointTest = os.platform() === 'win32' && (!isRemoteTest || !gdbAsync)`.

That makes sense - async is better all the time anyway. 

---

> Secondly, it seems the `while` loop in the test [...]

Is the breakpoint modified arriving *before* the `waitForEvent` (you should see it in the log file)? If so, this issue is quite prevalent in the adapter tests. I have fixed them in some places but the basic solution is you have to issue the `waitForEvent` *before* the operation that may trigger the event. Here is a simple example https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/c8015bc29ee5aea8432d7c186b49d5336c20aeb8/src/integration-tests/pause.spec.ts#L43-L46

If that isn't the cause, let me know.

asimgunes:
> Is the breakpoint modified arriving _before_ the `waitForEvent` (you should see it in the log file)? If so, this issue is quite prevalent in the adapter tests. I have fixed them in some places but the basic solution is you have to issue the `waitForEvent` _before_ the operation that may trigger the event. Here is a simple example

I think the issue might be different. When a "-break-insert" command send in Windows (for software breakpoint), after the configuration you can see in the following log that we are receiving a `breakpoint-modified` event. 

```
[16:31:14.859 UTC] GDB command: 6 -break-insert --source "D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c" --line 4
[16:31:14.875 UTC] GDB result: 6 done,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0x00007ff7da50146b",func="main",file="count.c",fullname="D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c",line="4",thread-groups=["i1"],times="0",original-location="-source D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c -line 4"}
[16:31:14.875 UTC] To client: {"seq":0,"type":"response","request_seq":3,"command":"setBreakpoints","success":true,"body":{"breakpoints":[{"id":1,"line":4,"verified":true}]}}
[16:31:14.876 UTC] From client: configurationDone(undefined)
[16:31:14.876 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"console","output":"\nIn the Debug Console view you can interact directly with GDB.\nTo display the value of an expression, type that expression which can reference\nvariables that are in scope. For example type '2 + 3' or the name of a variable.\nArbitrary commands can be sent to GDB by prefixing the input with a '>',\nfor example type '>show version' or '>help'.\n\n"}}
[16:31:14.876 UTC] GDB command: 7 -exec-continue
[16:31:14.889 UTC] GDB result: 7 running
[16:31:14.889 UTC] GDB exec async: running,thread-id="all"
[16:31:14.890 UTC] To client: {"seq":0,"type":"event","event":"continued","body":{"threadId":1,"allThreadsContinued":true}}
[16:31:14.890 UTC] To client: {"seq":0,"type":"response","request_seq":4,"command":"configurationDone","success":true}
[16:31:14.892 UTC] GDB notify async: breakpoint-modified,bkpt={number="1",type="breakpoint",disp="keep",enabled="y",addr="0x00007ff7da50146b",func="main",file="count.c",fullname="D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c",line="4",thread-groups=["i1"],times="1",original-location="-source D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c -line 4"}
[16:31:14.893 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"\n"}}
[16:31:14.893 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"Thread 1 hit Breakpoint 1, main () at count.c:4\n"}}
[16:31:14.893 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"4\t        count ++; \n"}}
[16:31:14.893 UTC] GDB exec async: stopped,reason="breakpoint-hit",disp="keep",bkptno="1",frame={addr="0x00007ff7da50146b",func="main",args=[],file="count.c",fullname="D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c",line="4",arch="i386:x86-64"},thread-id="1",stopped-threads="all"
[16:31:14.894 UTC] To client: {"seq":0,"type":"event","event":"stopped","body":{"reason":"breakpoint","allThreadsStopped":true,"threadId":1}}
[16:31:14.894 UTC] From client: disconnect(undefined)
[16:31:14.894 UTC] GDB command: 8 disconnect
``` 

On the other hand, when the similar code run for inserting a hardware breakpoint, there is no "breakpoint-modified" event found at the output. Recently I saw that the reason of the "stopped" event is different between the "software" and "hardware" breakpoint samples. Is that meaningful to you? 

```
[16:31:14.043 UTC] GDB command: 6 -break-insert -h --source "D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c" --line 4
[16:31:14.059 UTC] GDB result: 6 done,bkpt={number="1",type="hw breakpoint",disp="keep",enabled="y",addr="0x00007ff7da50146b",func="main",file="count.c",fullname="D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c",line="4",thread-groups=["i1"],times="0",original-location="-source D:\\a\\cdt-gdb-adapter\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\count.c -line 4"}
[16:31:14.059 UTC] To client: {"seq":0,"type":"response","request_seq":3,"command":"setBreakpoints","success":true,"body":{"breakpoints":[{"id":1,"line":4,"verified":true}]}}
[16:31:14.060 UTC] From client: configurationDone(undefined)
[16:31:14.060 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"console","output":"\nIn the Debug Console view you can interact directly with GDB.\nTo display the value of an expression, type that expression which can reference\nvariables that are in scope. For example type '2 + 3' or the name of a variable.\nArbitrary commands can be sent to GDB by prefixing the input with a '>',\nfor example type '>show version' or '>help'.\n\n"}}
[16:31:14.060 UTC] GDB command: 7 -exec-continue
[16:31:14.073 UTC] GDB result: 7 running
[16:31:14.073 UTC] GDB exec async: running,thread-id="all"
[16:31:14.074 UTC] To client: {"seq":0,"type":"event","event":"continued","body":{"threadId":1,"allThreadsContinued":true}}
[16:31:14.074 UTC] To client: {"seq":0,"type":"response","request_seq":4,"command":"configurationDone","success":true}
[16:31:14.075 UTC] From client: disconnect(undefined)
[16:31:14.075 UTC] GDB command: 8 -exec-interrupt --all
[16:31:14.089 UTC] GDB result: 8 done
[16:31:14.324 UTC] GDB notify async: thread-created,id="3",group-id="i1"
[16:31:14.324 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"[New Thread 2984]\n"}}
[16:31:14.324 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"\nThread "}}
[16:31:14.325 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"3 received signal SIGINT, Interrupt.\n"}}
[16:31:14.325 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"[Switching to Thread 2984]\n"}}
[16:31:14.325 UTC] To client: {"seq":0,"type":"event","event":"output","body":{"category":"stdout","output":"0x00007ffd731f4c8b in KERNELBASE!CtrlRoutine () from target:C:\\Windows\\System32\\KernelBase.dll\n"}}
[16:31:14.325 UTC] GDB exec async: stopped,reason="signal-received",signal-name="SIGINT",signal-meaning="Interrupt",frame={addr="0x00007ffd731f4c8b",func="KERNELBASE!CtrlRoutine",args=[],from="target:C:\\Windows\\System32\\KernelBase.dll",arch="i386:x86-64"},thread-id="3",stopped-threads="all"
[16:31:14.326 UTC] GDB command: 9 disconnect
```

On last detail: As far as I searched from the Windows test logs (the gdb test logs downloaded from github pipeline), there is no previously implemented 'hardware' breakpoint test running for Windows. So I believe, this is the reason why we didnot observe this issue in the current breakpoint tests. (I simply searched for the following regex: `-break-insert(.*)-h`)

src/mi/breakpoint.ts
asimgunes: Update the comments.

jonahgraham:
@asimgunes yesterday I looked at this and couldn't figure out what was happening at all. Today with a clear head I can see in the hardware case that the breakpoint simply isn't working, the execution is never hitting the breakpoint.

The breakpoint modified is sent in the software case because the breakpoint is being hit and the modification is that `times` is incremented.

Maybe we can discuss what to do on the call today?

jonahgraham:
During our call today we decided to defer hardware breakpoint support on Windows host development. Therefore the issues discussed in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/350#issuecomment-2636935359  are now resolved. See the follow-up https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/351

src/integration-tests/breakpointModes.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+    expect(outputs?.body.output).includes(
+        `type="${
+            breakpointMode === 'hardware' ? 'hw breakpoint' : 'breakpoint'
+        }"`
+    );
+};
+
+// Only do the hardware breakpoint tests in Windows in remote debugging with gdbAsync on
+const skipHardwareBreakpointTest =
+    os.platform() === 'win32' && (!isRemoteTest || !gdbAsync);
```
jonahgraham, line 97: Lets make this test simply skip on win32 until #351 is addressed. See https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/350#issuecomment-2640501410 also

```suggestion
    os.platform() === 'win32';
```

src/integration-tests/breakpointModes.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+    breakpointMode: MIBreakpointMode
+) => {
+    expect(response.body.breakpoints.length).eq(1);
+    expect(response.body.breakpoints[0].verified).eq(true);
+    expect(response.body.breakpoints[0].message).eq(undefined);
+    await dc.configurationDoneRequest();
+    // 'breakpoint-modified' output doesn't received in hardware breakpoints in Windows
+    if (os.platform() === 'win32' && breakpointMode === 'hardware') {
+        return;
+    }
```
jonahgraham, line 78: With the suggestion at line 97, I think this can be removed?

```suggestion
```

src/integration-tests/breakpointModes.spec.ts
```diff
@@ -0,0 +1,279 @@
...
+        // If cannot get correct event, program timeout and test case failed.
+        outputs = await dc.waitForEvent('output');
+        isCorrect = outputs.body.output.includes('breakpoint-modified');
+    }
+    expect(outputs?.body.output).includes(
+        `type="${
+            breakpointMode === 'hardware' ? 'hw breakpoint' : 'breakpoint'
+        }"`
+    );
+};
```
jonahgraham, line 93: Add a check here that we actually stop at the given breakpoint. See for example https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/ffeb53a09495ba20968179303ecb35168415cd1f/src/integration-tests/breakpoints.spec.ts#L174-L179

The checking for breakpoint modified is a proxy test for the breakpoint being hit as when a breakpoint is hit the `times` field is updated. But using the snippet above allows us to check that the breakpoint is actually hit.

Note, not every test needs to make sure breakpoint is actually hit - for example https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/ffeb53a09495ba20968179303ecb35168415cd1f/src/integration-tests/breakpoints.spec.ts#L218 just checks breakpoint properties. But at least some of breakpoints.spec.ts do check a breakpoint is hit.

src/integration-tests/breakpointModes.spec.ts
asimgunes: Updated.

src/integration-tests/breakpointModes.spec.ts
asimgunes: Updated.

src/integration-tests/breakpointModes.spec.ts
asimgunes: Thanks for the suggestion, I added the `dc.waitForEvent('stopped')` logic and also checked the result whether it is stopped because of the `breakpoint` or not. I believe test is more comprehensive now.

jonahgraham (approved):

jonahgraham:
This change is included in v1.0.5 on npm: https://www.npmjs.com/package/cdt-gdb-adapter/v/1.0.5

PR #357, opened 2025-04-16
Add Device Reset

thorstendb-ARM:
Adds a Device Reset capabillity for pyOCD.
See https://pyocd.io/docs/command_reference.html#reset


jreineckearm:
Thanks, @thorstendb-ARM . Let's explore first on your fork what options we have to solve this. I need to raise an issue here first before we create PRs for it.

Issue #361, opened 2025-04-25
Continues attempt to connect if GDB server terminated early


jreineckearm:
**Problem**

Providing faulty input to a GDB server can cause it to terminate
* after the GDB server was at first successfully spawned...
* but before GDB tries to attempt the remote target connection.

This leaves the CDT GDB adapter in a state where the user sees the connection won't succeed. But they have to wait for the connection timeout.

**Proposal**

Still learning about the adapter in this area. But we should have information about the gdbserver having exited.
Hence, we should be able to terminate the debug session attempt earlier. Either event-driven or by polling the server state in `GDBBackend::sendCommand`. This should mitigate the above seen problems.

While at it, we should check the caught exception and see if we can transform the error message into something more useful. I see this on Windows which isn't very helpful to the average user in this situation:
![Image](https://github.com/user-attachments/assets/2576c8b4-af0f-4cbf-a213-8e3a12213e05)

It is understood we need to be extra cautious around the various supported scenarios (web vs desktop, local vs remote debug).

PR #384, opened 2025-06-02
Improve error reporting when missing remote configurations

omarArm:
Context:

When using the extension for target debugging. If a user forgets to set the port number for the gdb connection, connection is definitely going to fail. However, the reported error by the extension was not clear enough about what is the cause of the error.

Previous behaviour:

The previous reported error message used to be a complete log of the debug-console, which obviously was not user friendly.

Current suggested behaviour:

An error message will now pop up with the value 
```
Port number not specified, cannot connect
```
I also updated some launch related tests to accommodate the change.

### DIFF
--- a/src/integration-tests/launch.spec.ts
+++ b/src/integration-tests/launch.spec.ts
@@ -64,7 +64,7 @@ describe('launch', function () {
                     program: emptyProgram,
                     target: {
                         serverPortRegExp: 'Not a correct regex',
-                        portDetectionTimeout: 4000,
+                        portDetectionTimeout: 1000,
                     },
                 } as TargetLaunchRequestArguments)
             )


jonahgraham:
PS There are failing tests that should help guide you on getting the special cases resolved. Let me or @jreineckearm  know if you need any additional input

src/integration-tests/launch.spec.ts
```diff
@@ -52,6 +55,9 @@ describe('launch', function () {
     });
 
     it('reports an error when specifying a non-existent binary', async function () {
+        if (isRemoteTest) {
```
jreineckearm, line 58: Do we need two separate tests to cover remote and local use cases?

src/integration-tests/config.spec.ts
```diff
@@ -59,14 +63,31 @@ describe('config', function () {
...
     });
 
     it('can specify program via --config-frozen=', async function () {
-        const config = { program: emptyProgram };
+        let config;
+        if (isRemoteTest) {
+            config = {
+                program: emptyProgram,
+                target: {
+                    port: 2333,
```
jonahgraham, line 81: I don't understand why all the tests need ports adding here (but I think it relates to my other comment). Remote tests were automatically assigning ports (by launching gdbserver, reading the output for the assigned port and connecting to it). By hardcoding usage of port 2333 it means that tests are affected by global state on dev's machine. If dev is running JLink already the test will fail or try to connect to the running server.


> There are failing tests that should help guide you on getting the special cases resolved.

My previous comment on failing tests may have sent you down the wrong path, sorry about that. There may be some special cases in handling remotes, but most all the `*.spec.ts` files that have the new `if (isRemoteTest) {` blocks need them removed.

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -186,6 +186,13 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                                 ? target.serverStartupDelay
                                 : 0
                         );
+                    } else {
```
jonahgraham, line 189: I don't think this else is nested in the correct place. It is inside the individual line handler `checkTargetPort` callback.

The `checkTargetPort` is called on each line of output from gdbserver, some of the lines won't match the regex (i.e. `m` will be `null`). I think this means you will get false negative error responses that the port is not specified.

I think all your test changes mean that tests that used to run the code on lines 168-188+ no longer run those lines since you have port number explicitly set.

> LGTM [...]

Sorry that my previous review was insufficient as I didn't realize this was in the wrong place at that time. Now with the tests updated I can see that this isn't correct.

jonahgraham (commented):
@omarArm Sorry that there seems to be some misalignment here on the design and implementation. I hope we can come to a resolution.

I think one thing missing is a better problem statement, something that I (and others) can do to reproduce the bug that you are fixing here.

If you want to discuss further, such as have a zoom call about this, please drop me an email jonah (at) kichwacoders.com and we can perhaps discuss faster - perhaps when @jreineckearm is also available.

src/desktop/GDBTargetDebugSession.ts
omarArm: oh, ouch, That is actually why I needed to change all of the tests. It's a good thing you caught it now. I am open for a short meeting to get into this, I will send you an email right away then.

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -131,6 +130,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
     }
 
     protected async startGDBServer(
+        response: DebugProtocol.Response,
```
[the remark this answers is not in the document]
omarArm: I removed it now, thanks for noticing

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -68,7 +68,8 @@ export class GDBTargetDebugSession extends GDBDebugSession {
     protected async attachOrLaunchRequest(
         response: DebugProtocol.Response,
         request: 'launch' | 'attach',
-        args: TargetLaunchRequestArguments | TargetAttachRequestArguments
+        args: TargetLaunchRequestArguments | TargetAttachRequestArguments,
+        connectionTimeout?: number
```
jonahgraham, line 72: Nothing sets this argument.

> Make the timeout configurable

What I meant by this is add the timeout to `TargetLaunchRequestArguments`. It doesn't need to be a user documented one. See other comments that flesh out what I mean.

In my code suggestions I called it `connectionTimeout` - but that name isn't accurate. It isn't a connection timeout, but rather a port identification timeout. 

A connection timeout is a useful thing, and maybe we can add that too - that would be for cases when gdbserver launches but fails to connect to target in sufficient time.

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -165,6 +166,11 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                         : 0
                 );
             } else {
+                const timeoutForFindingPort = setTimeout(() => {
+                    reject(
+                        'Error: Port number not specified or regex is incorrect, cannot connect'
+                    );
+                }, connectionTimeout ?? 4000);
```
jonahgraham, line 173:
```suggestion
                }, target.connectionTimeout ?? 4000);
```

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -131,7 +131,8 @@ export class GDBTargetDebugSession extends GDBDebugSession {
     }
 
     protected async startGDBServer(
-        args: TargetLaunchRequestArguments
+        args: TargetLaunchRequestArguments,
+        connectionTimeout?: number
```
jonahgraham, line 135:
```suggestion
        args: TargetLaunchRequestArguments
```

src/desktop/GDBTargetDebugSession.ts
omarArm: Ah I see, sorry for the confusion. I will do it now

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -86,9 +87,8 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                 );
                 return;
             }
-            await this.startGDBServer(launchArgs);
+            await this.startGDBServer(launchArgs, connectionTimeout);
```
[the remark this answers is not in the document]
omarArm: Done

src/integration-tests/launch.spec.ts
```diff
@@ -51,6 +54,29 @@ describe('launch', function () {
...
+    it('receives an error when no port is provided nor a suitable regex', async function () {
+        if (!isRemoteTest) {
+            this.skip();
+        }
+        const errorMessage = await new Promise<Error>((resolve, reject) => {
+            dc.launchRequest(
+                fillDefaults(this.test, {
+                    program: emptyProgram,
+                    target: {
+                        serverPortRegExp: 'Not a correct regex',
```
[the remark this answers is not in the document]
omarArm: Done

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -180,6 +184,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                         setTimeout(
                             () => {
                                 gdbserverStartupResolved = true;
+                                clearTimeout(timeoutForFindingPort);
```
jreineckearm, line 187: I am a little worried that we create a race condition clearing the timeout here rather than before the `setTimout` call. Scenario is:
* Port gets detected by regexp.
* We are close to reaching the `portDetectionTimeout`.
* `serverStartupDelay` is not `0` and has to pass before the timeout callback clears the port detection timeout.
=> we successfully detected the port but exit the connection anyway because the delay extends the waiting time

jreineckearm:
One mild concern is that this change leaves behind running GDB servers. When it ran indefinitely before, a user was forced to click the stop button which terminated the server.

This isn't the only scenario that can lead to this and I am planning to enhance the process management as part of #361.
I guess the main question is whether we can live with this until the process management rework happened.

src/desktop/GDBTargetDebugSession.ts
jonahgraham: Good catch @jreineckearm - the clearTimeout is in the wrong place here. It should be cleared as soon as target.port is set (line 180 in this revision)

src/integration-tests/launch.spec.ts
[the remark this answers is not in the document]
jonahgraham: Please shorten the timeout in testing - by having 4000 here it means that the fastest this test can run is 4 seconds, but I don't think there is value in that. 1 second (or even less) should be ok.

See the [log](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/15734555002/job/44343700552?pr=384#step:8:514)

```
      ✔ /remote/receives an error when no port is provided nor a suitable regex: 4012ms
```

This 4 seconds accounts for >6% of the [total runtime](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/15734555002/job/44343700552?pr=384#step:8:698) of the tests:

```
  165 passing (59s)
```

jonahgraham (approved):
@jreineckearm I'll leave you to merge when it is convenient for you

jreineckearm:
About to merge #392 . Changes there should in theory help with the leftover processes I could see with this PR. Hence merging this PR here now.

jreineckearm:
Can confirm that the gdbserver gets terminated if used in combination with #392 .

PR #385, opened 2025-06-04
Setting a warning for enable/disable breakpoint commands

omarArm:
Unfortunately enable/disable bp commands supported by GDB cannot be reflected in the GUI using the normal Breakpoint event.

For now, a warning message will be printed in the debug console to the user whenever these commands are being used

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -1267,7 +1267,8 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 ) {
                     this.sendEvent(
                         new OutputEvent(
-                            'warning: "enable" and "disable" commands cannot be reflected in the GUI', 'stdout'
+                            'warning: "enable" and "disable" commands cannot be reflected in the GUI',
+                            'stdout'
                         )
                     );
                 }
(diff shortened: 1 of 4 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -1256,6 +1255,18 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             }
 
             if (args.expression.startsWith('>') && args.context === 'repl') {
+                const regexDisable = new RegExp('.*disable.*');
```
jreineckearm, line 1258: I think we need to be more specific about expecting spaces before the commands and before the breakpoint number. Can the argument be something else than a decimal number?
As it is, it will catch any expression that contains `disable` / `enable`. Even if part of another word. Try for example
`> break my_disable_variable`.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -1256,6 +1255,18 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                const regexEnable = new RegExp('.*enable.*');
+                if (
+                    args.expression.slice(1, -1).search(regexDisable) ||
+                    args.expression.slice(1, -1).search(regexEnable)
+                ) {
+                    this.sendEvent(
+                        new OutputEvent(
+                            'warning: enable and disable commands cannot be reflected in the GUI'
```
jreineckearm, line 1266: Maybe highlight `enable` and `disable` by adding quotes around them?
Maybe mention that they take effect despite not showing as such in the GUI.

jreineckearm (changes requested):
Please review the regular expressions. They unfortunately have side effects.
Might also be an opportunity to refine the error message.

PR #386, opened 2025-06-04
Avoid potential promise race conditions in `multithread` tests

jreineckearm:
For the time being a PR to reproduce CI failures and try timeout changes

### DIFF
--- a/src/integration-tests/multithread.spec.ts
+++ b/src/integration-tests/multithread.spec.ts
@@ -213,20 +213,22 @@ describe('multithread', async function () {
             ],
         });
 
+        const waitForStop = dc.waitForEvent('stopped');
         await dc.configurationDoneRequest();
-        await dc.waitForEvent('stopped');
+        await waitForStop;
 
         const threads = await dc.threadsRequest();
 
         // make sure that there is at least 2 threads.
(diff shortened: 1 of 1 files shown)


jonahgraham:
@jreineckearm I believe the current timeouts are very long and that normal operations complete dramatically faster than the timeouts. The reason win32 has long timeouts is because process creation on Windows is very very slow compared to Linux.

Historically tests timing out is race conditions, often fixed by using Promise.all to make sure the events aren't missed (see existing uses of Promise.all in the tests)

jreineckearm:
Oh, thanks for that hint, @jonahgraham !

This looks very much like a race condition we had before with breakpoint tests where the `waitForEvent` function needs to be set off before the actual operation provoking it, but then only wait for the promise after the operation completed. Like
```
    await dc.pauseRequest({ threadId: thread.id });
    await dc.waitForEvent('stopped');
```
where the stopped event may have been fired because of `pauseRequest` before `waitForEvent` starts.

Of course, the test succeeds now. Will probably anyway adjust it. Probably need to make this a focused action to go through all tests if we see this again.

jonahgraham (approved):

PR #390, opened 2025-06-25
Changed evaluation of variables to functions of gdb

thorstendb-ARM:
While adding "gdbtarget" to the extension "RTOS Viewer" I was wondering about the relatively complex handling of variable evaluation inside function handleVariableRequestObject() in src/gdb/GDBDebugSessionBase.ts. The issue was, that some of the RTOS variables could not be evaluated because of the array handling. Struct members were encapsulated into [] brackets where they shouldn't.

I refactored parts of that module to use gdb mi commands, see: https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Variable-Objects.html
-var-info-expression (for display elements)
-var-info-path-expression (to create and store expressions)

A "major" change is, that array elements are now displayed as "0 = ..." instaed of "[0] = ...".

After this change, the RTOS Viewer displayed FreeRTOS, and my tests with global nested struct / array variables did not show any issue.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2041,7 +2041,9 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         const stackDepth = await mi.sendStackInfoDepth(this.gdb, {
             maxDepth: 100,
         });
+
         const depth = parseInt(stackDepth.depth, 10);
+
         // we need to keep track of children and the parent varname in GDB
         let children;
         let parentVarname = ref.varobjName;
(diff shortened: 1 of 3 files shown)


asimgunes:
Hi @thorstendb-ARM ,

Previously there was a performance issue in the read operations of large array. When user adds a large array (e.g. array length 1K+ items) into the watch variables, there was a significant latency during the step into/over functions which is caused by reading the array again and again in when the thread stops. I afraid introducing the gdb commands into the for loop cause O(n) complexity and can reduce the performance in large arrays. It would be nice to check if this approach is re-introducing the previous issue or not, and I was just wondering if there is any other possible solution to avoid the GDB command executions inside the for loop?

PR #391, opened 2025-06-26
Fix variales array expression

thorstendb-ARM:
While adapting RTOS Viewer Extension for gdbtarget, I figured out that some variable expressions are turned into array expressions where they imho should not. A struct or array member which is an array itself gets [] for the access expression, which results in an expression: `((TCB_t*)0x20005cd0)[pcTaskName]`, where `pcTaskName` is a char [16] array.

I have changed this to use GDB directly, but this could cause bandwidth issues when running GDB remotely, see: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/390

If I got it correctly,
- `isArrayParent` determines if the element itself is an array
- `isArrayChild` determines if the element is part of an array

I want to understand why `isArrayParent` elements also gets brackets `[]`. I thought about two-dimensional arrays, but they run fine on just using `isArrayChild`.

This change for arrayParent fixes the issue on parsing (my) expressions and survives 2D arrays:
`const isArrayParent = arrayRegex.test(child.type) && arrayChildRegex.test(child.exp);`

@asimgunes, @jonahgraham, can someone bring some light into the darkness?

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2104,7 +2104,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 let name = `${ref.varobjName}.${child.exp}`;
                 const varobjName = name;
                 const value = child.value ? child.value : child.type;
-                const isArrayParent = arrayRegex.test(child.type);
+                const isArrayParent = arrayRegex.test(child.type) && arrayChildRegex.test(child.exp);
                 if(isArrayParent) {
                     console.log(`Found array parent: ${name}`);
                 }
@@ -2119,7 +2119,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 }
                 const variableName = isArrayChild ? name : child.exp;
                 const evaluateName =
-                    (isArrayParent || isArrayChild) && numberRegex.test(child.exp)
+                    isArrayParent || isArrayChild
                         ? `${topLevelPathExpression}[${child.exp}]`
                         : `${topLevelPathExpression}.${child.exp}`;
                 variables.push({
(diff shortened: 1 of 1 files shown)


jreineckearm (changes requested):
I agree with Thorsten, it's not entirely clear to me why we put the child expression into square brackets if the parent is an array. The parent access should be handled when the parent gets treated.

Also, I think we need to be a little more relaxed if there is no varObj (no parent). I think in that case it's enough to check if the child expression is a number/index. Otherwise we can easily create corner cases where we don't consider array children because of the current mechanisms...variable requests give us evaluate expressions for the parent but no type info.
Having said that, we need to check how this behaves when we use another variable as an index...

I see the code has seen some transformation over the years, especially around the `isArrayChild` and `isArrayParent` parts. So, may have accidentally come to the current behavior.

Unfortunately, I believe that also the changes in this PR need change. As they are, only arrays that are themselves array children ("indexes") are considered as arrays. Which is too narrow.

I agree with Jonah, that we need to add some test cases. In particular with combinations of `evaluate` requests and `variable` requests as they happen in the RTOS viewer uses case.
Unfortunately, I believe we have to live with the risk of breaking something to get this eventually right. And increase the test coverage as we go.

jonahgraham:
> I believe we have to live with the risk of breaking something to get this eventually right. And increase the test coverage as we go.

+1 to :point_up:

jreineckearm:
> it's not entirely clear to me why we put the child expression into square brackets if the parent is an array. The parent access should be handled when the parent gets treated.

That was non-sense (though, in my defense, I had the right conclusion in my head ;-) ). What I meant to write is "it's not entirely clear to me why we put the child expression into square brackets if it is an array parent. The array access should be handled when the array children are handled."

PR #392, opened 2025-06-27
Fixes and robustness of remote target GDB session connect/disconnect/exit on desktop

jreineckearm:
(Potentially) addresses
* #361 
* TODO add more issues that should be addressed by this.

Purpose of this PR is to fix defects and improve robustness of remote target GDB session process management for desktop. And related changes and fixes on the way.
* Avoid leftover processes...
  * ...if exiting through CLI (`quit`, `exit` commands).
  * ...if gdb or gdbserver get terminated outside regular usage (e.g. crashes or signals).
  * ...before connection completed but after gdbserver (and potentially gdb) processes were created already.
     Examples:
    * gdbserver gets launched, gdb gets launched, but gdbserver terminates unexpectedly because of bad CLI args processed later during startup. Especially if target connection parameters are given, hence no wait for port regular expression.
    * connections look good, but gdbserver suffers from fatal error while connecting to HW attached via JTAG debugger.
* Exit ASAP during connection when an error occurred instead of trying to complete the connection.
* Exit handler for gdb processes.
* Recover from gdb commands in-flight while gdbserver or gdb terminates.
* Only attempt to terminate gdb / gdbserver if really necessary.
* Added connection states to support the above.
* More verbose logging for diagnostics.
* Fixes lack of considering both process exit codes and exit signals.
* Extend `fillDefaults()` test util to prefer input arguments over defaults.

**Important**:
* Syncs web GDBTargetDebugSession with changed desktop code. However, no way to test the web variant. This requires help.
* Does not consider reuse for local GDB sessions. Might be worth to add in future to terminate IDE session when GDB/debugged application unexpectedly get lost. But will leave this to someone else.

**Out of scope** but planned for following PRs. This one became quite big already:
* Improve error messages. For example average user doesn't know what to make of term `ENOENT` or what "joining drives" on Windows got to do with launching a debug session (target select timeout error).
* Investigate more graceful ways to terminate gdbserver/give time to exit cleanly. A debugger backend may need time to cleanly shut down and turn off debug logic in target hardware. Could be by introducing shutdown delay after `disconnect`. Might also required to introduce support for TerminateRequest (which is less harsh than DisconnectRequest if launch request).
* Reduce redundant connect/disconnect state log messages.
* Improve error messages.
* Select target step can take a while to timeout. Find out if exit can be accelerated when hitting that call while something terminates in the backend.

### DIFF
--- a/src/desktop/GDBTargetDebugSession.ts
+++ b/src/desktop/GDBTargetDebugSession.ts
@@ -241,7 +241,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                 );
             }
             this.gdbserver = await this.gdbserverProcessManager.start(args);
-            this.logGDBRemote(`Spawned GDB Server (PID ${this.gdbserver.pid})`);
+            this.logGDBRemote(
+                `Spawned GDB Server (PID ${this.gdbserver.getPID()})`
+            );
             await this.setSessionState(SessionState.GDBSERVER_LAUNCHED);
 
             let gdbserverStartupResolved = false; // GDB Server ready for connection
--- a/src/desktop/processManagers/GDBServerFileSystemProcessManager.ts
+++ b/src/desktop/processManagers/GDBServerFileSystemProcessManager.ts
@@ -20,13 +20,13 @@ import {
     IStdioProcess,
 } from '../../types/gdb';
 
-type ConvertChildProcess = ChildProcess & GetPIDType;
+type ConvertedChildProcess = ChildProcess & GetPIDType;
 
 export class GDBServerFileSystemProcessManager
     extends GDBFileSystemProcessManagerBase
     implements IGDBServerProcessManager
 {
-    protected proc?: ChildProcess;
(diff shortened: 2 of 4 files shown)


thorstendb-ARM (approved):
Changes look promising to not leave gdb as zombies. 
Tested (Mac M3) >q (Quit), and by killing pyOCD and GDB processes. Debug Session always exits.
Will use in my current development and further test though.

jreineckearm:
@thorstendb-ARM , if you are still using this branch for local testing, could you please pull? Required some changes after merging dependency updates.

jreineckearm:
Rebased

jreineckearm:
Reminder that I am planning to merge tomorrow morning unless there is feedback from committers with more experience with this repo (@jonahgraham , @asimgunes ). We have more work in the pipe building up on this one. And unfortunately need to move forward for product releases on our end.
I am more than happy to work on code improvements and other feedback regarding this work even after merge of the PR.

jreineckearm:
Rebased

PR #401, opened 2025-07-15
Improve Breakpoint Error Logging

omarArm:
Problem:
When a user is using remote GDB and tries to set more than the amount of HW breakpoints, GDB sends a warning message and cannot step anymore.

Solution:
I created this PR, where a red error message will be printed to the user to ask for the extra breakpoints to be erased.

Future work:
When the user sets all of the breakpoints before starting the debug session the adapter sends an error message to VSCode, which in turn triggers a disconnect request back to the adapter. This should be mitigated somehow for a consistent behaviour between before and after the debug session starts

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -62,6 +62,12 @@ const numberRegex = /^-?\d+(?:\.\d*)?$/; // match only numbers (integers and flo
 const cNumberTypeRegex = /\b(?:char|short|int|long|float|double)$/; // match C number types
 const cBoolRegex = /\bbool$/; // match boolean
 
+// Interface for output category pair
+interface OutputCategory {
+    output: string;
+    category: string;
+}
+
 export function hexToBase64(hex: string): string {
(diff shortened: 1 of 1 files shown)


jreineckearm:
Addresses part 1) of #402

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -271,6 +277,20 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.sendResponse(response);
     }
 
+    private switchOutputToError(
+        input: string,
+        category: string
+    ): StreamOutput {
```
jreineckearm, line 283:
```suggestion
    private switchOutputToError(input: string, category: string): StreamOutput {
```

prettier

jreineckearm (approved):
LGTM! Apologies for applying review feedback myself. @omarArm is out for a couple of days.

I think `switchOutputToError` will evolve over time. I suspect there is more severe errors hidden in the blue logs. Improve the error transformation code when adding more.

PR #404, opened 2025-07-24
Run format-check before tests

jreineckearm:
Small change but hopefully big development efficiency improvement:
Run the prettier format check before running tests on Linux rather than after.
This should save us an ~8 minute waiting time to find out we forgot to run `yarn format` before commit.


thorstendb-ARM (approved):
LVGTM
v = very :-)

jonahgraham (approved):

PR #405, opened 2025-07-25
Give gdbserver time to gracefully disconnect before terminating it

thorstendb-ARM:


src/types/session.ts
```diff
@@ -148,6 +148,8 @@ export interface TargetLaunchArguments extends TargetAttachArguments {
     cwd?: string;
     // Maximum time allowed for detecting the port number, default is 10sec
     portDetectionTimeout?: number;
+    // Gracefully kill gdb-server when disconnecting, timeout in milliseconds, default is 5000
```
jreineckearm, line 151: Didn't we agree on a default of `0`?

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -145,7 +146,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         if (request === ExitSessionRequest.EXIT) {
             const shouldSendTerminateEvent =
                 this.sessionInfo.state >= SessionState.SESSION_READY;
-            await this.doDisconnectRequest(shouldSendTerminateEvent);
+            await this.doDisconnectRequest(0, shouldSendTerminateEvent);
```
jreineckearm, line 149: I believe this should use the timeout. It can be triggered by typing `exit` in the debug console.
GDB terminates first, kicking off these exit mechanisms. In that case, gdbserver still needs a bit of time to shutdown gracefully.

src/types/session.ts
jreineckearm: As discussed, we give it a try with 1 sec to have some waiting time by default but with low risk to break shutdown behavior for other use cases than ours.

thorstendb-ARM: renamed variable and set timeout to 1sec as discussed.

src/desktop/GDBTargetDebugSession.ts
thorstendb-ARM: done

src/desktop/GDBTargetDebugSession.ts
```diff
@@ -221,6 +225,11 @@ export class GDBTargetDebugSession extends GDBDebugSession {
 
         this.launchGdbServer = true;
         this.killGdbServer = target.automaticallyKillServer !== false;
+        this.serverDisconnectTimeout =
+            target.serverDisconnectTimeout !== undefined &&
+            target.serverDisconnectTimeout > 0
```
jreineckearm, line 230:
```suggestion
            target.serverDisconnectTimeout >= 0
```
We need to support value `0`.

jreineckearm (changes requested):

src/types/session.ts
```diff
@@ -148,6 +148,8 @@ export interface TargetLaunchArguments extends TargetAttachArguments {
     cwd?: string;
     // Maximum time allowed for detecting the port number, default is 10sec
     portDetectionTimeout?: number;
+    // Gracefully kill gdb-server when disconnecting, timeout in milliseconds, default is 1000
```
jreineckearm, line 151: Nit: the comment should be rephrased. It's about giving gdbserver time to gracefully shut down before we do the hard shutdown/termination.

jreineckearm (commented):
One minor comment. Apart from that good.
Please turn to full PR if good to get merged from your point of view.
Gonna give one more test tomorrow morning.

jreineckearm (approved):
Thanks for the changes! Really good to see our gdbserver properly shutting down and turning off the debug logic so that the application can continue running. :-)

PR #406, opened 2025-07-25
cdt-gdb-vscode 173: Add `watchServerProcess` setting

jreineckearm:
Fixes [Regression in version 2.1.0: Remote debugging with gdbtarget no longer starts correctly](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/173)

Adds a new setting `target`>`watchServerProcess` to monitor and handle the gdbserver process state, e.g. if unexpected early exists. Defaults to `true`, use to ignore such unexpected changes in the gdbserver process state. For example of `target`>`server` is a launcher for the actual gdbserver.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,6 @@
 # Change Log
 
-# Unreleased
-
+- Fixes [cdt-gdb-vscode 173](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/173): Add `target`>`watchServerProcess` to ignore early exit of `server` executable, e.g. if a launcher for the actual gdbserver.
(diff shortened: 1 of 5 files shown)


chroberino:
@jreineckearm I just tested it and it works like a charm!
`"watchServerProcess": false` => brings back the original behaviour from version 2.0.6 and earlier
`"watchServerProcess": true` => the new behaviour I have observed with version 2.1.0 (https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/173)

Thank you!

omarArm (approved):
I tested with  `"watchGdbServer": true` and `"watchGdbServer": false`. LGTM

PR #407, opened 2025-07-25
Fix getting stuck on concurrent breakpoint setup

cwalther:
### Problem

After being prompted using the `initialized` event, VS Code sends us `setBreakpoints`, `setFunctionBreakpoints`, and `setInstructionBreakpoints` requests in quick succession, without waiting for each to complete. The current code in `pauseIfNeeded` is not prepared for that, it can only deal with one request at a time. The immediate symptom is that when attaching to our embedded system, starting the debug session hangs.

The failure is only seen with targets that do not automatically stop all threads on attaching, so `pauseIfNeeded` actually does anything, such as ours (an embedded system behind our own GDB stub), unlike what gdb and gdbserver do when attaching to a Unix process. That is probably the reason why the problem has not been noticed previously.

### Analysis and proposed changes

The **first commit** adds a test, currently failing, that demonstrates the problem. It simulates such a target by issuing an `-exec-continue` right after attaching. It works in both all-stop and non-stop mode, although the situation I am interested in (used with our embedded target) is non-stop mode. Let me know if this test could be done better differently or elsewhere, I am still feeling my way around how the tests work.

What specifically happens is that when the second call to `pauseIfNeeded` arrives, it overwrites the previous `waitPaused` resolve function, and the first call then waits indefinitely on a promise that never resolves because its resolve function was lost.

This is fixed in the **second commit** by daisy-chaining the previous resolve function on the new promise.

That gets us a little further, however, we still redundantly issue three thread info requests and pause requests to GDB in a row when one would be sufficient, and worse, the first `continueIfNeeded` call from whichever of the three breakpoint requests completes first already ends the pause even though the other two are not done with their work yet (in my case, that does not cause any disruption, because our stub is capable of setting breakpoints with running threads). The second and third `continueIfNeeded` calls are answered by GDB with `error,msg="Cannot execute this command while the selected thread is running."`, although that doesn’t seem to be bothering anyone at this point.

There is an additional extra `sendExecContinue` in `configurationDoneRequest` whose `error,msg="Cannot execute this command while the selected thread is running."` response finally shuts down the debug session before it has a chance to become useful, but that will be dealt with further below.

The **third commit** fixes the redundant requests and early unpausing by introducing a reference count of active pause requests, so that only the first call to `pauseIfNeeded` will start a pause, and only the last call to `continueIfNeeded` will end it.

That still leaves the imperfection that if the breakpoint requests were to arrive not concurrently but sequentially (which they do not in my experience, but the protocol would allow it), with all previous ones complete before the next one arrives, we would still unnecessarily pause and continue multiple times, because the refcount would go to zero in between.

This is fixed by the **fourth commit**: During the configuration phase (which is the time span bounded by the `initialize` event and the `configurationDone` response), if and when the first request that requires pausing the target arrives, add another outer bracket of `pauseIfNeeded`/`continueIfNeeded` around the whole remainder of the phase, so that the refcount never goes to zero in between. As a side effect, this also fixes the extra `sendExecContinue` that previously killed the debug session, which is now covered by that final `continueIfNeeded`.

### Notes

I am a little puzzled about how this can ever have worked for anyone working on embedded systems, which seems to be the majority audience here. Does everyone stop their embedded system when attaching a debugger? I hope I am not breaking those people’s use cases in turn (tests still pass though).

I am planning to do further work in this area, but that is for separate PRs:
- I would like to make this pausing business optional, because our stub and target is capable of setting breakpoints without pausing any threads, and in fact we must not (even temporarily) pause arbitrary threads on attaching, because that could wreak havoc on the real-time behavior of a running industrial machine.
- The debug adapter protocol is not implemented correctly in that the `launch`/`attach` response is not delayed until after the `configurationDone` response as specified (clearly) in https://github.com/microsoft/vscode/issues/4902#issuecomment-368583522 and (a little less clearly, in my opinion) in the [protocol specs](https://microsoft.github.io/debug-adapter-protocol/overview#launch-sequencing).

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -55,6 +55,27 @@ class ThreadWithStatus implements DebugProtocol.Thread {
     }
 }
 
+/**
+ * Keeps track of where in the configuration phase (between initialized event
+ * and configurationDone response) we are.
+ */
+const enum ConfiguringState {
+    /** Configuration phase has not started yet. */
+    INITIAL,
+    /** Configuration phase has started, target is running, no requests that
+     * require pausing it have arrived yet. */
+    CONFIGURING,
+    /** Configuration phase has started, at least one request that requires
+     * pausing the target has arrived or it has been paused to begin with. */
+    CONFIGURING_PAUSED,
+    /** Configuration phase is completed, the next unpausing is the one
+     * associated with the end of the phase. */
+    FINISHING,
+    /** Configuration phase is completed, any following unpausing corresponds
+     * to a pausing outside of the configuration phase. */
+    DONE,
+}
+
 // Allow a single number for ignore count or the form '> [number]'
 const ignoreCountRegex = /\s|>/g;
 const arrayRegex = /.*\[[\d]+\].*/;
@@ -135,6 +156,9 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     // reference count of operations requiring pausing, to make sure only the
     // first of them pauses, and the last to complete resumes
     protected pauseCount = 0;
+    // keeps track of where in the configuration phase (between initialize event
+    // and configurationDone response) we are
+    protected configuringState: ConfiguringState = ConfiguringState.INITIAL;
     protected isInitialized = false;
 
     /**
@@ -365,8 +389,18 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 });
             }
         }
-        this.sendEvent(new InitializedEvent());
+        this.sendInitializedEvent();
         this.sendResponse(response);
+    }
+
+    protected sendInitializedEvent() {
+        if (this.isRunning) {
+            this.configuringState = ConfiguringState.CONFIGURING;
+        } else {
+            this.configuringState = ConfiguringState.CONFIGURING_PAUSED;
+            this.pauseCount++;
+        }
+        this.sendEvent(new InitializedEvent());
         this.isInitialized = true;
     }
 
@@ -433,6 +467,16 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected async pauseIfNeeded(requireAsync: true): Promise<void>;
 
     protected async pauseIfNeeded(requireAsync = false): Promise<void> {
+        // If we are in the configuration phase and this is the first request
+        // that requires pausing, add another pauseIfNeeded/continueIfNeeded
+        // bracket around the whole phase so we don't unnecessarily pause/
+        // continue more than once. Matching continueIfNeeded is in
+        // configurationDoneRequest.
+        if (this.configuringState === ConfiguringState.CONFIGURING) {
+            this.configuringState = ConfiguringState.CONFIGURING_PAUSED;
+            this.pauseIfNeeded(); // no need to await
+        }
+
         this.pauseCount++;
         if (this.pauseCount === 1) {
             this.waitPausedNeeded =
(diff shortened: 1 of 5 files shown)


jreineckearm:
Thanks for creating this PR, @cwalther . We'll have a look at this as soon as we can.

Would you mind creating issues for the remaining items you are planning to look into? This shall help to keep track of what's currently in the works, avoid duplicated efforts, and discuss approaches in more detail before an implementation.

> I am a little puzzled about how this can ever have worked for anyone working on embedded systems, which seems to be the majority audience here.

We haven't stumbled over the breakpoint issues yet that you treat in this PR. But we also see the need to relax the `pauseIfNeeded` logic for debugging embedded systems. And planned to look into this as part of improving the adapter's behavior for non-stop mode. AFAIK, gdb itself will pause the target system when setting a breakpoint in all-stop mode.
Do you operate your stub in non-stop mode?

cwalther:
OK, I will open some issues. (After investigating the CI failures on Windows – I had only tested on Linux locally.)

> AFAIK, gdb itself will pause the target system when setting a breakpoint in all-stop mode. Do you operate your stub in non-stop mode?

Yes, we only use non-stop mode. All-stop mode doesn’t make sense on a system where real-time-critical (e.g. motion control) threads need to continue running while you are debugging a non-real-time-critical (e.g. production sequence) issue, and also where the debug communication itself is handled by one of the threads running on the target that therefore must not be stopped.

You are right that GDB itself, even in non-stop mode, briefly stops one thread when attaching to a target where all threads are running, and I am not sure yet how to deal with that. I will have to, in one way or another. I would like to avoid patches to GDB as far as possible (we had more of them in the past). I may have to reintroduce some of the behavior we are currently using with Eclipse, where the stub “lies” to GDB and presents threads as “stopped” when they are actually not (according to what GDB considers “stopped”, i.e. “won’t continue unless I tell it to”), but just temporarily waiting on a synchronization object or descheduled because a thread of higher priority is running, so the situation that all threads are running from the point of view of GDB never occurs. But that required patches to the front-end (CDT DSF) in Eclipse, and now with Visual Studio Code I am still hoping to be able to work with an unmodified front-end.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -426,38 +465,84 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
-                this.waitPaused = resolve;
-            });
-            if (this.gdb.isNonStopMode()) {
-                const threadInfo = await mi.sendThreadInfoRequest(this.gdb, {});
+        this.pauseCount++;
+        if (this.pauseCount === 1) {
+            this.waitPausedNeeded =
+                this.isRunning && (!requireAsync || this.gdb.getAsyncMode());
+            if (this.waitPausedNeeded) {
+                let prevResolve = this.waitPaused;
```
jreineckearm, line 483: Would we need to also save `this.waitPauseThreadId` in a local variable in case multiple threads are paused and chained during init for non-stop?

cwalther: The idea is that when multiple pause operations are in flight at the same time, they all use the same `waitPauseThreadId` and only one thread gets paused, so no.

There is however a race condition in which that does not work: When the second `pauseIfNeeded` call of the mentioned `pauseIfNeeded`-`continueIfNeeded`-`pauseIfNeeded` sequence comes at a time when the first one is awaiting `queryCurrentThreadId()`, so `waitPausedThreadId` is still 0 and therefore `queryCurrentThreadId()` is called again, _and_ GDB between these two queries changes its current thread so they return different numbers, then two threads will get paused but only one of them continued later. Do we want to handle such convoluted edge cases? It could probably be done by keeping around not just the number but also the promise that resolves to it, so that the second call can also await the query of the first one instead of starting one of its own.

jreineckearm: In my experience, esoteric corner cases happen sooner than one would imagine. But I see that this would make this logic more complex. Let's go with this and see if it becomes a real problem.

jreineckearm:
Thanks for responding to my questions. And for updating the changelog. I tend to agree we may have to revisit if to insist on changelog updates in times of frequent activity is it is currently the case. I'll stop nagging people about it for now. :-)

Let me test it again for a bit, then I am happy to approve.

src/web/GDBTargetDebugSession.ts
```diff
@@ -485,10 +480,9 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             );
             // Connection completed, announce the adapter is ready for
             // other protocol commands.
-            this.sendEvent(new InitializedEvent());
+            this.sendInitializedEvent();
             this.sendResponse(response);
             await this.setSessionState(SessionState.SESSION_READY);
-            this.isInitialized = true;
```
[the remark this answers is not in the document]
jreineckearm: Tested a couple of happy paths, error cases with connection failures, and unexpected process losses.
Couldn't spot regressions.

PR #413, opened 2025-07-29
Unifying Error message between Windows and macOS in the case of Target Attaching error

omarArm:
In the `gdbtarget` mode, whenever there is gdbserver timeout error, the GDB-Adapter would always print the error message provided by GDB.

on macOS, the message is `could not connect: Operation timed out.`
on Windows, the message is `could not connect (error 138): The system tried to join a drive to a directory on a joined drive.`

That difference between platforms is not something that we would want for our users, and also the Windows error message is not very descriptive to the actual cause of the error.

I unified the error message to be always `could not connect: Operation timed out.`

### DIFF
--- a/src/desktop/GDBTargetDebugSession.ts
+++ b/src/desktop/GDBTargetDebugSession.ts
@@ -481,6 +481,16 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         return wrappedFunction;
     }
 
+    private switchAttachOutputToError(input: string): string {
+        const outputMessage = 'could not connect: Operation timed out.';
+        const returnPair: string = input.startsWith(
+            'could not connect (error 138): The system tried to join a drive'
+        )
+            ? outputMessage
+            : input;
+        return returnPair;
+    }
+
     protected async startGDBAndAttachToTarget(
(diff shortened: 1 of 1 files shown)


src/desktop/GDBTargetDebugSession.ts
```diff
@@ -633,7 +633,20 @@ export class GDBTargetDebugSession extends GDBDebugSession {
             // Complete connection failure response
             const errorMessage =
                 err instanceof Error ? err.message : String(err);
-            this.sendErrorResponse(response, 1, errorMessage);
+            if (
```
jreineckearm, line 636: It would be great to treat this similar to #401 . I expect more error messages needing transformation going forward. Then, having all of this in-place will become hard to read.
After this PR, we should also think about a more centralized approach of transforming error messages to something meaningful like in this case on Windows.

jreineckearm: Is this something that could be easily provoked in CI on Windows.
Having written that, I fear the `select-target` timeout to provoke this....though I believe it's configurable through a CLI `set` command...

Issue #362, opened 2025-04-25
Cannot execute CLI commands like `> interrupt` from Debug Console while CPU is running


jreineckearm:
**Problem**

One cannot use the Debug Console to execute GDB commands while a CPU/thread is running.
You get an error message about no `frameId` being available.

This includes CLI commands which are not dependent on a stopped CPU or a frame/thread ID.
Examples are
* `> help`
* `> interrupt`

**Proposal**

We should consider to NOT suppress Debug Console input when the CPU is running as we currently do in the `evaluateRequest` handler:
* I would expect GDB can gracefully handle commands that it cannot execute while CPU is running. This may even depend on the debugged architecture which may be even more so a reason to let GDB and/or the GDB server handle it.
* More generic commands that don't even need to access the target system shouldn't be affected by the CPU state.
* `-interpreter-exec` doesn't always require `--frame` and `--thread`.

Allowing non-GDB-command expressions from the console while the CPU is running is probably more debatable. It highly depends on the variable type (global vs local C/C++ variables) and target architecture. Variables at a fixed memory location (for example global C variables) could be for example read from an Arm Cortex-M based system while the CPU is running. If GDB itself doesn't just handle this gracefully, then we should consider treating this part separately from enabling GDB commands while CPU is running.

Note: success of GDB commands presumably depends on `non-stop` mode support. Again I'd expect GDB to reject gracefully.

PR #414, opened 2025-07-29
Allow sending commands to gdb in TargetDebug mode while target is running

thorstendb-ARM:
When target (on gdb-server) is running, the cdt-gdb-adapter blocks all commands and expressions.
Work for this has already been prepared here: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/383

Effectively this lets GDB decide if it can execute a command ior not.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -1386,7 +1386,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected async doEvaluateRequest(
         response: DebugProtocol.EvaluateResponse,
         args: DebugProtocol.EvaluateArguments,
-        allowIncompleteCommand?: boolean
+        allowIncompleteCommand: boolean
     ): Promise<void> {
         response.body = {
             result: 'Error: could not evaluate expression',
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -1385,14 +1385,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected async doEvaluateRequest(
         response: DebugProtocol.EvaluateResponse,
         args: DebugProtocol.EvaluateArguments,
-        allowIncompleteCommand: boolean // if true, allows evaluation of expression without a frameId
+        alwaysAllowCliCommand: boolean // if true, allows evaluation of expression without a frameId
     ): Promise<void> {
         response.body = {
             result: 'Error: could not evaluate expression',
             variablesReference: 0,
         }; // default response
         try {
-            if (!allowIncompleteCommand && args.frameId === undefined) {
+            const allowCliCommand =
+                alwaysAllowCliCommand && args.expression.startsWith('>');
+
+            if (!allowCliCommand && args.frameId === undefined) {
                 throw new Error(
                     'Evaluation of expression without frameId is not supported.'
                 );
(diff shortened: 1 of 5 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -273,6 +273,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         response.body.supportsWriteMemoryRequest = true;
         response.body.supportsSteppingGranularity = true;
         response.body.supportsInstructionBreakpoints = true;
+        response.body.supportsTerminateRequest = true;
```
jreineckearm, line 276: This change doesn't belong here. @thorstendb-ARM , please rebase. The PR with terminate request has been merged now.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -1378,13 +1379,21 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
         response: DebugProtocol.EvaluateResponse,
         args: DebugProtocol.EvaluateArguments
+    ): Promise<void> {
+        return this.doEvaluateRequest(response, args, false);
+    }
+
+    protected async doEvaluateRequest(
+        response: DebugProtocol.EvaluateResponse,
+        args: DebugProtocol.EvaluateArguments,
+        allowIncompleteCommand: boolean
```
jreineckearm, line 1389: I assume `incomplete` refers to some other (resume) command is still running?
Probably makes sense to put some commentary in explaining the purpose of this parameter. It may not be clear to everyone reading the code for the first time.

In general, I believe we wouldn't need to that cautious here regarding local and remote debug. But appreciate we can't always test all possible GDB settings variations. Hence, may be better to keep like this until someone sees a need for this during local debug and has the means to properly test it.

src/integration-tests/evaluate.spec.ts
```diff
@@ -56,6 +57,10 @@ describe('evaluate request', function () {
     });
 
     it('should reject evaluation of expression without a frame', async function () {
+        if (isRemoteTest) {
+            return;
```
jreineckearm, line 61: Please use `this.skip();` as done for other tests.

src/integration-tests/launchRemote.spec.ts
```diff
@@ -156,4 +156,26 @@ describe('launch remote', function () {
             true
         );
     });
+
+    it('should not reject evaluation of expression without a frame', async function () {
```
jreineckearm, line 160: I assume we still have the counterpart for failure in `launch.spec.ts`?

src/integration-tests/launchRemote.spec.ts
```diff
@@ -156,4 +156,26 @@ describe('launch remote', function () {
...
+                    type: 'remote',
+                } as TargetLaunchArguments,
+            } as TargetLaunchRequestArguments),
+            {
+                path: emptySrc,
+                line: 3,
+            }
+        );
+
+        const ret = await dc.evaluateRequest({
```
jreineckearm, line 174: Are you sure the target is running at this point?
You may need to set it running first (probably use the `loopForever` program instead).

src/integration-tests/launchRemote.spec.ts
thorstendb-ARM: there are tests for local gdb-server (including this one's origin) testing for rejection.

src/integration-tests/launchRemote.spec.ts
thorstendb-ARM: fixed

jreineckearm: You still would need to set the target program running.

src/integration-tests/launchRemote.spec.ts
```diff
@@ -156,4 +157,22 @@ describe('launch remote', function () {
             true
         );
     });
+
+    it.only('should not reject evaluation of expression without a frame', async function () {
```
jreineckearm, line 161: Unfortunately missed the `.only` here. @thorstendb-ARM , please open another PR to fix it.

jreineckearm:
Addresses #362

Issue #402, opened 2025-07-17
Better handle setting too many breakpoints


jreineckearm:
If you have a target system
* with a limited number of HW breakpoints,
* where a program runs from non-volatile memory which requires HW breakpoints,

then you can run into situations where you set more breakpoints than supported. This leads to errors when trying to run/step the target system.

1) These errors show in the Debug Console. However they don't look scary enough (printed in blue) given that you can't proceed with program execution.

<img width="738" height="181" alt="Image" src="https://github.com/user-attachments/assets/5357bbf3-cadc-42b5-9367-cad02c4db8cd" />

They should be printed in red like proper errors.

2) It looks like something chokes in the debug connection (either IDE to adapter, or adapter to gdb, or gdb to gdbserver). And you cannot recover from this state. The IDE thinks target is running, but the pressing the pause button has no more effect.

Some investigations point to the MI parser command queue removing commands too early to properly report back an error and hence to return to the stopped state of the IDE.

PR #415, opened 2025-07-29
Handle recovery from late out-of-band error arrivals

jreineckearm:
Address #402 , but actually goes beyond to fix the root cause.

Problem: If you set too many breakpoints
* gdb may report a completed (e.g.) `continue` command as `running`/`done`.
* gdb/gdbserver may install breakpoints and start running.
* Along the way, gdbserver may reject adding breakpoints if not enough HW resource is available.
* gdb may send a late `error` response referring to the previously successfully completed `continue` command.

Before this change, this could heavily confuse the adapter states and provoke funny behavior.

This PR changes this to recover from such situations. It
* stores additional, artificial information with the MI result data in the MI parser, prefixed with `cdt-` to avoid confusion with real MI payload
  * 'cdt-token': token of a command
  * 'cdt-command': original command string of a completing command that is still in the queue.
* Changes the MI parser command queue to add this data.
* Adds a `resultAsync` event between MI parser and session classes sending GDB async results. 'cdt-command' is only added for the first result for a token.
* Extends the thread state to include the token of the last command setting it running. (Cleared on stopped notifications)
  * Handle `done`/`running` results of successful commands that resume the target (sorry about that list of potential commands, I know it's ugly). Store their token per thread.
  * Handle `error` results for commands with tokens stored in the thread states. These are the out-of-band errors and require returning internal and VS Code thread states to stopped.

Note: Still a draft as I need to add a few pinpointed tests.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -45,18 +45,7 @@ import { calculateMemoryOffset } from '../util/calculateMemoryOffset';
 import { isWindowsPath } from '../util/isWindowsPath';
 import { sendResponseWithTimeout } from '../util/sendResponseWithTimeout';
 import { DEFAULT_STEPPING_RESPONSE_TIMEOUT } from '../constants/session';
-
-class ThreadWithStatus implements DebugProtocol.Thread {
-    id: number;
-    name: string;
-    running: boolean;
-    lastRunToken: string | undefined;
-    constructor(id: number, name: string, running: boolean) {
-        this.id = id;
-        this.name = name;
-        this.running = running;
-    }
-}
+import { ThreadWithStatus } from './common';
 
 /**
  * Keeps track of where in the configuration phase (between initialized event
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2260,6 +2260,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 if (stopAll) {
                     // All threads are stopped
                     this.sendStoppedEvent('error', this.threads[0]?.id ?? 1, true);
+                    this.isRunning = false;
                 } else {
                     // Selection of threas is stopped, send individual events
                     stopThreads.forEach(thread => this.sendStoppedEvent('error', thread.id, false));
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -50,7 +50,7 @@ class ThreadWithStatus implements DebugProtocol.Thread {
     id: number;
     name: string;
     running: boolean;
-    lastRunToken: string|undefined;
+    lastRunToken: string | undefined;
     constructor(id: number, name: string, running: boolean) {
         this.id = id;
         this.name = name;
@@ -90,6 +90,7 @@ const threadIdRegex = /.*\s+\-\-thread\s+(\d+).*/;
 const threadAllRegex = /.*\s+\-\-all(\s+.*|$)/;
 
 const resumeCommands = [
+    // GDB MI
     '-exec-continue',
     '-exec-finish',
     '-exec-jump',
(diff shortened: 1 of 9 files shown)


cwalther:
No worries, you are not obligated to respond quickly. From a quick merge test there doesn’t seem to be any overlap with what I’m doing, and if there is, I will adapt.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2047,6 +2092,87 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                        const threadId = threadIdMatch
+                            ? parseInt(threadIdMatch[1], 10)
+                            : undefined;
+                        updateThreads = this.threads.filter(
+                            (thread) => thread.id === threadId
+                        );
+                    }
+                    updateThreads.forEach(
+                        (thread) =>
+                            (thread.lastRunToken = notifyData['cdt-token'])
```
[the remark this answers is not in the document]
jreineckearm: Thanks! Well spotted, indeed doesn't make sense to access `notifyData` again.

thorstendb-ARM (approved):
Tested on Jens' branch with STLink on pyOCD.
- Before: gdb session crashes with a strange error message and is not reacting anymore (must be killed).
- After: "HW breakpojnt limit reached, ..." is reported and session remains reactive. After removing some BPs, session runs as expected. Supports up to 6 BPs (or 5 + tempBP @ main)

image

jreineckearm:
Rebased

jreineckearm:
Preparing a 1.2.0 release without this change. This needs more testing for the `non-stop` mode. Or alternatively a guard to apply only to all-stop mode in first incarnation.

src/integration-tests/lateAsyncErrorsRemote.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+        let stoppedEvent: DebugProtocol.StoppedEvent | undefined = undefined;
+        await Promise.all([
+            new Promise<void>(async (resolve) => {
+                stoppedEvent = (await dc.waitForEvent(
+                    'stopped'
+                )) as DebugProtocol.StoppedEvent;
+                resolve();
+            }),
+            dc.configurationDoneRequest(),
+        ]);
```
cwalther, line 71: This looks overcomplicated, why not
```ts
        let [stoppedEvent] = await Promise.all([
            dc.waitForEvent('stopped'),
            dc.configurationDoneRequest(),
        ]);
```
?

(Just some bikeshedding from a quick glance, not a thorough review.)

src/integration-tests/lateAsyncErrorsRemote.spec.ts
```diff
@@ -0,0 +1,158 @@
...
+        'inner1 stop': 0,
+    };
+
+    this.beforeAll(function () {
+        resolveLineTagLocations(src, lineTags);
+    });
+
+    beforeEach(async function () {
+        dc = await standardBeforeEach('debugTargetAdapter.js');
+        await dc.launchRequest(
```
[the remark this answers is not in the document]
jreineckearm: Awesome, thanks for reviewing! Was about to ask if you could have a look as you are looking into these at the moment. 👍 
If it's not too much extra work for you later, I'd like to get this fix in today or early next week. Planning to make a release next week and finishing this up was pending way too long on my ToDo list.

src/integration-tests/lateAsyncErrorsRemote.spec.ts
jreineckearm: Updated in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/415/commits/27de94ae3c2de734a31ff22f21bd02a2f2f452cf

jreineckearm:
Rebased and added a test for the command queue and state recovery over DAP.
No functional change, already received feedback on added test.
Waiting for re-review of my changes, planning to merge latest on Monday.

src/integration-tests/lateAsyncErrorsRemote.spec.ts
cwalther: Heh, so do I! We’re learning together. 🙂

PR #416, opened 2025-07-29
Avoid unnecessary ThreadInfoRequests

### LINKED ISSUE #408: Avoid unnecessary ThreadInfoRequests
In order to figure out GDB’s current thread, `GDBDebugSessionBase.pauseIfNeeded` calls `mi.sendThreadInfoRequest`, which requires an expensive roundtrip to the target, only to throw away all the information it got from the target, because the current thread ID is known by GDB itself. There should be a simpler, more efficient way of querying that, something along the lines of `-data-evaluate-expression $_gthread`.

cwalther:
To query GDB's current thread ID, don't use `-thread-info`, which does an expensive roundtrip to the target (several kilobytes of XML transmitted over the remote protocol connection). The current thread is known by GDB itself, and while there does not seem to be a command to query it directly, it can be read by evaluating convenience variable `$_gthread`.

Closes #408

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -434,12 +434,22 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.waitPaused = resolve;
             });
             if (this.gdb.isNonStopMode()) {
-                const threadInfo = await mi.sendThreadInfoRequest(this.gdb, {});
-
-                this.waitPausedThreadId = parseInt(
-                    threadInfo['current-thread-id'],
-                    10
+                const threadIdResult = await mi.sendDataEvaluateExpression(
+                    this.gdb,
+                    this.gdb.gdbVersionAtLeast('7.11')
+                        ? '$_gthread'
+                        : '$_thread'
                 );
+
+                // Fallback to -1 means things still happen to work
+                // because "--thread -1" is equivalent to omitting
+                // "--thread" (that is probably an implementation detail
+                // though).
+                this.waitPausedThreadId =
+                    threadIdResult.value !== undefined &&
+                    threadIdResult.value !== 'void'
+                        ? parseInt(threadIdResult.value, 10)
+                        : -1;
                 this.gdb.pause(this.waitPausedThreadId);
             } else {
                 this.gdb.pause();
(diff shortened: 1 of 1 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -434,12 +434,22 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.waitPaused = resolve;
             });
             if (this.gdb.isNonStopMode()) {
-                const threadInfo = await mi.sendThreadInfoRequest(this.gdb, {});
-
-                this.waitPausedThreadId = parseInt(
-                    threadInfo['current-thread-id'],
-                    10
+                const threadIdResult = await mi.sendDataEvaluateExpression(
```
jreineckearm, line 437: We do have some level of abstraction for GDB version based behavior already here: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/main/src/gdb/GDBBackend.ts

Maybe it would make sense to extend https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/dbc7f50cf23f46ba9f4aa0696edb4cd43db2d77d/src/types/gdb.ts#L78 by a new function to get the current thread ID, and move the "how" to the GDBBackend. I came across some code where such a cheaper variant of getting the current thread ID might become handy in future.

jreineckearm:
Thanks, @cwalther . Much appreciate the updates!
`query....` sounds good to me.

Thanks for the heads up on your holidays. I hope to get back to your PRs after our releases today.
Will merge if no more changes are necessary.
Glad to see these new contributions coming in from you, thanks for that! :-)

cwalther:
Thank you for your quick reactions!

No more changes necessary from my side for two weeks, I am off now (tomorrow is a holiday).

jreineckearm:
> Thank you for your quick reactions!
> 
> No more changes necessary from my side for two weeks, I am off now (tomorrow is a holiday).

Enjoy the time off!

jreineckearm (approved):
Changes LGTM!
Using -1 should be fine as far as I can see from a quick investigation of using this thread ID and reading on the web.
Using the global thread ID if available (opposed to the current inferior thread ID) also makes sense to me.

jreineckearm:
Congratulations to your first merged PR on this repo, @cwalther ! 🎉 

One thing I missed to mention was that we maintain a CHANGELOG.MD that servers as release notes.
We normally review the CHANGELOG vs actual commits before release. But any brief entry done as part of PRs saves some effort for the release preps.

PR #417, opened 2025-07-30
Removed .only from test

thorstendb-ARM:


jreineckearm:
Follow up to #414 where we missed a `.only` committed for tests.

PR #420, opened 2025-08-08
Disabling evaluate request error messages when hovering over comments

omarArm:
When hovering over comments in code, vscode sends an evaluateRequest to the adapter, which in turn sends it to GDB. This causes an unnecessary popping error message to users all the time.

I disabled the error message only when hovering


jreineckearm (approved):
LGTM

PR #426, opened 2025-08-12
Covering the case when source file for Breakpoint is unknown

omarArm:
Background:

When debugging, sometimes the debugger won't know the source file in which a breakpoint is set. Accordingly when the adapter gets a notification that is associated with such a breakpoint, it would try to set the value of `source` and `line` in the breakpoint interface to be sent in a breakpoint event. This is a faulty behaviour and it causes problems in the breakpoints window in the IDE.

Changes:

Now whenever the adapter receives a GDB notification for a modified and unknown BP, it will look first if the source information is provided, if not, then it won't use it

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2021,7 +2021,9 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     if (notifyData.bkpt.disp === 'del') {
                         break;
                     }
-                    const breakpoint: DebugProtocol.Breakpoint = {
+                    let breakpoint : DebugProtocol.Breakpoint;
+                    if (notifyData.bkpt.line) {
+                        breakpoint = {
                         id: parseInt(notifyData.bkpt.number, 10),
                         verified: true,
                         source: {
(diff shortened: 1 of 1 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2021,15 +2021,25 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
-                        id: parseInt(notifyData.bkpt.number, 10),
-                        verified: true,
-                        source: {
-                            name: notifyData.bkpt.fullname,
-                            path: notifyData.bkpt.file,
-                        },
-                        line: parseInt(notifyData.bkpt.line, 10),
-                    };
+                    let breakpoint: DebugProtocol.Breakpoint;
+                    if (notifyData.bkpt.file) {
```
jreineckearm, line 2025: Could there be a `fullname` without `file`?

jreineckearm (commented):
Changes look sensible (only one question but I guess it's fine).
I suspect this corner case cannot be easily tested in CI?
Please add a CHANGELOG entry.

Let's try to add issues for such fixes in future, even if just for the books. Makes it easier to add in releases notes. But not for this PR anymore.

src/gdb/GDBDebugSessionBase.ts
omarArm: From what I have seen so far, it is not possible to have one attribute without the other

PR #433, opened 2025-08-21
Fix a flaky test

cwalther:
Some of the tests in breakpointModes.spec.ts occasionally fail with `Error: no event 'stopped' received after 2500 ms`. This is because the stopped event sometimes comes before the test starts waiting for it. A `dc.waitForEvent` call must always be done _before_ whatever causes the event, in this case `await dc.configurationDoneRequest()`. This can be done either using `await Promise.all([…])`, as in my proposed solution, or by saving the promise returned by `waitForEvent` in a local variable and awaiting it later.

Moreover, these tests wait for multiple output events using repeated calls to `await dc.waitForEvent('output')`. This is unsafe and may miss some events, because promise continuations, i.e. anything that comes after `await`, are executed asynchronously. Multiple events may have been emitted before the continuation that handles the first and starts waiting for the next one gets to run.

I don’t think there is any way of properly handling this using `waitForEvent`, so my solution drops down to the building blocks that `waitForEvent` is built upon, NodeJS event handlers, which are executed synchronously and therefore do not suffer from this problem.

There are more tests that suspiciously use `await dc.waitForEvent(…)` after `await dc.configurationDoneRequest()` and therefore are potentially flaky, but examining and fixing them all is left for later.

I am omitting a changelog entry here since this is not something that affects users, is that what is expected?

### DIFF
--- a/src/integration-tests/breakpointModes.spec.ts
+++ b/src/integration-tests/breakpointModes.spec.ts
@@ -65,22 +65,45 @@ const expectBreakpoint = async (
     expect(response.body.breakpoints.length).eq(1);
     expect(response.body.breakpoints[0].verified).eq(true);
     expect(response.body.breakpoints[0].message).eq(undefined);
-    await dc.configurationDoneRequest();
 
-    let isCorrect;
-    let outputs;
-    while (!isCorrect) {
-        // Cover the case of getting event in Linux environment.
-        // If cannot get correct event, program timeout and test case failed.
-        outputs = await dc.waitForEvent('output');
-        isCorrect = outputs.body.output.includes('breakpoint-modified');
-    }
+    // Wait for the first output event that contains 'breakpoint-modified' and
+    // for the first stopped event after that.
+    // Cannot use repeated dc.waitForEvent('output') for that because it might
+    // miss some.
+    const [[outputs, stoppedOutput]] = await Promise.all([
+        new Promise<[DebugProtocol.Event, DebugProtocol.Event]>(
+            (resolve, reject) => {
+                const timeout = setTimeout(
+                    () =>
+                        reject(
+                            new Error(
+                                'No correct sequence of events received after 2500 ms'
+                            )
+                        ),
+                    2500
+                );
+                function checkOutputEvent(outputEvent: DebugProtocol.Event) {
+                    if (
+                        outputEvent.body.output.includes('breakpoint-modified')
+                    ) {
+                        dc.off('output', checkOutputEvent);
+                        dc.once('stopped', (stoppedEvent) => {
+                            clearTimeout(timeout);
+                            resolve([outputEvent, stoppedEvent]);
+                        });
+                    }
+                }
+                dc.on('output', checkOutputEvent);
+            }
+        ),
+        dc.configurationDoneRequest(),
+    ]);
+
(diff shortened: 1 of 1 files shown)


cwalther:
> but examining and fixing them all is left for later.

Okay, “later” has (in part) come sooner than expected and I’ve tacked on a fix for another one that has now started to annoy me. This fix is simpler than the first one.

cwalther:
Whoops, I thought I had done the code formatting. This should fix the CI failure.

src/integration-tests/breakpointModes.spec.ts
```diff
@@ -65,22 +65,45 @@ const expectBreakpoint = async (
...
-        // If cannot get correct event, program timeout and test case failed.
-        outputs = await dc.waitForEvent('output');
-        isCorrect = outputs.body.output.includes('breakpoint-modified');
-    }
+
+    // Wait for the first output event that contains 'breakpoint-modified' and
+    // for the first stopped event after that.
+    // Cannot use repeated dc.waitForEvent('output') for that because it might
+    // miss some.
+    const [[outputs, stoppedOutput]] = await Promise.all([
```
jreineckearm, line 73: 👍 Nice approach

src/integration-tests/breakpointModes.spec.ts
```diff
@@ -65,22 +65,45 @@ const expectBreakpoint = async (
...
+                                'No correct sequence of events received after 2500 ms'
+                            )
+                        ),
+                    2500
+                );
+                function checkOutputEvent(outputEvent: DebugProtocol.Event) {
+                    if (
+                        outputEvent.body.output.includes('breakpoint-modified')
+                    ) {
+                        dc.off('output', checkOutputEvent);
```
jreineckearm, line 89: I see we are creating a new `dc` before each test. So, should be fine to leave it registered for the error case.

src/integration-tests/breakpointModes.spec.ts
```diff
@@ -65,22 +65,45 @@ const expectBreakpoint = async (
...
+                        dc.once('stopped', (stoppedEvent) => {
```
jreineckearm, line 90: Guess there is still a minimal chance of race conditions here between detecting `breakpoint-modified` and registering the `stopped` handler. But looks much more reliable than before. 👍 
Wondering if we can turn this somehow into a more generic helper function. But probably would need to see if there are more occurrences of checking such sequences.

jreineckearm (approved):
Nice! Thanks for taking some initiative here. Good to go with this changes.

src/integration-tests/breakpointModes.spec.ts
cwalther: I don’t think there is a race condition anymore? That sequence is all synchronous now, no promises/`await`s involved anymore.

Anyway, I’m not sure whether we couldn’t just have waited for the stopped event from the beginning, but I just replicated what the previous code was doing here.

I think I have seen more such occurrences in passing, but didn’t examine in detail. Next time a flaky test annoys me…

cwalther:
Thanks! I’m just tackling things that get in my way while I’m trying to scratch my itch… this is two levels deep, first level is coming in the next PR, and then I can get back to work on what I actually wanted to do. 🙂

PR #438, opened 2025-08-26
Fix my own flaky test

cwalther:
Turns out I am guilty myself of the crime that I accused others of in #433: Tests that start waiting for an event only after the call that triggers the event and therefore occasionally miss it (introduced in #407).


jreineckearm (approved):
LGTM

Issue #99, opened 2019-05-23
Pause doesn't work on Windows


dschaefer:
CDT has a complex system for sending gdb a Control C event to get it to pause. We'll need to bring that over and update it (still works with Windows 2000 though :) ).

tmirza:
Ping. Any update on this?

jonahgraham:
Pause does not work still on Windows native debugging. It works whereever we can run GDB async, which is generally all targets except native windows.

tromey:
> Pause does not work still on Windows native debugging. It works whereever we can run GDB async, which is generally all targets except native windows.

Windows target-async was implemented last year: https://sourceware.org/pipermail/gdb-patches/2022-August/191190.html

jonahgraham:
Thanks @tromey - I appreciate pointing me at that update. I'll try a newer version of GDB on Windows and we'll advertise that as the minimum version.

noneghost:
Whether it is caused by the special mechanism of CTRL-C-EVENT on Windows

noneghost:
I tried this version  _GCC 13.1.0 (with POSIX threads) + LLVM/Clang/LLD/LLDB 16.0.5 + MinGW-w64 11.0.0 (UCRT) - release 5   (LATEST)_  [https://github.com/brechtsanders/winlibs_mingw/releases/download/13.1.0-16.0.5-11.0.0-ucrt-r5/winlibs-x86_64-posix-seh-gcc-13.1.0-llvm-16.0.5-mingw-w64ucrt-11.0.0-r5.7z](url) of gdb async is pauspendable. GDB version is **GNU gdb (GDB for MinGW-W64 x86_64, built by Brecht Sanders) 13.2
**
_

jonahgraham:
Thanks @noneghost for testing this. 

> [...] we'll advertise that as the minimum version.

Therefore GDB 13 is the minimum version on Windows host to support native debugging. I am not sure where to document it. For now I am going to change the title of this bug.

noneghost:
> Thanks @noneghost for testing this.
> 
> > [...] we'll advertise that as the minimum version.
> 
> Therefore GDB 13 is the minimum version on Windows host to support native debugging. I am not sure where to document it. For now I am going to change the title of this bug.

noneghost:
> Thanks @noneghost for testing this.
> 
> > [...] we'll advertise that as the minimum version.
> 
[...]

PR #440, opened 2025-08-28
Fix erroneous assumption of async mode when target does not support it

cwalther:
The `-gdb-set mi-async on` command currently used ([documentation](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Asynchronous-and-non_002dstop-modes.html)), even when it suceeds, does not guarantee that GDB will actually enter async mode, it only announces a preference. Entering async mode additionally requires support from the target. If the adapter is later going to make decisions based on whether GDB is in async mode, it needs to check whether that is actually the case, not just rely on its own preference.

In particular, the `native` target on Windows does not support async mode, so during local debugging on Windows with a launch configuration that does not set `gdbAsync: false`, `gdb.getAsyncMode()` currently erroneously returns true and any decisions based on that are wrong. In such cases, the adapter may issue commands, expecting that they will be serviced promptly, and get stuck because they are in reality only serviced next time the target stops.

In the current main branch, there is only one such decision, in `GDBDebugSessionBase.pauseIfNeeded`, but more are about to be added (by me) in #437 and #439. The first one may be harmless, as it only guards a `gdb.pause()` call that should also work in sync mode (I’m not sure whether it actually does on Windows, see https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/commit/a481615a446038bbd5c1419d028b96b996414018). The one from #437 might cause the adapter to get stuck when setting breakpoints while the program is running, but I have not tried it (there is a test for that, but it is skipped on Windows – possibly because the feature has never worked there, for the same reason). The one from #439 definitely causes problems, as seen in a failed CI run there.

So this must either be fixed, or #437 and #439 must be reworked because `GDBBackend.getAsyncMode()` does a different thing than I expect and I am misusing it in those. I hereby vote for the former.

According to the documentation,
> After the frontend has started the executable or attached to the target, it can find if asynchronous execution is enabled using the `-list-target-features` command.

Some experimentation and a look at the source code seem to suggest that the executable need not actually be started (which is good, because once it’s started, it’s running, and if we are unexpectedly in sync mode, we can’t even send the command to discover that at that time), but GDB does need to be connected to a target. Connection to a target happens either explicitly using the `target`/`-select-target` command (in the remote case) or implicitly to the `native` target using `run`/`-exec-run` or `attach`/`-target-attach` (see [`set auto-connect-native-target`](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Starting.html#set-auto_002dconnect_002dnative_002dtarget)) (in the local case). These commands happen in slightly different places in the adapter in remote/local/launch/attach scenarios, and they may even be part of `target.connectCommands` supplied by the user rather than explicitly issued by the adapter (the latter is the case for Indel).

I think I have covered all those places in the attached proposed fix. I would however in particular appreciate another set of eyes on where I explicitly do `-target-select native` because when it would be done implicitly by `-exec-run` later it would be too late in the unexpectedly-sync case. Are we actually guaranteed to end up with the `native` target at this point so this does not break anything, or am I missing any corner cases?

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -352,6 +352,15 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         );
     }
 
+    protected warnAsyncDisabled() {
+        this.sendEvent(
+            new OutputEvent(
+                'The target does not support gdbAsync. Most debugger interactions will only work while the target is stopped.',
+                'important'
+            )
+        );
+    }
+
     protected async attachOrLaunchRequest(
         response: DebugProtocol.Response,
         request: 'launch' | 'attach',
@@ -384,6 +393,26 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             this.sendEvent(
                 new OutputEvent(`attached to process ${attachArgs.processId}`)
             );
+            if (this.gdb.getAsyncMode()) {
+                if (await this.gdb.confirmAsyncMode()) {
+                    this.warnAsyncDisabled();
+                }
+            }
+        } else {
+            if (this.gdb.getAsyncMode()) {
+                // Checking whether the target supports async mode needs a
+                // target connection. Launching will implicitly select the
+                // "native" target (see docs of `set
+                // auto-connect-native-target`), so we may just as well do it
+                // explicitly here.
+                await mi.sendTargetSelectRequest(this.gdb, {
+                    type: 'native',
+                    parameters: [],
+                });
+                if (await this.gdb.confirmAsyncMode()) {
+                    this.warnAsyncDisabled();
+                }
+            }
         }
 
         await this.gdb.sendCommands(args.initCommands);
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -393,11 +393,6 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             this.sendEvent(
                 new OutputEvent(`attached to process ${attachArgs.processId}`)
             );
-            if (this.gdb.getAsyncMode()) {
-                if (await this.gdb.confirmAsyncMode()) {
-                    this.warnAsyncDisabled();
-                }
-            }
         } else {
             if (this.gdb.getAsyncMode()) {
                 // Checking whether the target supports async mode needs a
@@ -409,13 +404,18 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     type: 'native',
                     parameters: [],
                 });
-                if (await this.gdb.confirmAsyncMode()) {
-                    this.warnAsyncDisabled();
-                }
             }
         }
 
         await this.gdb.sendCommands(args.initCommands);
+
+        // Check for async mode support after initCommands have possibly
+        // selected a different target
+        if (this.gdb.getAsyncMode()) {
+            if (await this.gdb.confirmAsyncMode()) {
+                this.warnAsyncDisabled();
+            }
+        }
 
         if (request === 'launch') {
             const launchArgs = args as LaunchRequestArguments;
--- a/src/desktop/GDBTargetDebugSession.ts
+++ b/src/desktop/GDBTargetDebugSession.ts
@@ -612,6 +612,11 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                     )
                 );
             }
+            if (this.gdb.getAsyncMode()) {
+                if (await this.gdb.confirmAsyncMode()) {
+                    this.warnAsyncDisabled();
+                }
+            }
 
             await this.setSessionState(SessionState.CONNECTED);
 
--- a/src/desktop/GDBTargetDebugSession.ts
+++ b/src/desktop/GDBTargetDebugSession.ts
@@ -612,11 +612,6 @@ export class GDBTargetDebugSession extends GDBDebugSession {
                     )
                 );
             }
-            if (this.gdb.getAsyncMode()) {
-                if (await this.gdb.confirmAsyncMode()) {
-                    this.warnAsyncDisabled();
-                }
-            }
 
             await this.setSessionState(SessionState.CONNECTED);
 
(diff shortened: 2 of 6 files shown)


cwalther:
> I agree the explicit `-target-select native` needs some extra attention.

What I was mainly wondering in that regard: Is `GDBTargetDebugSession` only intended for the `remote` and `extended-remote` targets? Is `GDBDebugSession` only intended for the `native` target? What about other [targets](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Target-Commands.html) like `core` and `sim`, are they even supported, does anyone use them?

> (e.g. log a warning or an error if not).

Good point, I agree that a warning would be in order for a user that explicitly requested async mode. For one that did not however, but just got it as the default, not so much. They probably don’t care about it, don’t know what getting sync mode means for them, and wouldn’t understand what the warning is trying to tell them. I will look into how easy these cases are to distinguish and report back with findings and/or adjusted code.

For now, here is just the unmodified code (but with some comments adjusted to the GDB 13 situation I wasn’t aware of) rebased to clean up the spaghetti history.

> Please see #99 - note that that issue used to be titled "Pause doesn't work on Windows"

Thanks, that’s good to know (TLDR: Async mode on Windows native is supported from GDB 13 on). That doesn’t really answer the question of whether our `GDBBackend.pause` method would work _in sync mode_ with GDB ≥13, but it does make it less important because there is less reason to use sync mode.

cwalther:
I have tried a few options in the meantime. 🙂

* Rebased on main.
* Added a warning when async mode is explicitly requested but not available, as a GUI notification and in the debug console. (Squashed into the first commit.)
* In the second commit, move the check after `initCommands`, but leave the target selection before (my suggestion above).
* Alternatively, in branch [confirmasync-main-a](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/compare/main...indel-ag:cdt-gdb-adapter:confirmasync-main-a) ([4e12432](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/commit/4e12432bb93b4a2b4a6276878d9544f2c4abb711)), move both the target selection and the check after `initCommands`, with target selection conditional on whether the feature check returns an empty list. The case in which this does the wrong thing is when the user selects a different target than `native` in `initCommands`, but one that supports neither `async` nor `reverse`. In that case, the user’s choice will be overwritten with `native` again. Also, this costs an additional GDB command in the usual case (feature check is done twice).

Maybe you can have a look at those, weigh the hypotheticals, and see which of the three options you prefer (first commit only, second commit here, second commit alternative).

I’m taking a look at the CI failures now.

jreineckearm:
@cwalther , let's go with your suggestion. Overriding the the `native` select with whatever is in the `initCommands` sounds in the meanwhile like the more sensible thing to do. Will give it another quick try and merge in the morning and when CI is going again on Windows.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -352,6 +352,15 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         );
     }
 
+    protected warnAsyncDisabled() {
+        this.sendEvent(
+            new OutputEvent(
+                'The target does not support gdbAsync. Most debugger interactions will only work while the target is stopped.',
+                'important'
```
jreineckearm, line 359:
```suggestion
                'The target does not support gdbAsync. Most debugger interactions will only work while the target is stopped.\n'
```

How does this render for you in the Debug Console?
For me it's
image

If I change to the above, it gets (note the linebreak, without it the empty line before the gdb disclaimer misses)
image

Also, it seems I use a slightly different gdb flavour on my Windows machine. That supports async and I had to provoke the output by changing the returned feature array in a Typescript debug session of the adapter.

jreineckearm (approved):
Thanks for working on the feedback, looks good to get merged. Minor comment can be addressed separately.

src/gdb/GDBDebugSessionBase.ts
cwalther: I admit I had only tested this with verbose mode on, in which case the line breaks in the debug console are a bit of a mess anyway. Thanks for checking this.

I can confirm your observation, but I am not sure your suggestion is an improvement. I deliberately send the message in the `'important'` category, instead of the default `'console'` one as your suggestion does, so that there is a GUI notification, for users who don’t have the Debug Console open. Would you prefer not to have that?

It does seem a bit inconsistent that VS Code renders `important` in the same color as `stdout` (meant for output from the debuggee) and not in the same color as `console` (meant for output from the debugger). But I would rather not work around that if it incurs other costs.

I am a bit puzzled by the `Breakpoint 1 at 0x…` line in your screenshots – the check for async support is supposed to come before setting breakpoints (the `initialized` event in between should enforce this order). If this is reproducible for you, could I have a full log with `verbose` enabled in the launch configuration?

> Also, it seems I use a slightly different gdb flavour on my Windows machine. That supports async

Is it version 13 or newer? That would explain that. It’s only the older versions that don’t support it, and only when launching/attaching directly – connecting via `gdbserver` supported it before.

jreineckearm: One correction:
I must have had a different GDB in the path when I wasn't able to provoke the problem initially. Retrying with 11.2 and an extension build from a branch including your changes shows it without meddling with variables in the TS debugger.

I still see the breakpoint hit before the message. Note that I set a breakpoint in `initCommands`. Here my launch config
```
        {
            "type": "gdb",
            "request": "launch",
            "name": "gdb launch",
            "cwd": "${workspaceFolder}",
            "gdb": "gdb",
            "gdbAsync": true,
            "verbose": true,
            "program": "xxxxxxxx\\cdt-gdb-adapter\\src\\integration-tests\\test-programs\\loopforever.exe",
            "initCommands": [
                "break main"
            ]
        }
```

And here a zip with logs with and without the change: [logs.zip](https://github.com/user-attachments/files/22425837/logs.zip)

cwalther: > Note that I set a breakpoint in `initCommands`.

Ah, that’s fine then. Nothing unexpected in that case. I was just worried about the adapter setting breakpoints and making decisions about pausing before it had confirmed async support.

cwalther:
Thanks for the merge and for the reviews! Moving on to #439 now.

PR #444, opened 2025-09-11
Adding more robustness to warning messages of the evaluateRequest

omarArm:
Apparently the GDB commands `enable` and `disable` have multiple ways to be sent to GDB. I added these variants to the regex that is responsible for issuing the warning


omarArm:
@cwalther you are right, I will have to create a new PR to cover all of the possible scenarios then. Hopefully once the DAP is updated to include enable/disable in breakpoint interfaces the whole need for a warning message will be eliminated

PR #450, opened 2025-09-29
Cleans up and fixes tests for unexpected early exits for remote launches

jreineckearm:
Tests were ineffective due to incorrect checking of  sinon spy results. These don't work well with the debug client logger and how it gets created. Switched to inspecting stdout output event contents. Fixed tests which were checking for wrong output (early gdb server exit).


thorstendb-ARM (approved):
LGTM

PR #454, opened 2025-10-02
Support auxiliary GDB with target parameters/connectCommands

jreineckearm:
Follow up to feedback for #449 :

Enables support for auxiliary GDB connections when using target parameters or target connect commands.
See https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/449#pullrequestreview-3289635301


jonahgraham (approved):

jreineckearm:
> I am a bit surprised that the two new tests succeed in the non-remote scenarios, but didn’t investigate whether that is expected or accidental.

Ah, well spotted. This is a mixture of both. I copy-and-pasted the test which tested the argument validation which expects things to fail rather than skipping. Probably should change the tests to skip for the two new tests to save some time.

About to leave my desk in a couple of minutes and it's a public holiday tomorrow in Germany, so may only fix this on Monday.

cwalther:
LGTM.

PR #452, opened 2025-10-01
New test for multiple variables in multiple threads at multiple depths

jonahgraham:

Issue #451, opened 2025-10-01
Explicitly set thread ID when getting stack depth


jreineckearm:
Found by @jonahgraham during code review:

We have a couple of calls to `mi.sendStackInfoDepth` which do not receive a threadId parameter although we have the knowledge in the surrounding code.

This should in theory be fine because the affected functionality is always executed on the currently selected thread. Which should be correctly selected in GDB. But we should make use of exact knowledge when we have it, like for most of these calls. Especially if there is a lot going on in the GUI which could change the thread selection while operations are in flight.

jreineckearm:
Thanks, @jonahgraham . I agree something is wrong. And also start having a feeling depth is redundant because of carrying a frame ID in the variable key....

jonahgraham:
Thanks @jreineckearm - my reading of the above messages show me you went down a similar rabbit hole that I did. 

One important thing to note about our implementation is that I don't think we use [gdb var objects](https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Variable-Objects.html) well. We do have a var manager, but we don't take advantage of the functionality of var objects (in particular we don't ask gdb to notify us when vars change\*, we do everything by polling every single variable). This makes everything much more complicated than it probably needs to be, and we never delete vars, which means we probably have a memory leak/reduced performance on very long runs with lots of different variables. My open question is, do we need to keep var objects around at all, or can we do everything with either temporary var objects and/or direct methods. If we don't keep vars around, then we don't need those particular depth calls at all.

\* We do var-update on each variable rather than doing `-var-update *`

jreineckearm:
Quick update on the oddities I saw: these were caused by an optimized out pointer-to-struct object where I successfully ignored the `<optimized out>` value of the variable itself.
We should disallow to expand such variables. This otherwise ends up in some decent confusion as var object children don't show `<optimized out>` due to empty values. They show nothing or types.
Entering an expression in the Watch window which includes struct members/arrays again gives you the `<optimized out>` value as expected.

jreineckearm:
Made the sendStackInfoDepth calls consistent. But the varobject caching where this is used needs improvement. At the the moment, its strategy is based on thread, frame, and stack depth keys and doesn't seem to have a proper clear-strategy. There is some risk for long running real-world use cases where this may unnecessarily consume memory.
Keeping this ticket open until creating a follow-up for that which references the above history.

PR #455, opened 2025-10-04
Consistently use threadId in MI sendStackInfoDepth.

jreineckearm:
Related to #451 

Use the parameter consistently, it is available in all places.

Staring at the new test from #452 for a while and methods how to test the depth-stuff in general. But I am starting to have doubts that testing through variable access on integration level is the right level to do that.

What we really need to understand is whether the thread ID is propagated correctly, which means tracking logs to see parameters are passed as expected. And I am a little hesitant to spend a lot of effort on testing the caching right now which clearly needs a bit of rework.

Alternative would be to restore the previous behavior of passing threadId to stack-depth calls exactly as before to avoid regressions.


jonahgraham (approved):
I think passing the correct thread everywhere should be fine. Everything I said in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/451 still applies, but I agree that until we can fix it all, lets make it consistent.

PR #459, opened 2025-10-15
Fix evaluateName expressions for arrays and array children.

### LINKED ISSUE #400: Evaluation of variables to support RTOS Views extension
While adapting RTOS Views extension for gdbtarget, I figured out that some variable expressions are turned into array expressions where they imho should not. A struct or array member which is an array itself gets [] for the access expression, which results in an expression: ((TCB_t*)0x20005cd0)[pcTaskName], where pcTaskName is a char [16] array.

I have changed this to use GDB directly, but this could cause bandwidth issues when running GDB remotely, see: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/390

If I got it correctly,

isArrayParent determines if the element itself is an array
isArrayChild determines if the element is part of an array
I want to understand why isArrayParent elements also gets brackets []. I thought about two-dimensional arrays, but they run fine on just using isArrayChild.

This change for arrayParent fixes the issue on parsing (my) expressions and survives 2D arrays:
const isArrayParent = arrayRegex.test(child.type) && arrayChildRegex.test(child.exp);

@asimgunes, @jonahgraham, can someone help?

Ref. (two approaches):
- minimum required changes to make RTOS Views extension work (would prefer now):
  - https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/391
- greater approach (prob. later, if expressions require):
  - https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/390

jreineckearm:
Another attempt to fix #400 .

* Array names shouldn't be put into square brackets, only their children.
* Identify array entry indexes by being numbers if varobj parent not present.
* Additionally check parent varobj to be an array for array children. May be redundant but reduces diff to previous code for this commit.

Thanks to @thorstendb-ARM for the initial investigation of this in #390 and #391 !

Review of the functionality uncovered a few other oddities to address. See #460 .
However, those are not critical to unblock our use of the extension we aim to use in our solution. Hence, they can be addressed separately and step by step.

### DIFF
--- a/src/integration-tests/evaluate.spec.ts
+++ b/src/integration-tests/evaluate.spec.ts
@@ -239,7 +239,9 @@ describe('evaluate request global variables', function () {
         ];
         children.body.variables.forEach((variable, index) => {
             expect(variable.name).to.equal(childrenContents[index].name);
-            expect(variable.evaluateName).to.equal(`s0.${childrenContents[index].name}`);
+            expect(variable.evaluateName).to.equal(
+                `s0.${childrenContents[index].name}`
+            );
             if (childrenContents[index].hasChildren) {
                 expect(variable.variablesReference).not.to.equal(0);
             } else {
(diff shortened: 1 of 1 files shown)


thorstendb-ARM (approved):
Tested with RTOS Viewer / Arm:RTX5, works

PR #461, opened 2025-10-20
1.4.1 Patch Release Preparations

jreineckearm:
* Bumps version number to 1.4.1
* Updates CHANGELOG for release.

### DIFF
--- a/.github/workflows/build-pr.yml
+++ b/.github/workflows/build-pr.yml
@@ -50,15 +50,20 @@ jobs:
         run: git diff --exit-code
   Build-on-Windows:
     name: Build & Test on Windows
-    runs-on: Windows-2022
+    runs-on: windows-latest
     steps:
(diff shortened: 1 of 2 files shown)


jreineckearm:
Found how to make CI use the mingw installation from chocolatey. Which should help to pin versions going forward.

But the "newest" mingw version from chocolatey that works now has an even older GDB in it (10.2).
Mingw 12.x versions from chocolatey seem to contain already a GDB 12.1. Which is different from what was installed on the GH runners with mingw 12.x (GDB 11.2) according to the runner image info?
GDB 12.1 gives me errors about broken string tables for the unicode files.

Will have another look tomorrow. But may have to live with either that, or with disabling some Windows tests on new GDBs until we get to the bottom of that. Not entirely sure at this point what's the lesser evil.

Issue #463, opened 2025-10-24
Cannot use custom reset while CPU is running


jreineckearm:
It currently is not possible to execute the custom reset functionality if the CPU is running.
This is in all-stop mode where you are limited in the commands you can call.

Custom reset should silently stop the target before executing the commands for reset.

One could expect that the target is set running again after the reset, if it was running before.
However, it is hard to predict what commands are used for the custom reset. And when the target really stops based on the command sequence. Commands could include setting a breakpoint and a continue command, where the breakpoint is hit only after completion of the custom reset logic. `continueIfNeeded` would provoke errors if target was already/still running as part of reset.

Unfortunately, GDB/MI doesn't give us simple access to the current run state of the target. In order, to have a well-defined and predictable behavior, the custom reset does NOT proactively resume the device if it was running before pressing the reset button. It remains stopped unless the `customResetCommands` explicitly resume it.

Happy to get this further refined in future if someone has a good idea how to capture/restore the state after reset.

jreineckearm:
Potential future fix for #430 would enable better implementation than what's possible today.

PR #464, opened 2025-10-24
Silently pause target before reset if running

jreineckearm:
Addresses #463 

Updated 03/11/2025, see feedback and discussion in below thread
* Custom reset logic calls new `pauseIfRunning` before calling `customResetCommands`.
* Introduces separate logic to pause target if running and waiting for the pause to happen. Logic comes with usage constraints/assumptions, e.g. not being called during connection/configuration.
* Validates if `customResetCommands` is used with `gdbAsync`.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -659,31 +659,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
      *   - No requireAsync, supposed to be removed from pauseIfNeeded in future.
      */
     protected async pauseIfRunning(): Promise<void> {
+        // Check if running before actual wait promise
+        if (!this.isRunning) {
+            // At least one thread is already stopped, good enough
+            // to execute custom reset commands.
+            return;
+        }
         return new Promise<void>(async (resolve) => {
-            let doResolve = false;
-
-            let currentThreadId: number | undefined;
-            if (this.gdb.isNonStopMode()) {
-                // Get current thread ID
-                currentThreadId = await this.gdb.queryCurrentThreadId();
-                // Check if current thread is running.
-                const currentThread = this.threads.filter(
-                    (thread) => thread.id === currentThreadId
-                );
-                if (currentThread.every((thread) => !thread.running)) {
-                    // Current thread already paused, nothing left to do.
-                    doResolve = true;
-                }
-            } else {
-                if (!this.isRunning) {
-                    // Not running, nothing left to do.
-                    doResolve = true;
-                }
-            }
-            if (doResolve) {
-                resolve();
-                return;
-            }
+            // Get current thread ID in non-stop mode
+            const currentThreadId = this.gdb.isNonStopMode()
+                ? await this.gdb.queryCurrentThreadId()
+                : undefined;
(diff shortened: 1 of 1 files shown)


thorstendb-ARM (approved):
Tested with STLink on Nucleo-L476. Runs as expected.

jreineckearm:
Added validation of arguments. Ready for another review once CI is happy.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -623,6 +645,52 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+        if (this.gdb.isNonStopMode()) {
+            // this.isRunning means at least one thread is running. Check if current one.
+            const currentThread = this.threads.filter(
+                (thread) => thread.id === currentThreadId
+            );
+            if (currentThread.every((thread) => !thread.running)) {
+                // Current thread already paused
+                return;
+            }
+        }
```
cwalther: The comments here are incorrect, so the early-return logic can be simplified: `isRunning` means “all threads are running” (with a special case for no threads), not “any threads are running” – see `updateIsRunning`.

```suggestion
        if (!this.isRunning) {
            // Some threads paused already
            return;
        }
```

Note that technically there are race conditions here, but they are not new, so don’t need fixing in this PR: `isRunning` reflects the last state that GDB has confirmed to us, but what we should be checking here is the last state that we have asked GDB for. If `isRunning` is true but another pause request is already in flight, we don’t need another one; conversely if a continue request is in flight but we haven’t received confirmation of its effect yet so `isRunning` is still false, we do need a pause request.

jreineckearm:
Thanks again for the review, @cwalther . Should be addressed now. Also changed a bit the way of acquiring the resolve-callback which was a bit ugly in the version before.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -628,6 +650,57 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                if (currentThread.every((thread) => !thread.running)) {
+                    // Current thread already paused, nothing left to do.
+                    doResolve = true;
+                }
+            } else {
+                if (!this.isRunning) {
+                    // Not running, nothing left to do.
+                    doResolve = true;
+                }
+            }
```
cwalther: I still don’t understand why you need to make sure that the _current_ thread is stopped. Isn’t it sufficient to have _any_ thread stopped (which is tracked by `isRunning`)?

Isn’t the current thread somewhat arbitrary anyway? I haven’t checked what influences it. The only other place we call `queryCurrentThreadId()`, we just need _any_ thread ID, needn’t be the current one.

jreineckearm: Initial thinking was to avoid issues with execution commands that don't carry an explicit thread ID. But the more I think about it, the more I believe proper support for this mode needs to be developed against a real debug target. Which I don't have access to. At the moment there is too much guessing without having a reference platform with real-world behavior.

Hence, simplified logic and only check for `isRunning`. Assumption is that at least one thread stopped gives us enough control for what we need to do.

cwalther (approved):

Issue #465, opened 2025-10-27
UTF-8 'Failed to decode cstring' errors for GDB with CP1252 support only


jreineckearm:
In our solution, we are using a GDB variant on Windows which only supports 'CP1252' as host and target charset.

However, the CDT GDB adapter always decodes incoming character strings as UTF-8 in the MI parser module.
This can cause lots of unnecessary errors being printed to the Debug Console. "Unnecessary" because then after printing the error, the undecoded string is returned and things work nicely.

We should detect if a GDB supports any UTF format at all. Or if the supported charset is ASCII only like in our case (that's what's behind CP1252).

Adding knowledge about all possible charset names may become too big. Hence, the approach should explicitly exclude those cases which are known to be problematic.

jreineckearm:
Fixed with #466

Issue #446, opened 2025-09-16
CI fails on Windows


cwalther:
As discovered in [a CI run of](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/17765008636/job/50486212913?pr=440) #440, some tests now fail on Windows. They still worked [last week](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/17645432957/job/50142132911?pr=444).

The failing tests are:
* "after each" hook for _attach remote_ ▸ _can detach from a running program_, in all scenarios where it runs (defaults, remote, gdb-async-off, remote gdb-async-off).
* _launch_ ▸ _works with unicode in file names_, in all scenarios where it runs (defaults, gdb-async-off).

What I have found out so far:
* The difference is that [the OS image changed from Windows Server 2022 to Windows Server 2025](https://github.com/actions/runner-images/issues/12677).
* Apparently we are not using the GDB that is downloaded as part of the job, but one that is supplied with the OS image. That was version [11.2](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/17645432957/job/50142132911?pr=444#step:4:498) before and [16.2](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/17765008636/job/50486212913?pr=440#step:4:494) now.
* We could work around the CI failures by reverting to the `Windows-2022` image, but if there are actual problems with GDB 16, we should probably fix them anyway.
* I can reproduce the problem that GDB 16.2 becomes unresponsive when the `gdbserver` that it is attached to is killed while the program is running, after outputting `warning: Exception condition detected on fd`.

jreineckearm:
Thanks for investigating, @cwalther . I'd suggest to switch back to 2022 to unblock the tests until we have time to look into 16.2. This should unblock ongoing development. I've started draft #447 for the first.

jreineckearm:
Looks like now also `windows-2022` runner images got updated with new GNU toolchain parts. Which makes this pretty much a blocker unless one can explicitly select runner images/runners with specific images... 😒

jreineckearm:
Did some investigation and had a chat with a trusted AI:
It might be related to https://sourceware.org/bugzilla/show_bug.cgi?id=31577

[...]

PR #466, opened 2025-10-27
Replace utf8 package with using TextDecoder API

jreineckearm:
Addresses #465 

Some eyes on this before adding more tests here or in a separate PR would be appreciated. May save me some time if more rework required.

* Replaces use of utf8 package with using TextDecoder API which supports whatwg encodings ( https://encoding.spec.whatwg.org/ ).
* Detects and uses initial host charset encoding to configure MI parser for treating octal char sequences in c-strings.
* Only decode the octal char sequences opposed to previously decoding the entire stream output.
* Adds fixup map for encodings supported in GDB that are effectively aliases for whatwg encodings. Saw this with Windows CP65001 which effectively is utf-8. Extend as defects pop up, this is not making things worse than they were before.
* Synchronizes charset changes between MAIN and AUX connections.

**IMPORTANT**: It's important now to correctly set up the charsets to see the correctly decoded output in the IDE, i.e. make sure to configure a Windows GDB to CP65001 or UTF-8 if it launches as CP1252. TextDecoder gets configured per the host-charset setting.

_Not solved with this PR_:
Problems with newer GDBs (#446): It looks like we catch an double-encoding of the UTF-8 characters while loading the file resulting into becoming unreadable for GDB. Already tried to set the host encoding to UTF-8 before loading the program without success. Needs to be looked at separately but I don't think it got worse with this PR.

_Notes_:
* Command post-processing and the split between session base and GDB backend isn't great. But best I could come up with in short time and without major redesign.
* Re-read of `host-charset` after `set` commands isn't optional. However, strong doubts that switching charsets was heavily used in the past. Problems would have been much more prominent. Hence, I believe it's good enough for now and doesn't impact existing users.
* More testing needed in future.
* Still may misunderstand some things around the charsets. May need more refinement over time based on real user feedback and testing. I have only limited resource in terms of host OSes and GDB variants.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## Unreleased
+
+- Fixes [`465`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/465): UTF-8 'Failed to decode cstring' errors for GDB with CP1252 support only.
+
 ## 1.4.1
 
 - Fixes [`400`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/400): Evaluation of variables to support RTOS Views extension.


jreineckearm:
More thinking on this: start agreeing that skipping UTF8 decoding for CP1252 only isn't sustainable for future. But only covers the main use case for us where this flags up (a GDB variant on Windows which only supports CP1252). Similarly, always decoding UTF8 as is today is also only covering a single use case (probably the most common one, though?).

Full fix needs something like https://www.npmjs.com/package/iconv-lite (not investigated yet if there are better alternatives) to cover the myriads of charsets supported by GDB across the major OSes. See the `gdb/charset-list.h` in GDB source. But certainly will need more testing effort.

Intermediate solutions to at least silence the error messages, which don't really help our users, would be:
* Use the above incomplete fix as a first iteration. We need to get the charset as done here anyway for the proper decoding. And it's the main part of this PR. I am also happy to explicitly use `latin1` as `toString()` argument for CP1252.
* Add something like an error level option in addition to the verbose, warn, error log options. Though I am not sure if we want to open up that can of worms in form of discussions around preferences for what goes on which level.
* Log this particular error only as a verbose-log. Not sure at the moment if users can do anything about them anyway except from trying to change their (host) charset to UTF8 if their OS and GDB allows.

I probably have to park this for now unless we go for one of the cheaper intermediate solution. I've got more usability and stability issues on on my ToDo list for our next Arm CMSIS Debugger release.

jonahgraham (approved):
This looks ok to me - I guess it is a major pain to have a gdb that doesn't support host-charset of utf-8. Eclipse CDT afaict doesn't support this case either. CDT always sets host-charset to utf-8 and always sets sevenbit-strings to on.

PR #467, opened 2025-10-31
Not do set breakpoint request if there is no breakpoint when configuration phase has not completed yet

datphan9798:
In configuration phase, breakpoint requests are still sent even there is no breakpoint was set. When do "set breakpoint request", continue and interrupt command could be sent to gdb and those commands could change the state of target unexpectedly. So should not do set breakpoint request if there is no breakpoint when configuration phase has not completed yet.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -124,6 +124,11 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected gdb!: IGDBBackend;
     protected auxGdb?: IGDBBackend;
 
+    protected isBreakpointSent = false;
+    protected isDataBreakpointSent = false;
+    protected isFunctionBreakpointSent = false;
+    protected isInstructionBreakpointSent = false;
+
     protected isAttach = false;
     /**
      * True if GDB is using the "remote" or "extended-remote" target. In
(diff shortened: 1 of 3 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -717,6 +717,16 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         response: DebugProtocol.SetDataBreakpointsResponse,
         args: DebugProtocol.SetDataBreakpointsArguments
     ): Promise<void> {
+        // If there is no breakpoint and configuring has not done yet,
+        // do nothing to not change the state of target
```
jonahgraham:
```suggestion
        // If there are no breakpoints to set, and configuring is not done yet,
        // do not change state of the target by avoiding an unnecessary pause
```

cwalther (changes requested):
Logic looks good now and works for me.

Please run `yarn format && yarn lint` to fix the formatting.

Your tests have some bad parentheses – please run them and check that they work: Not just check that they pass, but also check that they fail when you break the thing that they are testing.

cwalther:
> @cwalther feel free to press "Approve and run" on PRs if they are not nefarious.

I would have but I don’t have the permission (needs a “maintainer”).

cwalther:
@datphan9798 Some of the additional formatting issues that CI (Ubuntu) found in GDBDebugSessionBase.ts, as I had, are whitespace-only, maybe that’s why you missed them when committing.

jonahgraham:
@datphan9798 the commit is still failing format checks. Did you run `yarn format` and commit those changes? Are you seeing a different result from `yarn format` than is happening on build machine?

datphan9798:
> The last test failure has nothing to do with your change but is just flakiness due to the race condition mentioned in #437.
> 
> So this looks good to me now, thanks! I am not a maintainer, so the rest is up to those now.

Thank you!

PR #471, opened 2025-12-03
Relaxed error handling

### LINKED ISSUE #473: Confusing error pop-ups without additional user value in some corner cases
Our users see a lot of activity in terms of error pop-ups.
* Race conditions between long lasting (and/or by the IDE delayed) window updates and user interaction like stepping/resume. (Late request arrivals from views, requests already fed into GDB command queue which then fail on GDB level after target state change).
* Views in other VS Code extensions not immediately ceasing to send DAP requests while the debug adapter is already shutting down.
* Recently introduced command queue flushing for graceful shutdown throws unhandled errors.

These things make the solution look premature and buggy. And users normally can't do anything about such situations. And most of the times they wouldn't even miss the lack of accurately returned data or errors in such corner cases. Accurate view updates become irrelevant during shutdown. And when setting the CPU running/stepping either views go blank anyway (non-aux-gdb mode), or get updated during the next periodic refresh/stopped event which is normally acceptable for users (note that this could be further refined in future by returning cached data, e.g. for variables, but I don't see an urgent need for that right now, hence returning responses with empty data should be fine).

jreineckearm:
Fixes #473 

**Approach**

I first experimented with adding synchronization elements (locks, barriers) on the GDB backend/MI parser command queue. But this won't work well in reality, even if going for full serialization of commands which contradicts with the idea of gdb async. It also will impact stepping performance perceived by users.

* Ended up with using knowledge about current debug adapter/GDB state in combination with specific error types for selected operations. To smoothen the user experience, we now return successful responses but with empty data to increase the odds of views stopping to send more requests (like when returning variable children). Original low-level errors and applied filtering is logged in verbose mode for future diagnostics.
* As part of that, I introduced new error classes derived from Error. This reduces the number of string compares in error handling and hopefully improves maintainability.
* Some minor issues have been sneaked into this PR. They are IMO not worth the overhead of a new PR.
* ~Side note: I believe the discard of errors after disconnect fix some of our flaky CI tests.~ Just saw that test fail again 😞 

Note: I fully appreciate that this approach might be controversial. I had loads of discussions on various debug solutions in the past about accurate error reporting vs usability. Hence, I am open to make this optional by a launch config setting, for example letting the user chose between a "normal" error handling mode and a "strict" mode. But would only do on demand.

**Testing**
* Tests for behavior after disconnect/terminate are added.
* Tests for race conditions between steps and variables requests are hard to recreate in E2E tests. Hence, not adding tests yet. Another one for a bigger variables rework.
  * They'd require modeling the scenarios where a GDB MI commands for variables and step requests are getting interleaved.
  * They'd require modeling where a variables request passes the getting the internal var handle before the continued event clears the handle cache.

### DIFF
--- a/src/integration-tests/launchRemote.spec.ts
+++ b/src/integration-tests/launchRemote.spec.ts
@@ -15,7 +15,12 @@ import {
     TargetLaunchArguments,
 } from '../types/session';
 import { CdtDebugClient } from './debugClient';
-import { fillDefaults, standardBeforeEach, testProgramsDir } from './utils';
+import {
+    fillDefaults,
+    gdbAsync,
+    standardBeforeEach,
+    testProgramsDir,
+} from './utils';
 import { expect } from 'chai';
 import * as os from 'os';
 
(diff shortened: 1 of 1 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -201,6 +202,47 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+     * implementation last.
+     * @param error The error to check.
+     * @return true if the error should be reported as a debug adapter protocol
+     * error response, false otherwise.
+     */
+    protected shouldReportError(error: unknown): boolean {
+        // GDBError occurrences for the main GDB backend.
+        if (error instanceof GDBError && error.backend === '') {
+            // Thread in GDB is running.
+            const threadRunningRegexp = /Selected thread is running/i; // Also handles messages that contain this starting with lower case 's'
```
jreineckearm: String compare probably should be moved to `GDBBackend` which then sends a not yet implemented `GDBThreadRunning` error.

jonahgraham (approved):
I'm ok with the overall approach and I have looked through some of the details of how you have implemented it. I have not done a line by line review though. Let me know if you want another set of eyes on those details.

> Hence, I am open to make this optional by a launch config setting, for example letting the user chose between a "normal" error handling mode and a "strict" mode.

I would be against this extra level of configuration personally. The "strict" mode would require users to have a fairly in depth understanding of GDB MI and probably DAP too. I think this case can and should be covered by the existing verbose logging alone.

PR #470, opened 2025-11-20
Add thread and frame params for assigning variables

tam18902:
Related Issue #469

PR #477, opened 2025-12-15
Run linting during CI (and other small cleanups)

cwalther:
Is there a reason we don’t run `yarn lint` during PR CI? Let’s see if it works. Because if there is no reason, then I am going to add it, otherwise the code style tends to stray occasionally, as right now.

Clean up that, as well as another typo from the same commit (#470).

Also, while I am messing around in this area, let’s run tests in a mode that catches accidentally committed `.only(` (see https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/415#issuecomment-3138937678).


jreineckearm (approved):
Thanks, LGTM!

cwalther:
Is there a strict policy on squashing/rebasing/non-ff-merging pull request branches? From what I have seen in this project so far, it seems like they are always squashed (so that’s what I did here too).

Personally, I am not a fan of doing that unconditionally, because sometimes the separation into multiple commits contains important information that is lost in squashing. In the case of this PR, I would probably have left the commits separate because they address independent concerns (but rebased on the target branch to keep a linear history). But I am okay with squashing too, because these commits touch disjoint sets of files, so the squashed commit is easy to mentally take apart again. On some of my previous PRs however I felt that squashing my carefully crafted commit sequence into one humongous commit lost something useful.

I realize that squashing is important to get rid of noisy “fixup” commits and exploratory dead-ends that need no historical record. But that’s something I do on my end before even submitting a PR, and in my own projects I would also ask a contributor to do it on their end before I would consider a PR ready for merging (or even for reviewing).

I am fine with deviating from my personal preferences though if there is a project policy.

jonahgraham:
I think there are only personal opinions on that within this project. I personally object to merge commits that don't have a very good reason. I think "fixups" should absolutely be squashed, and when there is logical sequence of commits they should be rebased (and not squashed).

As I do a lot of git history reading, I really appreciate logical commits with good quality summaries and messages.

I previously objected to some PRs that were not squashed because they were fixups commits. Perhaps my objection was taken with too much weight and lead to overly squashed PRs.

At the moment only rebased and squash are available in the UI IIRC. If there is a real need to add a merge on applying a pr, then the otterdog config needs to be temporarily updated.

FWIW, on other projects, such as Eclipse Platform only rebase is allowed, and contributors are expected to do any needed squashing before final review and approval.

Issue #478, opened 2025-12-18
Flaky multithread test


cwalther:
Moving this from https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/431#issuecomment-3665987397 so it doesn’t get lost:

The _multithread ▸ sees all threads_ test has been flaky lately (e.g. in [this run](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/20304707558/job/58326049530), but also seen locally on Linux), and it seems like this is due to a bug in the test, introduced in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/452: apparently the check in
https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/8d051fd40e0de770c94b3b722f8ca68eba9cf65e/src/integration-tests/multithread.spec.ts#L177-L184
is mistakenly finding the new `thread_id` variable of `inner_method`, so the heuristic is no longer correct.

PR #479, opened 2025-12-18
Adapt multithread test for all the places that the test program could stop

jonahgraham:
As MultiThread.cpp was added to, the 'sees all threads' wasn't updated to account for all the theoretical places that the program could stop. This commit supports all the current stopped locations of the secondary threads.

Fixes https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/478


cwalther:
LGTM, thanks for the quick reaction!

PR #482, opened 2026-01-07
feat(updateThreadInfo): add option to disable update thread info

### LINKED ISSUE #436: [feature]: Add serial option in `gdbtarget`
I would like to use this extension for `kgdbc`. First step is to set the baudrate which is basically not possible due to late initialization (discussion ongoing in #432 ). So maybe would be nice to have options like:
```json
"target":{
    "parameters": {
        "type": "tcp" # or serial
         "args": [
         ]
    },
    "serial": {
        "device": "/dev/ttyX",
        "baudrate": "115200"
    }
}
```
This will allow to directly select the type and custom commands like `set serial baud 115200` won't be needed.
Next idea, while debugging the kernel on embedded systems should be fine to have an option to stop refreshing the threads. With this baudrate the response is so slow there is basically no way to use it. On the other hand, if I comment out the thread request like:
```ts
try {
    /*if (!this.isRunning) {
    const result = await mi.sendThreadInfoRequest(this.gdb, {});
    this.threads = result.threads
           .map((thread) => this.convertThread(thread))
           .sort((a, b) => a.id - b.id);
    }*/

    response.body = {
        threads: this.threads,
    };

    this.sendResponse(response);
```
system is more-less responsible and could be used for this purposes.

wdty?

andy9a9:
In very slow systems using eg. serial-console for remote debugging refreshing the threads makes the debugging unusable. On the other hand when the gdb hits a bp, the system is more-less responsible then getting the threads is disabled.

Closes #436

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -160,6 +160,9 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected threads: ThreadWithStatus[] = [];
     protected missingThreadNames = false;
 
+    protected updateThreadInfo: 'missing' | 'when-requested' | 'never' =
+        'missing';
+
     /**
      * State variables for pauseIfNeeded/continueIfNeeded logic, mostly used for
      * temporary pause to insert breakpoints while the target is running.
@@ -313,10 +316,19 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
      * @param args the request arguments to validate.
      */
     protected validateRequestArguments(
-        _args: LaunchRequestArguments | AttachRequestArguments
+        args: LaunchRequestArguments | AttachRequestArguments
     ) {
-        // No validation for 'gdb' debugger type. Derived classes may override.
-        // Added to base class for consistency.
+        if (
+            args.updateThreadInfo !== undefined &&
+            args.updateThreadInfo !== 'missing' &&
+            args.updateThreadInfo !== 'when-requested' &&
+            args.updateThreadInfo !== 'never'
+        ) {
+            throw new Error(
+                `Invalid value for 'updateThreadInfo': '${args.updateThreadInfo}'. ` +
+                    "Valid values are: 'missing', 'when-requested', or 'never'."
+            );
+        }
     }
 
     /**
@@ -329,6 +341,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.customResetCommands = args.customResetCommands;
         this.steppingResponseTimeout =
             args.steppingResponseTimeout ?? DEFAULT_STEPPING_RESPONSE_TIMEOUT;
+        this.updateThreadInfo = args.updateThreadInfo ?? 'missing';
     }
 
     /**
@@ -479,6 +492,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     ) {
         this.validateRequestArguments(args);
         await this.setupCommonLoggerAndBackends(args);
+        this.initializeSessionArguments(args);
 
         await this.spawn(args);
         if (request == 'launch') {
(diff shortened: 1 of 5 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -1562,16 +1578,26 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
             // code (remote.c remote_target::putpkt_binary) says otherwise.
-            // The effect in situations excluded by this logic is that missing
-            // thread names are not filled in until the client issues this
-            // request at a time when the target is stopped.
+            // The pollThreadInfo setting controls when thread info is retrieved:
+            // - 'missing' (default): Thread info is retrieved when target is stopped
+            //   or when thread names are missing and retrieval is allowed while running
+            // - 'when-requested': Thread info is retrieved when client requests it (when stopped
+            //   or when allowed while running)
+            // - 'never': Thread info is never retrieved, cached data is used
```
cwalther: Nitpicking:
- leftover `pollThreadInfo`
- items should be consistently ordered from “least often” to “most often” (or the other way around)
- “cached data is used” is wrong, with `never` there never is any data that could be cached
- clarify `when-requested` because `missing` also only applies when requested

```suggestion
            // The updateThreadInfo setting controls when thread info is retrieved:
            // - 'never': Thread info is never retrieved, no thread names will be displayed
            // - 'missing' (default): Thread info is retrieved when target is stopped
            //   or when thread names are missing and retrieval is allowed while running
            // - 'when-requested': Thread info is retrieved every time the client requests
            //   it, as far as possible (when stopped or when allowed while running)
```

In the documentation, we will need to spell this out in more detail and also include examples of which setting one would want to choose in which situation.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -477,6 +492,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     ) {
         this.validateRequestArguments(args);
         await this.setupCommonLoggerAndBackends(args);
+        this.initializeSessionArguments(args);
```
jreineckearm: This also enables use of `customResetCommands` and `steppingResponseTimeout` which certainly hasn't been tested with `gdb` type. Don't expect problems in reality though because average users wouldn't be pointed to this in docs or launch.json autocompletion.
Rather for awareness of the side-effect it has.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -311,10 +314,21 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
      * @param args the request arguments to validate.
      */
     protected validateRequestArguments(
-        _args: LaunchRequestArguments | AttachRequestArguments
+        args: LaunchRequestArguments | AttachRequestArguments
     ) {
         // No validation for 'gdb' debugger type. Derived classes may override.
         // Added to base class for consistency.
```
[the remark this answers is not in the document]
andy9a9: applied

cwalther (approved):
Looking good now, thanks! I have also tested the distinction between `missing` and `when-requested` now, works as expected.

I have now merged the other PR that was in flight, so if you rebase again, you can add the changelog entry without causing conflicts. Could you please do that, @andy9a9? Afterward, this is good to merge.

Then, it would be appreciated if you could open a PR on [cdt-gdb-vscode](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode) adding the new option to README.md and package.json and CHANGELOG.md (see [the latest commit](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/commit/bed4615b19b26bc6ae1e80717c359ca6bb2fa6de) for an example). I think the documentation should, in addition to explaining the conditions similarly to the comment in the source code, mention something along the lines of that `never` is good for slow links or targets that have no thread names or are single-threaded, `missing` for targets whose thread names/details don’t change, `when-requested` when thread names/details do change. (It’s probably not sufficient to avoid a stale display for the latter, but that’s for later to examine and fix, no need to mention.) I’m not sure whether that all better goes into the table or into a separate paragraph.

Issue #485, opened 2026-01-15
Read Memory request error message is not comprehensive enough


omarArm:
The error message of readMemory request just has the statement "Unable to read memory."
It would be a better idea if the memory reference where the error happens is mentioned in the message

PR #486, opened 2026-01-15
Adding address to error message of read memory request

omarArm:
Adding memory reference to the error message of readMemory request https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/485

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2620,6 +2620,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     addr += ` offset: ${args.offset}`;
                 }
             }
+            addr += ')';
             const errorMessage =
                 err instanceof Error ? err.message + addr : String(err) + addr;
             if (!this.shouldReportError(err)) {


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2610,8 +2610,11 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.sendResponse(response);
             }
         } catch (err) {
+            const addrInHex = Number(args.memoryReference).toString(16);
```
jreineckearm: `memoryReference` can be an expression. So, you may need to handle both cases number and expression string.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2610,8 +2610,11 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
             const errorMessage =
-                err instanceof Error ? err.message : String(err);
+                err instanceof Error
```
jreineckearm: Nit: if you slightly rearrange the code, you get away with writing the ` (at address....` string once. Makes it easier to modify the display value if needed in future.

jreineckearm (commented):
I do support this change, but it needs some adjustments.
- See comments.
- `offset`, if it exists, is added to the resolved `memoryReference` address, hence should be displayed, too.
- In my experience, `count` might also make a difference in an access'es success. For example if starting an access at the boundary of a valid address range, i.e. if the access reaches into an inaccessible address range.

Also, can we please add tests for this change?

cwalther:
I can’t even figure out how to test this. So, unless somebody wants to explain that to me, and it’s even possible in my setup, I will leave the review to Jens. How do I make Visual Studio Code send such requests? All I can find is the possibility to look at a local variable in a hex editor, and it’s hard to get invalid addresses that way. How do I look at arbitrary memory addresses?

There seem to be two methods reading memory in `GBDDebugSessionBase`, `memoryRequest()` and `readMemoryRequest()`. Would it make sense to have this in both? (Do they even need to be separate?)

jreineckearm:
Thanks for having a look, @cwalther . No worries, you can leave this review to me.

> I can’t even figure out how to test this.

You should be able to provoke this with custom views that make use of `readMemory`. We saw this while developing a new view for [vscode-cmsis-debugger](https://github.com/Open-CMSIS-Pack/vscode-cmsis-debugger) . But I think one should be able to reproduce with [vscode-memory-inspector](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector) and [vscode-peripheral-inspector](https://github.com/eclipse-cdt-cloud/vscode-peripheral-inspector), too, which both use `readMemory`.

> There seem to be two methods reading memory in GBDDebugSessionBase, memoryRequest() and readMemoryRequest(). Would it make sense to have this in both? (Do they even need to be separate?)

Thanks for highlighting that, need to have another look. I believe `memoryRequest` was a path specific a custom DAP extension for [cdt-amalgamator](https://github.com/eclipse-cdt-cloud/cdt-amalgamator) memory requests with routing to its sub-sessions. But IIRC, amalgamator may rather be removed going forward. So, probably not worth to spend the effort.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2667,8 +2610,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.sendResponse(response);
             }
         } catch (err) {
+            // If address is an expression, then leave it as is. If it's a number, add offset andconvert to hex.
+            const addr =
+                ' at address ' +
```
jreineckearm: This looks a bit odd after the error message ending with a full-stop.
Maybe change to something like `(@<address>)` and `(@<memoryReference>, offset:<offset>)` depending on what's available in the arguments.

jreineckearm (approved):
Thanks, looks good now! Tested a couple of valid and invalid combinations with Memory Inspector. Working as expected now.

PR #496, opened 2026-02-23
Support target extended-remote launch connecting to gdbserver started…

fzuuzf:
… with the --multi option

Patch is a 'works for me' solution to
1) Manually start 'gdbserver --multi :2345' on the remote target. 2) Debug the program from vscode by the following launch entry:

```
{
    "name": "Launch on remote target, which has gdbserver --multi :2345 running",
    "type": "gdbtarget",
    "request": "launch",
    "program": "${workspaceFolder}/out/build/x64_clang++/program",
    "gdbNonStop": true,
    "openGdbConsole": true,
    "target": {
	"type": "extended-remote",
	"parameters": [
		"192.168.22.1:2345"
	]
    }
    "initCommands": [
	"set remote exec-file /target/path/to/program"
    ]
}
```

### DIFF
--- a/src/desktop/GDBTargetDebugSession.ts
+++ b/src/desktop/GDBTargetDebugSession.ts
@@ -265,11 +265,15 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         await this.setupCommonLoggerAndBackends(args);
         this.initializeSessionArguments(args);
 
+        this.isAttach = true;
         if (request === 'launch') {
-            const launchArgs = args as TargetLaunchRequestArguments;
-            await this.startGDBServer(launchArgs);
+            if (args.target?.type !== 'extended-remote') {
+                const launchArgs = args as TargetLaunchRequestArguments;
+                await this.startGDBServer(launchArgs);
+            } else
+                this.isAttach = false;
         }
-        await this.startGDBAndAttachToTarget(response, args);
+        await this.startGDBAndConnectToGDBSERVER(response, args);
     }
 
     protected async launchRequest(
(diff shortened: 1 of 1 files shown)


jreineckearm:
Thanks for the creating the PR, @fzuuzf .
From the description here and due to the lack of a separate [issue](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues), it is not clear to me in what way it didn't work for you before.
Have you tried `"request": "attach"` instead? That's how you normally attach to an already running GDB server like in your scenario. See also the corresponding extension [docs](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode?tab=readme-ov-file#launch-configurations).

cwalther:
I don’t know yet what the proper solution is, but I am also not convinced yet that it is this. There seems to be some overloading of what `"request": "launch"` means. In the context of `"type": "gdbtarget"`, as far as I understand, it has up to now meant “launch `gdbserver` (or some other stub)”, not “launch the program”. In your case, that is clearly a problematic choice. However, changing it now, but only in one particular case, as you do here, seems confusing and potentially breaking existing configurations.

Maybe we need a new configuration option for this case, to avoid disturbing existing configurations?

It probably adds to the confusion that we also seem to be a bit sloppy in our terminology regarding “attach”. Strictly speaking, as far as I understand, “attach” is what happens between stub and program. What happens between GDB and stub is called “connect”.

For what it’s worth, we (Indel) use `target extended-remote` for attaching to an embedded system (but not in `target.type`, but in an internal command invoked from `target.connectCommands`), but I don’t know what the historical reasons for that were (it seems like it shouldn’t make a difference).

cwalther:
Thanks for the ping. I still think this probably needs a new option to prevent breaking existing configurations, following either Jens’ suggestion or mine (or a different one entirely), unless all major downstream adopters of the adapter come forward and say that their use cases are not affected.

As far as we at Indel are concerned, I think the current proposal would not affect us, as we are never in the `request === 'launch'` case (even when `args.request === 'launch'`).

Contrary to what I stated above, we do now (since #548) use the exclusive features of `extended-remote`, but still only for late attaching, not launching on the remote side.

PR #506, opened 2026-03-09
Support value formatting

omarArm:
Should solve https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/210

# New implementation:

1. Updated the function `sendVarSetFormatToHex` to make it more generic to different value formats. Also changed the return type from void to string, to retrieve the actual value.
2. In the `evaluate` request, added support for setting value format for expressions
3. Added a new property `displayFormat` to `VarObjType`

Now users can provide format specifiers to their expressions. It can be 

```
,x  -> hex
,o  -> octal
,b -> binary
,d -> decimal
,z -> zero-hexadecimal
``` 

# Old implementation:

Supporting value formatting in evaluate requests.

1. Updated function sendVarSetFormatToHex to return string that includes updated value in hex instead of void
5. Checking on `format` property of evaluate request arguments before sending result

# Notes

- the feature only works with primitive data types so far

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2462,7 +2462,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.sendErrorResponse(
                     response,
                     1,
-                    'Unknown expression format specifier, supported specifiers are: x, d, o, b, z'
+                    'Unknown expression format specifier, supported specifiers are: x, d, o, b, t, z'
                 );
                 return;
             }
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2093,7 +2093,6 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     false,
                     false,
                     varCreateResponse,
-                    'hexadecimal',
                     ref.type
                 );
                 await mi.sendVarSetFormat(
@@ -2468,14 +2467,6 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     frameRef,
                     frame: frameRef ? 'current' : 'floating',
                 });
-                // if format is undefined, skip the format command and return the value as-is, otherwise apply the requested formatting)
-                if (expressionDisplayFormat !== 'natural') {
-                    varCreateResponse.value = await mi.sendVarSetFormat(
-                        gdb,
-                        varCreateResponse.name,
-                        expressionDisplayFormat
-                    );
-                }
                 varobj = gdb.varManager.addVar(
                     frameRef,
                     depth,
@@ -2483,18 +2474,8 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     false,
                     false,
                     varCreateResponse,
-                    expressionDisplayFormat
                 );
             } else {
-                const currentFormat = varobj.displayFormat;
-                if (currentFormat !== expressionDisplayFormat) {
-                    varobj.value = await mi.sendVarSetFormat(
-                        gdb,
-                        varobj.varname,
-                        expressionDisplayFormat
-                    );
-                    varobj.displayFormat = expressionDisplayFormat;
-                }
                 const vup = await mi.sendVarUpdate(gdb, {
                     name: varobj.varname,
                 });
@@ -2517,31 +2498,32 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                             expression,
                             frameRef,
                         });
-                        if (expressionDisplayFormat !== 'natural') {
-                            await mi.sendVarSetFormat(
-                                gdb,
-                                varCreateResponse.name,
-                                expressionDisplayFormat
-                            );
-                        }
                         varobj = gdb.varManager.addVar(
                             frameRef,
                             depth,
                             expression,
                             false,
                             false,
-                            varCreateResponse,
-                            expressionDisplayFormat
+                            varCreateResponse
                         );
                     }
                 }
             }
             if (varobj) {
                 const frameHandle = args.frameId ?? -1;
+                const hasChildren = args.context === 'variables' && Number(varobj.numchild);
                 const result =
-                    args.context === 'variables' && Number(varobj.numchild)
+                    hasChildren
                         ? await this.getChildElements(varobj, frameHandle)
-                        : varobj.value;
+                        : expressionDisplayFormat !== 'natural' ? await mi.sendVarSetFormat(
+                              gdb,
+                              varobj.varname,
+                              expressionDisplayFormat
+                          ) : varobj.value;
+                if (expressionDisplayFormat !== 'natural') {
+                    // restore original format after evaluation
+                    await mi.sendVarSetFormat(gdb, varobj.varname, 'natural');
+                }
                 const variablesReference =
                     parseInt(varobj.numchild, 10) > 0
                         ? this.variableHandles.create({
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -14,6 +14,7 @@ import {
     BreakpointEvent,
     Handles,
     InitializedEvent,
+    InvalidatedEvent,
     Logger,
     logger,
     LoggingDebugSession,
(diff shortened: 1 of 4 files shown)


jonahgraham (commented):
In addition, please write a test or two to show requests with the format set.

src/mi/var.ts
```diff
@@ -69,6 +69,10 @@ export interface MIVarPathInfoResponse {
     path_expr: string;
 }
 
+export interface MIVarSetFormatToHexResponse {
```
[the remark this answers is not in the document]
omarArm: done

jreineckearm (commented):
Thanks for the updates, see some comments.

asimgunes (commented):
Minor comments about code quality.

cwalther (approved):
I am not familiar enough with these code areas for a thorough architectural review, but in the details it looks sensible now.

In some anecdotal testing on an Indel target, it appears to work.

I notice that when I make a watch expression with a local variable and a format specifier, that also affects the display of that variable in the _Variables_ view. Is that intentional? Also, when I have two watch expressions with the same expression, but different format specifiers, they affect each other – that is probably not intentional. I think being able to see the same value in two different formats would sometimes be useful.

jreineckearm (commented):
Given feedback makes a lot of sense, thanks to everyone helping here.
Some of the feedback should probably go into a new PR to keep the changes better reviewable. Omar is working on some final changes for this one here.

I am away until Thursday next week. Can please one of the other committers help with merging after Omar did these changes and while I am away? Thanks

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2389,23 +2410,56 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                    response,
+                    1,
+                    'Unknown expression format specifier, supported specifiers are: x, d, o, b, z'
+                );
+                return;
+            }
+            if (expressionDisplayFormat !== 'natural') {
+                // Slice off the format specifier to get the actual expression to evaluate, the format will be applied later via var-set-format
+                expression = expression
+                    .slice(0, expression.lastIndexOf(','))
```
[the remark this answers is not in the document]

cwalther (approved):
I am fine with merging this as-is (squashed and rebased) and addressing the crosstalk separately, but I’ll defer to @jonahgraham.

jonahgraham:
I see you have continued to work on this - I just realized I never commented in response to https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/506#issuecomment-4175668136 - but since the current implementation can show incorrect data I don't think we should merge until resolved. I see @omarArm has added new commits since with title "eliminating interference between same expressions with different valu…" so it sounds like that is the way things are progressing anyway.

omarArm:
@jonahgraham yes, we came to the same conclusion that we should not merge without fixing the interference issue. After some investigation, I opted for the current implementation.

jonahgraham (commented):
This LGTM - I tried it out and had a look. I know you are still polishing and resolving test failures, but besides that we should be good to go.

cwalther:
Would using `-var-evaluate-expression -f <format>` instead of toggling the format achieve the same thing, but save one MI call?

omarArm:
@cwalther nice catch. I must have missed this command while reading the documentation. I tried it and it worked

src/mi/var.ts
```diff
@@ -189,14 +206,20 @@ export function sendVarAssign(
...
-    const command = `-var-evaluate-expression ${params.varname}`;
-    return gdb.sendCommand(command);
+): Promise<string> {
+    let command = `-var-evaluate-expression`;
+    if (params.format) {
+        command += ` -f ${params.format}`;
+    }
+    command += ` ${params.varname}`;
+    const result: MIVarEvalResponse = await gdb.sendCommand(command);
+    return result.value;
```
cwalther: Is `sendVarEvaluateExpression()` considered public API? If so, changing its return value breaks clients.

In this case, I would rather not change it, to be safe, since the alternative looks simple (even simpler than changing it, I would argue): replace `await mi.sendVarEvaluateExpression(…)` by `(await mi.sendVarEvaluateExpression(…)).value` at the call site.

src/integration-tests/evaluate.spec.ts
```diff
@@ -13,7 +13,7 @@ import { CdtDebugClient } from './debugClient';
 import {
     expectRejection,
     fillDefaults,
-    gdbAsync,
+    //gdbAsync,
```
cwalther: One more leftover comment.

Also, I think the commits that touch this should be squashed separately from the rest, because that is not really related to this PR, but rather an amendment of #523. (Making a separate PR for it is probably overkill, but if you want I can merge it.)

src/mi/var.ts
```diff
@@ -189,14 +206,20 @@ export function sendVarAssign(
     return gdb.sendCommand(command);
 }
 
-export function sendVarEvaluateExpression(
+export async function sendVarEvaluateExpression(
```
cwalther: Nitpick: Making the function async now seems like an unnecessary complication.

cwalther (changes requested):

src/mi/var.ts
omarArm: it should have been async from the beginning in my opinion. The function sends a command to gdb and waits for it's return value.

src/integration-tests/evaluate.spec.ts
omarArm: Done. I will squash before merging then

src/mi/var.ts
cwalther: OK, as you wish.

cwalther (approved):
Code looks good to me now, thanks for the patience! (You could have left the async, if you prefer it that way, my comments are not meant as orders.) I would like to give it another test, but that will have to wait until Monday. If you would prefer to have it merged before the weekend, I am fine with @jreineckearm or anyone else merging.

jreineckearm:
Thanks for the review, @cwalther .

We had discussed offline why the `async` isn't needed anymore and that it worst case can add a tiny overhead at runtime (certainly not critical here). I am OK to wait until Monday, I just would like us to get done with it. Still have a full backlog on our main CMSIS Debugger project.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2414,7 +2456,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
             const [gdb, frameRef, depth] =
                 await this.getFrameContext(initialFrameRef);
-
+            const [actualExpression, expressionDisplayFormat] =
+                this.extractExpressionFormat(expression);
+            if (expressionDisplayFormat === 'unknown') {
+                this.sendErrorResponse(
+                    response,
+                    1,
+                    'Unknown expression format specifier, supported specifiers are: x, d, o, b, z'
```
cwalther:
```suggestion
                    'Unknown expression format specifier, supported specifiers are: x, d, o, b, t, z'
```

cwalther (commented):
Tested, looks good.

I assume @jreineckearm will merge once you are done tidying up commits; if not, ping me if you want me to do it.

src/gdb/GDBDebugSessionBase.ts

jreineckearm (approved):
Thanks, LGTM! Gave it a try, and it took me a while to realize I shouldn't add a space between `,` and `x` in `,x` to make it work. ;-)
Thanks, @cwalther , for your support on this one!
Merging

cwalther:
But please get the commits into a clean shape before merging. (As mentioned I would prefer two commits.)

cwalther:
You squashed it into one. Oh well, not a big deal.

jreineckearm:
> But please get the commits into a clean shape before merging. (As mentioned I would prefer two commits.)

Sorry, forgot about it.... let me know if you'd like me to revert and split up.

cwalther:
Do you have push access to the main branch? Last time I tried, I didn’t, the only write access I had was using the GitHub web UI to merge PRs.

But anyway, don’t bother if you have more important things to do. The piece that doesn’t belong here is small and not important.

jreineckearm:
I am afraid not, you should be able to see branch protection rules here: https://otterdog.eclipse.org/projects/ecd.cdt-cloud/repos/cdt-gdb-adapter#branch-protection-rules
I believe the important bit is `requires_pull_request`.

> The piece that doesn’t belong here is small and not important.

OK, sorry again. Had completely forgotten about it over the weekend and only saw your comment the very second that I had pushed squash and merge after cleaning up the commit message.....

cwalther:
No worries.

PR #512, opened 2026-03-24
Remove redundant sendVarDelete call that throws an error

duyhihi91:
_this is the updated description and summary, as this change was progressed the design and implementation changed substantially. See collapsed section below for original description._

This sendVarDelete is rarely reachable, generally only reachable when the same variable name is used in two different scopes within the same method. See use of `x` variable in
src/integration-tests/test-programs/watch_local_scope_transition.c for an example.

When debugging such code, and the `x` was in the watch window, the now removed sendVarDelete would run, but the var was already deleted by the varManager.removeVar on the line above. So the sendVarDelete would generate an exception, causing the `x` not to be evaluated.

Included in this change is a test that demonstrates the use case, the new test `evaluate request - watch local variable across lexical scope transition` in evaluate.spec.ts.

<details><summary>Original message</summary>
<p>

Summary: Improvement/wrap send var delete in try/catch after remove var in do evaluate request

*For ECA reasons, this is a continuation of https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/509*

**Background:** removeVar(...) called sendVarDelete(...) inside itself.
In case of doEvaluateRequest(...) method, it will throw "object not found" error when calling sendVarDelete(...) after removeVar(...).

**Solution:** add try/catch to safely wrap the sendVarDelete(...) to swallow the "object not found" error.
Other errors still be catch in catch (err) block in the end of doEvaluateRequest(...) as currently.

**Reason of NOT removing sendVarDelete(...) in doEvaluateRequest(...):** 
1/ In future, if a person removes sendVarDelete(...) inside removeVar(...) -> sendVarDelete(...) still need to be called after removeVar(...) - same as current development.
2/ Why I did not remove sendVarDelete(...) inside removeVar(...) : because removeVar(...) is used at other place , also maybe used in third party code. If I change removeVar(...) it will impact their implementation.

</p>
</details>

### DIFF
--- a/src/integration-tests/doEvaluateRequest.spec.ts
+++ b/src/integration-tests/doEvaluateRequest.spec.ts
@@ -1,0 +1,243 @@
+/*********************************************************************
+ * Copyright (c) 2026 Arm Limited and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *********************************************************************/
+
+import 'mocha';
+import { expect } from 'chai';
+
+import * as sinon from 'sinon';
+import { DebugProtocol } from '@vscode/debugprotocol';
+
+import { GDBDebugSessionBase } from '../gdb/GDBDebugSessionBase';
+import * as miVar from '../mi/var';
+
+class TestableSession extends GDBDebugSessionBase {
+    constructor(backendFactory: any) {
+        super(backendFactory);
+    }
+
+    public async callDoEvaluateRequest(
+        response: DebugProtocol.EvaluateResponse,
+        args: DebugProtocol.EvaluateArguments,
+        alwaysAllowCliCommand: boolean
+    ) {
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        return (this as any).doEvaluateRequest(
+            response,
+            args,
+            alwaysAllowCliCommand
+        );
+    }
+}
+
+function makeEvaluateResponse(): DebugProtocol.EvaluateResponse {
+    return {
+        type: 'response',
+        seq: 0,
+        request_seq: 1,
+        success: true,
+        command: 'evaluate',
+        body: {} as any,
+    } as DebugProtocol.EvaluateResponse;
+}
+
+describe('doEvaluateRequest - MI var deletion single-owner invariant', function () {
+    let session: TestableSession;
+    let sandbox: sinon.SinonSandbox;
+
+    let gdb: any;
+    let varManager: {
+        getVar: sinon.SinonStub;
+        addVar: sinon.SinonStub;
+        removeVar: sinon.SinonStub;
+    };
+    let backendFactory: any;
+
+    beforeEach(function () {
+        sandbox = sinon.createSandbox();
+
+        varManager = {
+            getVar: sandbox.stub(),
+            addVar: sandbox.stub(),
+            removeVar: sandbox.stub(),
+        };
+
+        gdb = { varManager };
+
+        backendFactory = {
+            create: sandbox.stub().returns(gdb),
+            dispose: sandbox.stub(),
+        };
+
+        session = new TestableSession(backendFactory);
+        (session as any).gdb = gdb;
+
+        // Make frameId resolvable
+        (session as any).frameHandles = new Map<number, any>([
+            [111, { id: 'f1' }],
+            [222, { id: 'f1' }],
+            [333, { id: 'f1' }],
+            [444, { id: 'f1' }],
+        ]);
+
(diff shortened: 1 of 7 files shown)


src/integration-tests/doEvaluateRequest.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+        sandbox.stub(miVar, 'sendVarCreate');
+        sandbox.stub(miVar, 'sendVarUpdate');
+        sandbox.stub(miVar, 'sendVarDelete');
+    });
+
+    afterEach(function () {
+        sandbox.restore();
+    });
+
+    it('out-of-scope: removeVar đã xóa MI → sendVarDelete "object not found" → swallow & recreate OK', async function () {
```
jonahgraham: Can this be translated to English? `đã xóa MI` - same comment applies in numerous places in the code.

jonahgraham (commented):
Can you provide a way to reproduce this error? I still don't understand how users see this problem. It also seems that none of the tests cover these lines of code.

I have provided some line comments, but the key thing I don't understand is whether the sendVarDelete actually can succeed. And if it can never succeed, then why can't we just delete it.

src/integration-tests/evaluate.spec.ts
```diff
@@ -346,3 +346,108 @@ describe('evaluate request global variables', function () {
         });
     });
 });
+
+describe('evaluate request - watch local variable across lexical scope transition', function () {
```
jonahgraham, line 467: This test is exactly what I was hoping for. It is really easy to understand what is happening and what the underlying issue is.

src/integration-tests/evaluate.spec.ts
```diff
@@ -346,3 +346,108 @@ describe('evaluate request global variables', function () {
...
+    const program = path.join(testProgramsDir, 'watch_local_scope_transition');
+    const src = path.join(testProgramsDir, 'watch_local_scope_transition.c');
+    const lineTags = {
+        FIRST_SCOPE: 0,
+        SECOND_SCOPE: 0,
+    };
+
+    before(function () {
+        resolveLineTagLocations(src, lineTags);
+        console.log('FIRST_SCOPE line:', lineTags.FIRST_SCOPE);
```
jonahgraham: Please remove the console.log or place them behind a DEBUG guard of somesort.

jonahgraham (commented):
I was pinged offline about this and realized that I had not submitted my below review. To avoid further delays I will apply the removal of console.log myself and this can be merged.

jonahgraham:
Updated title and original description to be:

This sendVarDelete is rarely reachable, generally only reachable when the same variable name is used in two different scopes within the same method. See use of `x` variable in
src/integration-tests/test-programs/watch_local_scope_transition.c for an example.

When debugging such code, and the `x` was in the watch window, the now removed sendVarDelete would run, but the var was already deleted by the varManager.removeVar on the line above. So the sendVarDelete would generate an exception, causing the `x` not to be evaluated.

Included in this change is a test that demonstrates the use case, the new test `evaluate request - watch local variable across lexical scope transition` in evaluate.spec.ts.

src/integration-tests/evaluate.spec.ts
jonahgraham: It would be nice to have an easy way to log within a test and have it in the same log file as the backend. For now, rather that deleting these, I have put them behind a DEBUG guard.

jonahgraham (approved):

cwalther:
I just noticed that this introduced [an eslint warning](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/24407440353/job/71294558847#step:7:15):

```
/home/runner/work/cdt-gdb-adapter/cdt-gdb-adapter/src/integration-tests/doEvaluateRequest.spec.ts
  30:9  warning  Unused eslint-disable directive (no problems were reported from '@typescript-eslint/no-explicit-any')

✖ 1 problem (0 errors, 1 warning)
  0 errors and 1 warning potentially fixable with the `--fix` option.
```

Apparently warnings don’t cause the CI to fail (I must have missed that in #477). Can/should we change that? (@jonahgraham, @jreineckearm)

And @duyhihi91, was that directive needed for you? If not, could you submit another PR to remove it?

jreineckearm:
I normally prefer without warnings. But if we want to be stricter, then we should probably rather change the rules to make everything an error. That sets expectations right when you build and lint locally. But can also become massively annoying while making the changes. Let's rather spend time on getting linter warnings/errors shown as annotations in the code review web frontend. We have that on most of our repos. Just don't know from the top of my head how to get them there.

@duyhihi91 , I just also noticed that you added **_new_** test files with copyright headers saying `Arm Limited and others`. While I think it's strictly speaking not completely wrong to do that (because of the `and others` part), I'd normally prefer if **_new_** files carry the contributor's actual company name (I'd probably know if it was Arm Limited 😉 ). Sorry, that I didn't find the time to review this earlier. Otherwise I'd have raised it before merge. Would you mind fixing that?

PR #523, opened 2026-04-07
Send invalidate event on radix change

omarArm:
should fulfill https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/207

Background:
When changing the debugee's global radix, values don't immediately update on VSCode. As a workaround users had to step or change vscode state somehow.

Solution:
Whenever a global radix is set, GDB sends the following message via the console "Output radix now set to <radix>". By tapping on GDB messages to the system, the adapter would send an `invalidated` event to prompt clients to refresh the system.

### DIFF
--- a/src/integration-tests/evaluate.spec.ts
+++ b/src/integration-tests/evaluate.spec.ts
@@ -167,7 +167,7 @@ describe('evaluate request', function () {
         });
         expect(res2.body.result).eq('10');
     });
-    
+
     it('should be able to use GDB command', async function () {
         const res1 = await dc.evaluateRequest({
             context: 'repl',
@@ -196,11 +196,11 @@ describe('evaluate request', function () {
 
         expect(err.message).eq('Undefined MI command: a');
     });
-    
+
     it('should send invalidate event when changing global radix through evaluate request', async function () {
         if (os.platform() === 'win32' || !(isRemoteTest && gdbAsync)) {
-                    this.skip();
-                }
+            this.skip();
+        }
         const event = dc.waitForEvent('invalidated');
         await dc.evaluateRequest({
             context: 'repl',
--- a/src/integration-tests/evaluate.spec.ts
+++ b/src/integration-tests/evaluate.spec.ts
@@ -13,6 +13,7 @@ import { CdtDebugClient } from './debugClient';
 import {
     expectRejection,
     fillDefaults,
+    gdbAsync,
     getScopes,
     isRemoteTest,
     resolveLineTagLocations,
@@ -21,6 +22,7 @@ import {
     testProgramsDir,
 } from './utils';
 import { expect } from 'chai';
+import * as os from 'os';
 
 describe('evaluate request', function () {
     let dc: CdtDebugClient;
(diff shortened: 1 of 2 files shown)


cwalther (commented):
Preliminary comments:

Is there a reason why you are checking for output and not for the `GDB notify async: cmd-param-changed…` that I mentioned in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/202#issuecomment-3966597776? Relying on output intended to be human-readable generally feels like a last-resort measure. I have not tried if that works, but if it doesn’t, that may be worth a mention.

jreineckearm (approved):
Code changes look good to me, thanks!

This PR is massively useful to further test the global radix setting feature. Need to double check with an extension build from `main` tomorrow. But I had to enter `> set radix a` to switch back to a radix of 10 because it interpreted the value `10` as a hex input when entered. May need some fine-tuning there.

Keeping unmerged until tomorrow to give other reviewers the chance to check requested changes.

src/integration-tests/evaluate.spec.ts
```diff
@@ -192,6 +195,20 @@ describe('evaluate request', function () {
 
         expect(err.message).eq('Undefined MI command: a');
     });
+
+    it('should send invalidate event when changing global radix through evaluate request', async function () {
+        if (!(isRemoteTest && gdbAsync)) {
+            this.skip();
+        }
```
cwalther: In my experience on Linux, this restriction is not needed. The test passes in all 7 scenarios. I didn’t try Windows.

cwalther (approved):
Good to go as far as I am concerned. I tried it on an Indel target and it appears to work.

Regarding @jreineckearm’s observation of needing `> set radix a`: That appears to only be an issue if the _input_ radix is also set to hex. I see no reason to do that, because hex numbers can always be input using a 0x prefix. As long as the input radix is unchanged, the _Set Global Output Radix to Decimal_ command that uses `> set output-radix 10` works for switching from hex to dec for me, and manually entering `> set output-radix a` results in an error `No symbol "a" in current context.`. I therefore consider the inability to deal with a non-decimal input radix a minor issue that does not block merging this (and the fix would belong in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/202 and not here anyway). Empirically, it seems to be possible to enter decimal numbers with a hex input radix using a 0d prefix, so maybe that would be an easy fix, but I cannot find any mention of that in the GDB docs.

See my note about the test, not a blocker.

The incremental commits should be squashed into fewer before merging (probably one in this case), and if I were to do the merging, I would appreciate if you could do that yourself, to save me the work of coming up with commit message(s). If @jreineckearm wants to merge I leave that up to him.

src/integration-tests/evaluate.spec.ts
cwalther: Are you sure you tested it _with_ the fix of using 16 instead of 10 (because changing from 10 to 10 doesn’t generate an event) and _without_ the guard? Among the commits you pushed here, these came both in the same commit, so I can’t tell at a glance which of the two it was that fixed CI.

jonahgraham: See https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/214 - the test probably should also be unambiguously setting the radix, although we actually have more control in a test since we "know" the input radix will be 10 (decimal)

PR #532, opened 2026-05-07
v1.8.1 Release Preparations

### LINKED ISSUE #530: Publish Package to NPM job fails on tag push — npm self-upgrade crash
The publish job for `v1.8.0` has been failing in the `Update npm (OIDC/provenance support)` step. I retried it 2 times, same crash both times, so `v1.8.0` is tagged but never reached npm — `latest` on npmjs.org is still `v1.7.1`.

Failing run: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/actions/runs/25483568015/job/74782943208

### What's happening

The step `npm i -g npm@latest` crashes:

```
npm error code MODULE_NOT_FOUND
npm error Cannot find module 'promise-retry'
npm error Require stack:
npm error - .../npm/node_modules/@npmcli/arborist/lib/arborist/rebuild.js
...
```

Nothing changed on our side — `v1.7.1` published fine on 2026-03-31 with the exact same workflow. What changed is the GitHub-hosted runner's Node 22 toolchain, which has rolled to `22.22.2`. That version bundles npm `10.9.7`/`10.9.8` with a stripped `node_modules` tree that is missing `promise-retry`, so `arborist/rebuild.js` blows up when the running npm tries to self-upgrade. Tracked upstream as [npm/cli#9151](https://github.com/npm/cli/issues/9151).

One detail that confused me at first: the upstream fix [npm/cli#9152](https://github.com/npm/cli/pull/9152) is already merged. But the fix lives in the *new* npm being installed, not the *running* npm doing the upgrade — and the running 10.9.x crashes before the new code is on disk. So the fix only helps once Node 22 ships a re-bundled, fixed npm.

### What I'd like to do

Bump only the `Publish-to-NPM` job to Node 24:

```yaml
  Publish-to-NPM:
    ...
      - uses: actions/setup-node@v4
        with:
          node-version: '24'        # was '22'
          registry-url: 'https://registry.npmjs.org'
      - name: Update npm (OIDC/provenance support)
        run: npm i -g npm@latest    # unchanged
```

Node 24's bundled npm already includes #9152, so `npm i -g npm@latest` works without any workaround. One-line diff. CI stays on Node 22 — Build-on-Ubuntu, Build-on-Windows, the integration test matrix and native PTY tests are not touched.

I went through the publish job and it doesn't need anything Node-version-specific. It runs `yarn` (→ `prepublish: yarn build` → `tsc` + `esbuild`) and `npm publish`. No tests, no `nativebuild` at publish time. The published tarball is the JS that tsc/esbuild produced plus the `.cc/.h` sources, and consumers compile the addon on their own machine via `install.js`. There's no `engines` field in `package.json` that ties us to a specific Node major.

### Fallback, if the Node 24 bump on publish is too much

Pin npm on Node 22:

```yaml
      - name: Update npm (OIDC/provenance support)
        # Workaround for npm/cli#9151 — npm@latest can't self-upgrade
        # from the npm bundled with Node 22.22.2.
        run: npm i -g npm@11.11.1
```

Comments on #9151 confirm `11.11.1` installs cleanly from the broken 10.9.x baseline, and it's recent enough for `--provenance` later if we wire that up. The downside is that it leans on benign file-removal ordering inside the broken npm, so a future runner update could shift timings — we'd want to revisit when Node 22.x ships a fixed bundled npm.

### Things I considered but didn't pursue

- Bumping CI to Node 24 as well — cleanest in spirit, but it would put the full integration test matrix (GDB breakpoints, async/non-stop, native PTY, serialport native builds on Ubuntu + Windows) on a different Node major. Worth a separate discussion, not part of this unblock.
- Two-step incremental upgrade (`npm@11.11.1 && npm@latest`) — just stacks workarounds.
- Removing the upgrade step entirely — quickest unblock but loses the OIDC/provenance prep added in #502.

### Asking

- OK to bump just the publish job to Node 24, or would you prefer the npm pin?

Plan once the fix lands: cut `v1.8.1` from the same commit as `v1.8.0` (rather than re-tag).

asimgunes:
- Fixes [`#530`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/530): Bump publish job to Node 24 to work around an npm self-upgrade crash that prevented 1.8.0 from being published to npm (workflow fix landed in #531).

There are no code changes from 1.8.0 — this release is solely to re-publish 1.8.0 to npm under the 1.8.1 version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change Log
 
+## 1.8.1
+
+- Fixes [`#530`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/530): Bump publish job to Node 24 to work around an npm self-upgrade crash that prevented 1.8.0 from being published to npm.
+
 ## 1.8.0
 
 - Implements [`#506`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/506): Support value formatting in expressions with `,<type>` (where `<type>` is one of `x` (hex), `d` (dec), `o` (oct), `t`/`b` (bin), `z` (zero-padded hex)).
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cdt-gdb-adapter",
-  "version": "1.8.0",
+  "version": "1.8.1",
   "description": "gdb adapter implementing the debug adapter protocol",
   "main": "dist/index.js",
   "browser": "dist/browser/web.js",


jreineckearm (approved):
LGTM!

jreineckearm:
Thanks for making the release, @asimgunes ! Please reach out to @omarArm for review if you'd also like to make a `cdt-gdb-vscode` release. I am away until 18.05.

Issue #526, opened 2026-04-24
Handle `memory-changed` notifications


jreineckearm:
### Problem

CDT GDB Adapter does not handle `memory-changed` notifications yet.
See bottom of https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Async-Records.html

This leads to warnings that the support is missing.

### Reproducer

Can be reproduced in my case with `gdbtarget` type, an embedded target that has valid memory or memory mapped peripherals at a certain address,  and CLI commands like this (just change the address as needed):
```
> set {unsigned int}0xE0000E40 = 0x0000000F
GDB unhandled notify: memory-changed: {"thread-group":"i1","addr":"0xe0000e40","len":"0x4"}
```

This can become noisy when sourcing GDB scripts that for example configure some peripherals in the target system.

### Proposal

Proposed solution is to handle those notifications and map them to `Memory` events of the DAP protocol: https://microsoft.github.io/debug-adapter-protocol/specification#Events_Memory . Exact mapping strategy needs to be figured out while working on this. (`addr` and `len` should be straight forward, don't know without further investigation what to do with `thread-group`...if anything is needed at all.)

### Testing

If I interpret this [code](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/cbfcb9e59aa55cb3953808fca2dbd2db37b585f7/src/plugin/session-tracker.ts#L176) and this [code](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/blob/cbfcb9e59aa55cb3953808fca2dbd2db37b585f7/src/webview/memory-webview-view.tsx#L207) correctly, then the [Memory Inspector](https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/tree/main) would listen for such events and refresh when affected memory is displayed. This could be used for more thorough testing.

I don't know if VS Code/Theia would refresh any built-in views on `Memory` events.

The most basic way of testing is to inspect the verbose GDB adapter output on the Debug Console.

We have a strong interest in getting this resolved, hence are happy to pick up ourselves.
If however someone else has a strong interested in picking this up (I expect it to be a good starter issue), then please comment here so that we don't duplicate efforts.

omarArm:
Done in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/533

PR #533, opened 2026-05-08
Handle memory notifications

omarArm:
Should solve https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/526

Handling GDB memory notifications by sending a DAP memory event.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -3169,12 +3169,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.handleCmdParamChanged(notifyData);
                 break;
             case 'memory-changed': {
-                const length = isNaN(notifyData.len) ? 0 : parseInt(notifyData.len, 10);
-                const memoryEvent = new MemoryEvent(
-                    notifyData.addr,
-                    0,
-                    length
-                );
+                const length = isNaN(notifyData.len)
+                    ? 0
+                    : parseInt(notifyData.len, 10);
+                const memoryEvent = new MemoryEvent(notifyData.addr, 0, length);
                 this.sendEvent(memoryEvent);
                 break;
             }
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -3168,14 +3168,15 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             case 'cmd-param-changed':
                 this.handleCmdParamChanged(notifyData);
                 break;
-            case 'memory-changed':
+            case 'memory-changed': {
                 const memoryEvent = new MemoryEvent(
                     notifyData.addr,
                     0,
                     parseInt(notifyData.len)
                 );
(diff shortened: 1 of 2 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -3167,6 +3168,14 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             case 'cmd-param-changed':
                 this.handleCmdParamChanged(notifyData);
                 break;
+            case 'memory-changed': {
```
asimgunes, line 3171: We might need to set the `supportMemoryEvent` in the `initializeRequest` method to pass that memory events are supported according to the specification:

https://microsoft.github.io/debug-adapter-protocol//specification.html#Requests_Initialize

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -3167,6 +3168,14 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                const length = isNaN(notifyData.len)
```
asimgunes: If `len` is malformed, we are supressing it. How about passing a warning message or a log message about it?

asimgunes (commented):
Hi @omarArm , this is a good update, I just pointed out some potential problems.

src/gdb/GDBDebugSessionBase.ts
omarArm: From my understanding, `supportMemoryEvent ` is something that the client has to set, not the adapter. That is why I didn't set it.

If vscode doesn't send it by default, I guess the way to go is to edit the `initializeRequest` in cdt-gdb-vscode and set it to true. I will test it and let you know

P.s. it's not available in the `InitializeResponse` interface in vscode.

src/gdb/GDBDebugSessionBase.ts
omarArm: good idea, I will add a warning message to it

src/gdb/GDBDebugSessionBase.ts
omarArm: Update: vscode does indeed send `supportMemoryEvent `

jreineckearm (approved):
Tested changes and events are sent as expected. There seems however to be an issue with Memory Inspector reading updated memory correctly but not displaying the changes. To be looked at separately and outside the scope of this PR.

My only point is coming back to Asim's comment about the initialize request: it's true that the GDB adapter doesn't need to set anything in its response. But the point of the client sending capabilities is to announce what it's able to deal with. Hence, we strictly speaking should only send the `memory` event if the client announces support (to digest it).

In reality, I'd expect a client to just ignore an event if not supported. Hence, I am good to merge as is to move forward.

But for completeness, we should add a small follow-up PR that makes sending `memory` events conditional based on received capabilities. Could you please take care of that, @omarArm ?

cwalther:
> But for completeness, we should add a small follow-up PR that makes sending `memory` events conditional based on received capabilities. Could you please take care of that, @omarArm ?

Done in 2ccafafa668058f9f3952088b8d8808f092cc058 as part of #544.

PR #535, opened 2026-05-08
Add support set expression request

omarArm:
Supporting DAP setExpression request available [here](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_SetExpression)

Users can then set values of expressions available in the `watch` window.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2178,8 +2178,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     }
 
     // In implementing this request. We first try and check if the variable exists as is
-    protected async setExpressionRequest(response: DebugProtocol.SetExpressionResponse,
-                                         args: DebugProtocol.SetExpressionArguments): Promise<void> {
+    protected async setExpressionRequest(
+        response: DebugProtocol.SetExpressionResponse,
+        args: DebugProtocol.SetExpressionArguments
+    ): Promise<void> {
         if (!this.canRequestProceed()) {
             this.logger.verbose(
                 'Debug adapter cannot process set variable request, skipping it.'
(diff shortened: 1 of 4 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2179,13 +2174,54 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
-    // }
+    protected async setExpressionRequest(
+        response: DebugProtocol.SetExpressionResponse,
+        args: DebugProtocol.SetExpressionArguments
+    ): Promise<void> {
+        if (!this.canRequestProceed()) {
+            this.logger.verbose(
+                'Debug adapter cannot process set variable request, skipping it.'
+            );
+            console.log(args);
```
jreineckearm: I don't think we need this console log.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2179,13 +2174,54 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                : undefined;
+            const [gdb, frameRef, depth] =
+                await this.getFrameContext(initialFrameRef);
+            const varObj = gdb.varManager.getVar(
+                frameRef,
+                depth,
+                args.expression
+            );
+            if (!varObj) {
+                throw new Error(`Variable ${args.expression} not found`);
```
jreineckearm, line 2204: Strictly speaking, such request could come without a var object having been created by a previous evaluate request. But OK to accept this as a limitation to keep the implementation simpler. This request is currently intended to be used in combination with views that show an expression value.

@omarArm , can you please add a comment here that this is a deliberate choice to not handle that scenario for now?

src/integration-tests/evaluate.spec.ts
```diff
@@ -462,120 +362,44 @@ describe('evaluate request global variables', function () {
...
-        //   2) directly via an extra mi.sendVarDelete()
-        //
-        // The second delete triggers the GDB/MI error "Variable object not found",
-        // causing watch reevaluation to fail instead of recreating the varobj for
-        // the new in-scope local variable in the second block.
-
-        const first = await dc.evaluateRequest({
+    it('should be able to set a global variable and have that change reflected in the debuggee', async function () {
+        // Read global variable using watch
+        const watchResParent = await dc.evaluateRequest({
```
jreineckearm, line 394: Should we have another test for a simple type without children?

jreineckearm (commented):
Code looking good, some minor comments. Giving it a try in the meanwhile

jreineckearm (commented):
Tested and did what it was supposed to do. Happy to approve once feedback has been addressed and no other reviewer feedback comes in by then.

src/gdb/GDBDebugSessionBase.ts
omarArm: yep, that was a residue from debugging while developing. Done

src/gdb/GDBDebugSessionBase.ts
omarArm: done

src/integration-tests/evaluate.spec.ts
omarArm: the thing was that the C file used for testing didn't have a primitive global variable, that is why I skipped it. I added a primitive global variable now and added a test case accordingly

jreineckearm (approved):
Thanks for the changes! Good to go as far as I am concerned.
Waiting until Tuesday evening with merge to give other reviewers chance to comment.

asimgunes (approved):
LGTM, I don't have any comments at this point. Thanks.

PR #536, opened 2026-05-12
Serialize concurrent breakpoint handlers with a mutex

### LINKED ISSUE #218: Race condition when using breakpoint modes to set HW/SW breaks
Follow-up to PR #211 :

Above PR added context menu entries to explicitly set HW/SW breakpoints opposed to setting a "normal" breakpoint and changing the mode afterwards.

The implementation had to apply some workarounds due to limitations in the VS Code API and its DAP protocol support.
This introduced a race condition between setting/removing breakpoints when the added commands apply the required sequence of setting/modifying a breakpoint programmatically. This caused the following two side effects, of which the removal of all breakpoints is probably more problematic than just seeing unexpected log messages.

Reproducers:
1. Set a breakpoint through the column/margin, 2.) use one of the two new commands => I always see the log output No breakpoint number <n>. Is there anything we can do to suppress this? It's not super-blocking and only happens when explicitly using one of the new commands. But made me look twice to ensure this isn't a result of something going wrong. I expect similar confusion at users' end.
1. Set a breakpoint through the column/margin, 2.) use "set software breakpoint", 3.) call a > info b from the debug console => all breakpoints are gone in GDB (still visible in GUI). I am afraid this needs another look. This doesn't happen when you use "set hardware breakpoint" first.

asimgunes:
Fixes eclipse-cdt-cloud/cdt-gdb-vscode#218

## Problem

When multiple breakpoint requests arrive concurrently (e.g. `setBreakpointsRequest` and `setInstructionBreakpointsRequest`), each handler performs a read-modify-write of GDB's global breakpoint list. Without synchronization, two handlers can read stale state and overwrite each other's changes, silently losing breakpoints.

This race is triggered in practice by the HW/SW breakpoint mode commands introduced in eclipse-cdt-cloud/cdt-gdb-vscode#211. Those commands must programmatically remove and re-add a breakpoint, issuing two DAP requests in quick succession. The second handler can act on a stale snapshot before the first has finished, causing all breakpoints to disappear from GDB while remaining visible in the UI.

`pauseIfNeeded`/`continueIfNeeded` coalesces target pauses but does not prevent concurrent handlers from racing on the breakpoint list itself.

## Solution

Add a single `breakpointMutex` (`async-mutex`) to `GDBDebugSessionBase`, shared across all four breakpoint handler types. Each handler is split into a thin public wrapper that acquires the lock and a protected `do*` method that holds the existing implementation. Only the `do*` body runs under the mutex, keeping the critical section narrow.

Handlers covered:

| Wrapper | Implementation |
|---|---|
| `setBreakPointsRequest` | `doSetBreakPointsRequest` |
| `setFunctionBreakPointsRequest` | `doSetFunctionBreakPointsRequest` |
| `setDataBreakpointsRequest` | `doSetDataBreakpointsRequest` |
| `setInstructionBreakpointsRequest` | `doSetInstructionBreakpointsRequest` |

## Testing

Added `src/integration-tests/breakpointMutex.spec.ts`, a unit-level spec that stubs the `do*` methods via a sinon sandbox to simulate async work. It asserts that `maxConcurrent` never exceeds 1, confirming the mutex serializes all concurrent callers without requiring a live GDB process.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
   "dependencies": {
     "@vscode/debugadapter": "^1.68.0",
     "@vscode/debugprotocol": "^1.68.0",
+    "async-mutex": "^0.5.0",
     "node-addon-api": "^8.4.0",
     "serialport": "^13.0.0"
   },
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -25,6 +25,7 @@ import {
     TerminatedEvent,
 } from '@vscode/debugadapter';
 import { DebugProtocol } from '@vscode/debugprotocol';
+import { Mutex } from 'async-mutex';
 import { ContinuedEvent } from '../events/continuedEvent';
 import { StoppedEvent } from '../events/stoppedEvent';
 import { VarObjType } from '../varManager';
@@ -188,6 +189,12 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     // reference count of operations requiring pausing, to make sure only the
     // first of them pauses, and the last to complete resumes
     protected pauseCount = 0;
+
+    // Serializes the bodies of the breakpoint request handlers so their
+    // read-modify-write of GDB's global breakpoint list cannot interleave.
+    // pauseIfNeeded/continueIfNeeded only coalesces the target pause, it does
+    // not prevent concurrent handlers from acting on stale snapshots.
+    protected breakpointMutex = new Mutex();
 
     // Pending requests to pause a thread silently (without sending stopped event).
     // At the moment only used with pauseIfRunning().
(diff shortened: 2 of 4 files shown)


omarArm (approved):
LGTM. I believe these changes would highly improve usability

jreineckearm (commented):
Good to go based on Omar's feedback. Just would like to check if the two initially reported problems went away with that.

jreineckearm (approved):
Tested with the original reproducer from https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/211#pullrequestreview-3976534662
The two issues supposed to be fixed with this PR are indeed fixed. The commands could still be deactivated from the command palette to avoid provoking the expected error in that case. But lower priority IMO.

Thanks for fixing this. Feel free to merge.

asimgunes:
Thanks all, I am merging it now.

asimgunes:
> The two issues supposed to be fixed with this PR are indeed fixed. The commands could still be deactivated from the command palette to avoid provoking the expected error in that case. But lower priority IMO.

@jreineckearm , Since we need to update the adapter reference in the VSCode extension, during that update I can check your feedback about the UI optimisation too. Thanks for your suggestion.

PR #537, opened 2026-05-13
Add breakpoint locations request

omarArm:
Solves https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/538

I added support for `breakpointLocationsRequest`. Documentation available [here](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_BreakpointLocations).

I utilised the gdb/mi command -symbol-list-lines available [here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Symbol-Query.html#GDB_002fMI-Symbol-Query)

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -1581,7 +1581,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         return { resolved, deletes };
     }
 
-    protected async breakpointLocationsRequest(response: DebugProtocol.BreakpointLocationsResponse, args: DebugProtocol.BreakpointLocationsArguments): Promise<void> {
+    protected async breakpointLocationsRequest(
+        response: DebugProtocol.BreakpointLocationsResponse,
+        args: DebugProtocol.BreakpointLocationsArguments
+    ): Promise<void> {
         // Check if debug adapter is in a state to proceed with the request.
         if (!this.canRequestProceed()) {
             this.logger.verbose(
@@ -1597,13 +1600,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             const file = args.source.path;
             if (!file) {
                 throw new Error('Source path is missing');
-            }  
-            const availableLines = await mi.sendSymbolListLines(this.gdb, file); 
+            }
+            const availableLines = await mi.sendSymbolListLines(this.gdb, file);
             // Compare available lines with requested lines range and return the ones that are available in GDB
             const startLine = args.line;
             const endLine = args.endLine || args.line;
             const breakpoints = availableLines.lines
-                .filter((line) => parseInt(line.line, 10) >= startLine && parseInt(line.line, 10) <= endLine)
+                .filter(
+                    (line) =>
+                        parseInt(line.line, 10) >= startLine &&
+                        parseInt(line.line, 10) <= endLine
+                )
                 .map((line) => ({
                     line: parseInt(line.line, 10),
                 }));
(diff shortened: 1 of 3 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -1580,6 +1581,63 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                breakpoints,
+            };
+            this.sendResponse(response);
+        } catch (err) {
+            if (
+                err instanceof Error &&
+                err.message.includes(
+                    '-symbol-list-lines: Unknown source file name.'
+                )
+            ) {
```
asimgunes, line 1625: When this if condition is not met, the function is not going to send a response, it would be better if we have an else block here and invoke `sendResponse` or `sendErrorResponse` method.

src/mi/symbols.ts
```diff
@@ -87,3 +94,13 @@ export function sendSymbolInfoFunctions(
     }
     return gdb.sendCommand(command);
 }
+
+export async function sendSymbolListLines(
+    gdb: IGDBBackend,
+    file: string
+): Promise<MISymbolListLinesResponse> {
+    const response: MISymbolListLinesResponse = await gdb.sendCommand(
+        `-symbol-list-lines ${file}`
```
asimgunes: some file names can contain space character, its better to use the `standardEscape` function here.

location: `src\util\standardEscape.ts`

asimgunes (commented):
LGTM in general, I added some potential issues above.

src/mi/symbols.ts
omarArm: Good catch, I didn't know we had such a function available

src/gdb/GDBDebugSessionBase.ts
omarArm: The idea that I had in mind is to fail silently, but I guess you are right. If a different error pops up for any reason, it should be printed out to the user.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -1580,6 +1581,69 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+        try {
+            const file = args.source.path;
+            if (!file) {
+                throw new Error('Source path is missing');
+            }
+            const availableLines = await mi.sendSymbolListLines(this.gdb, file);
+            // Compare available lines with requested lines range and return the ones that are available in GDB
+            const startLine = args.line;
+            const endLine = args.endLine || args.line;
+            const breakpoints = availableLines.lines
```
jreineckearm, line 1608: Nit: probably personal style but could reduce the number of calling `parseInt(line.line, 10)` by swapping the `map` and the `filter` calls. Also would make the filter condition less bulky.

jreineckearm (approved):
Changes look good to me, just one nit to review.

Functionality should be side effect free, advancing breakpoints with new request support should work as before.

One initial mild concern about the length of `-symbol-list-lines` responses and that this gets repeatedly called for each attempt to set a code breakpoint. But all on GDB level, hence not expecting much overhead in remote protocol comms.

Setting a breakpoint in VS Code "breakpoints everywhere" mode also is handled gracefully when the associated `-symbol-list-lines` call fails for a file.

Will approve and leave unmerged until early next week to give other reviewers chance to comment. But then happy to merge.

Let me know if you'd like to address the "nit". Then I'll re-approve.

src/gdb/GDBDebugSessionBase.ts
omarArm: nice suggestion. Done

jreineckearm (approved):
No more comments from my side, good to go.
Waiting with merge until Tuesday afternoon to give other reviewers a chance to comment.

asimgunes (approved):
LGTM, thanks.

Issue #538, opened 2026-05-13
Add support for breakpointLocations request


omarArm:
I think adding support for the request `breakpointLocations` would be quite useful in queuing which source code lines are associated with actual machine code.

omarArm:
Done in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/537

Issue #527, opened 2026-04-29
Support setting expressions' values


omarArm:
# Background: 
The current adapter state supports `setVariableRequest`, which is quite good for setting values to local variables and registers. However, a similar mechanism should be available for global variables.

# Action item:
After investigations, I think the way to go would be to support `setExpressionRequest` in the adapter. That would enable users to write global variables as an expression in the watch window and then set it's value accordingly.

omarArm:
done in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/535

PR #542, opened 2026-05-27
1.9.0 release preparations

### LINKED ISSUE #218: Race condition when using breakpoint modes to set HW/SW breaks

### LINKED ISSUE #219: Handling discrepancies between debug console and context menus when it comes to global radix change
As per the discussion in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/217 . It has been agreed that the extension needs a custom event to handle users changing radices from the debug console without breaking the current system provided by the GUI for changing radices.

This should be solved in 2 steps

1. Creating a custom event in the cdt-gdb-adapter to be sent, along with an invalidated event, whenever the session's radix changes.
2. This custom event should be handled by the extension and updates the context accordingly.

omarArm:
- Implements [`#526`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/526): Handle memory-changed notifications
- Implements [`#527`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/527): Support setting expressions' values
- Implements [`#538`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/538): Add support for breakpointLocations request
- Fixes [`#219`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/219): Handling discrepancies between debug console and context menus when it comes to global radix change
- Fixes [`#218`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/218): Race condition when using breakpoint modes to set HW/SW breaks

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Change Log
 
+## 1.9.0
+
+- Implements [`#526`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/526): Handle memory-changed notifications
+- Implements [`#527`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/527): Support setting expressions' values
+- Implements [`#538`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/538): Add support for breakpointLocations request
+- Fixes [`#219`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/219): Handling discrepancies between debug console and context menus when it comes to global radix change
+- Fixes [`#218`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/218): Race condition when using breakpoint modes to set HW/SW breaks
+
 ## 1.8.1
 
 - Fixes [`#530`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/530): Bump publish job to Node 24 to work around an npm self-upgrade crash that prevented 1.8.0 from being published to npm.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,9 @@
 ## 1.9.0
 
 - Implements [`#526`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/526): Handle memory-changed notifications
-- Implements [`#527`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/527): Support setting expressions' values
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Change Log
 
+## 1.9.0
+
+- Implements [`#526`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/526): Handle memory-changed notifications
+- Implements [`#527`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/527): Support setting expressions' values
```
jreineckearm: Nit: single tick after `expressions`

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+- Implements [`#538`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/538): Add support for breakpointLocations request
+- Fixes [`#219`](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/219): Handling discrepancies between debug console and context menus when it comes to global radix change
```
jreineckearm: Maybe simpler language: `Synchronize global radix change between debug console and context menus`

CHANGELOG.md
omarArm: done

CHANGELOG.md
omarArm: done

jreineckearm (approved):
LGTM, thanks!

Issue #539, opened 2026-05-20
Add `memoryReference` to evaluate response where possible


jreineckearm:
Found as part of a PR on Memory Inspector to add support for "Show in Memory Inspector" from watch window: https://github.com/eclipse-cdt-cloud/vscode-memory-inspector/pull/166#issuecomment-4501661169

For VS Code to set the `canReadMemory` context value to `true` for a watch window entry, the `evaluate` response needs to return a valid `memoryReference` property. See for example https://github.com/microsoft/vscode/blob/9ecd1e915b655f7c5d09fe4fa7fe908b930bc1e2/src/vs/workbench/contrib/debug/browser/watchExpressionsView.ts#L402

We should add this for better integration with that addition.

omarArm:
done in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/544

PR #544, opened 2026-05-28
sending memory reference as part of evaluate request response

omarArm:
Should implement https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/539

Changes:
1. Whenever a response for the `evaluateRequest` is finally getting prepared, the adapter would send an MI command retrieving the address of the expression (if available)
2. If address is available, use it as a memory reference.

The perks of implementing such a feature is that vscode will then set the `canReadMemory ` context. As mentioned in the original issue linked above

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -52,6 +52,7 @@ import { DEFAULT_STEPPING_RESPONSE_TIMEOUT } from '../constants/session';
 import { ThreadWithStatus } from './common';
 import { RESUME_COMMANDS, SET_ALL_CHARSET_REGEXPS } from '../constants/gdb';
 import { GDBThreadRunning } from './errors';
+import { MIGDBDataEvaluateExpressionResponse } from '../mi';
 
 /**
  * Keeps track of where in the configuration phase (between initialized event
@@ -157,6 +158,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected isRunning = false;
 
     protected supportsRunInTerminalRequest = false;
+    protected supportsMemoryReferences = false;
     public supportsGdbConsole = false;
 
     /* A reference to the logger to be used by subclasses */
@@ -419,6 +421,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     ): void {
         this.supportsRunInTerminalRequest =
             args.supportsRunInTerminalRequest === true;
+        this.supportsMemoryReferences = args.supportsMemoryReferences === true;
         this.supportsGdbConsole =
             os.platform() === 'linux' && this.supportsRunInTerminalRequest;
         response.body = response.body || {};
@@ -2674,10 +2677,35 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                               varobjName: varobj.varname,
                           })
                         : 0;
+                let memoryReferenceResult:
+                    | MIGDBDataEvaluateExpressionResponse
+                    | undefined;
+                let hasMemoryReference = false;
+                if (this.supportsMemoryReferences) {
+                    try {
+                        memoryReferenceResult =
+                            await mi.sendDataEvaluateExpression(
+                                gdb,
+                                `&(${varobj.expression})`
+                            );
+                        // Depending on the GDBServer being used, sometimes the result of evaluating an address of a symbol returns "<address> <symbol name>"
+                        if (memoryReferenceResult.value?.includes(' ')) {
+                            memoryReferenceResult.value =
+                                memoryReferenceResult.value.split(' ')[0];
+                        }
+                        hasMemoryReference = true;
+                    } catch {
+                        hasMemoryReference = false;
+                    }
+                }
                 response.body = {
                     result,
                     type: varobj.type,
                     variablesReference,
+                    memoryReference:
+                        hasMemoryReference && memoryReferenceResult
+                            ? memoryReferenceResult.value
+                            : undefined,
                 };
             }
 
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -159,6 +159,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
 
     protected supportsRunInTerminalRequest = false;
     protected supportsMemoryReferences = false;
+    protected supportsMemoryEvent = false;
     public supportsGdbConsole = false;
 
     /* A reference to the logger to be used by subclasses */
@@ -422,6 +423,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.supportsRunInTerminalRequest =
             args.supportsRunInTerminalRequest === true;
         this.supportsMemoryReferences = args.supportsMemoryReferences === true;
+        this.supportsMemoryEvent = args.supportsMemoryEvent === true;
         this.supportsGdbConsole =
             os.platform() === 'linux' && this.supportsRunInTerminalRequest;
         response.body = response.body || {};
@@ -3339,6 +3341,9 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.handleCmdParamChanged(notifyData);
                 break;
             case 'memory-changed': {
+                if (!this.supportsMemoryEvent) {
+                    break;
+                }
                 let length = 0;
                 if (isNaN(notifyData.len)) {
                     logger.warn(
--- a/src/integration-tests/debugClient.ts
+++ b/src/integration-tests/debugClient.ts
@@ -141,6 +141,7 @@ export class CdtDebugClient extends DebugClient {
         if (!args) {
             args = {
                 supportsRunInTerminalRequest: true,
+                supportsMemoryReferences: true,
                 adapterID: this['_debugType'],
                 linesStartAt1: true,
                 columnsStartAt1: true,
--- a/src/integration-tests/debugClient.ts
+++ b/src/integration-tests/debugClient.ts
@@ -141,6 +141,7 @@ export class CdtDebugClient extends DebugClient {
         if (!args) {
             args = {
                 supportsRunInTerminalRequest: true,
+                supportsMemoryEvent: true,
                 supportsMemoryReferences: true,
                 adapterID: this['_debugType'],
                 linesStartAt1: true,
--- a/src/integration-tests/evaluate.spec.ts
+++ b/src/integration-tests/evaluate.spec.ts
@@ -531,6 +531,26 @@ describe('evaluate request global variables', function () {
         });
         expect(evalRes.body.result).to.equal('25');
     });
+
+    it('response for variables should include memory reference if the expression has an lvalue', async function () {
+        // Read a global variable using evaluateRequest
+        const watchRes = await dc.evaluateRequest({
+            context: 'watch',
+            expression: 'global_int',
+            frameId: scope.frame.id,
+        });
+        expect(watchRes.body.memoryReference).to.not.be.undefined;
+    });
+
+    it('should not include memory reference if the expression does not have an lvalue', async function () {
+        const watchRes = await dc.evaluateRequest({
+            context: 'watch',
+            expression: 'global_int + 1',
+            frameId: scope.frame.id,
+        });
+
+        expect(watchRes.body.memoryReference).to.be.undefined;
+    });
 });
 
 describe('evaluate request - watch local variable across lexical scope transition', function () {
--- a/src/integration-tests/breakpoints.spec.ts
+++ b/src/integration-tests/breakpoints.spec.ts
@@ -444,6 +444,7 @@ describe('breakpoints', async function () {
             // win32 host can only pause remote + mi-async targets
             this.skip();
         }
+        const firstStoppedEvent = dc.waitForEvent('stopped');
         let response = await dc.setBreakpointsRequest({
             source: {
                 name: 'count.c',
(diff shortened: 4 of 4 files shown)


src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2666,6 +2667,23 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
                         ).value
                       : varobj.value;
+                let memoryReferenceResult:
+                    | MIGDBDataEvaluateExpressionResponse
+                    | undefined;
+                let hasMemoryReference = false;
+                try {
+                    memoryReferenceResult = await mi.sendDataEvaluateExpression(
+                        gdb,
+                        `&${varobj.expression}`
```
cwalther: This should probably be `&(${varobj.expression})` because the `&` operator has relatively high precedence, otherwise on an expression like `x + 1` we would get nonsense. (Maybe add such a thing to the test cases?)

Also, the [DAP specification](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Evaluate) says
> For pointer type eval results, this is generally a reference to the memory address contained in the pointer.

so in that case we shouldn’t apply the `&` operator but use the value directly. Can we somehow determine whether the expression is pointer-typed?

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2666,6 +2667,23 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                    );
+                    if (memoryReferenceResult.value?.includes(' ')) {
```
cwalther: In what circumstances does this occur? A comment with an example would be helpful for understanding the code without having to try it.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2678,6 +2696,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     result,
                     type: varobj.type,
                     variablesReference,
+                    memoryReference:
```
cwalther, line 2707: DAP spec says this should only be set when the client advertises capability `supportsMemoryReferences`, so we could skip determining it at all otherwise.

cwalther (commented):
Looks like a useful addition! I haven’t tried it, but a few comments from reading the code.

src/gdb/GDBDebugSessionBase.ts
omarArm: I added this check because some evaluate requests would be for expressions that don't have memory addresses associated to it. An example would be a simple expression or a primitive local variable that gets stored in a core register instead of being written to the memory.

This in turn triggers an error in GDB and makes the whole process fail, even if the client does support memory references. Therefore, I set `memoryReference` to `undefined` whenever this happens for only some expressions

src/gdb/GDBDebugSessionBase.ts
omarArm: Done, I added the comment, thanks for pointing that out

src/gdb/GDBDebugSessionBase.ts
cwalther: I am not asking why you added the check. I am suggesting that there should be an additional check: put the whole `try` block from line 2674 into an `if (client supports memory references) {` so that `memoryReference` is also left out, and we don’t do any unnecessary work, when that is not the case.

src/gdb/GDBDebugSessionBase.ts
omarArm: I added the extra parentheses and added an extra test for expressions that don't have an lvalue as well. Thanks for pointing that out

src/gdb/GDBDebugSessionBase.ts
omarArm: Ah ok, apologies for the confusion.

I thought because it's an inline comment that you meant adding the client's check would negate the idea of having the current existing checks

omarArm: I added the check for both `memoryReferences` and `memoryEvents`. Thanks for pointing that out

src/integration-tests/evaluate.spec.ts
```diff
@@ -531,6 +531,26 @@ describe('evaluate request global variables', function () {
...
+        // Read a global variable using evaluateRequest
+        const watchRes = await dc.evaluateRequest({
+            context: 'watch',
+            expression: 'global_int',
+            frameId: scope.frame.id,
+        });
+        expect(watchRes.body.memoryReference).to.not.be.undefined;
+    });
+
+    it('response for variables should include memory reference if the expression does not have an lvalue', async function () {
```
cwalther:
```suggestion
    it('response for variables should not include memory reference if the expression does not have an lvalue', async function () {
```

src/gdb/GDBDebugSessionBase.ts
cwalther: There is a bit of duplication now, I would just have moved the closing `}` of the `if` a bit further up (and the `let hasMemoryReference = false;` initialization out of the block), or even moved the whole `if` after `response.body = { … }` (so you wouldn’t even need the `hasMemoryReference` variable but could just fill in the `memoryReference` exactly when available), but I leave that up to you.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2666,19 +2671,58 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                                gdb,
+                                `&(${varobj.expression})`
+                            );
+                        // Depending on the GDBServer being used, sometimes the result of evaluating an address of a symbol returns "<address> <symbol name>"
+                        if (memoryReferenceResult.value?.includes(' ')) {
+                            memoryReferenceResult.value =
+                                memoryReferenceResult.value.split(' ')[0];
+                        }
+                        hasMemoryReference = true;
+                    } catch (err) {
```
cwalther: Did you see the lint warning on this line? (I see it inline in the GitHub diff view.)

src/gdb/GDBDebugSessionBase.ts
omarArm: done

src/gdb/GDBDebugSessionBase.ts
omarArm: I don't see any warnings, can you please show me a screenshot? maybe it's a committer only thing?

cwalther: I don’t see it right now either, probably because CI is still running. Check again when it’s done.

cwalther: image

Sorry about the mess with the canceled jobs, I was trying to re-run the one with the apparently flaky test at the same time as you were pushing a new commit. I will re-run it again now as it failed again. (That will probably make the warning disappear again until it’s done.)

omarArm: oh, I wasn't sure why are jobs failing and I tried to do some dummy commits to rerun CI (I don't have access to run CI without code changes). Thanks for letting me know

omarArm:
@cwalther I squashed a few commits and CI is now working, could you please do a final review and merge if it's acceptable to you?

cwalther:
This point from one of my first comments is still open:
> Also, the [DAP specification](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Evaluate) says
> 
> > For pointer type eval results, this is generally a reference to the memory address contained in the pointer.
> 
> so in that case we shouldn’t apply the `&` operator but use the value directly. Can we somehow determine whether the expression is pointer-typed?

I leave it to you (or @jreineckearm if he is around) whether to fix this here or keep it as a known deficiency to be fixed later. If “later” is not in the near future, an issue should be opened for it. Let me know what you prefer (or if you disagree entirely).

Apart from that, the code looks good to me, and I have now tried it and it appears to work as intended. 👍

The commits are still a bit of a mess: e.g. the [last one](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/544/commits/a4fccabe56241a0525db3128a9be19fb3b67eb13) includes a change in GDBDebugSessionBase.ts that has nothing to do with what the commit message says.

There are two changes here that I would prefer to keep in separate commits because they are unrelated to the topic of the PR:
- the handling of `supportsMemoryEvent` (part of https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/544/commits/26f523d822ef88f53bb6b8b59582896922b6ffa8) (that is the solution to https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/533#pullrequestreview-4328155339)
- the fixing of the flaky test in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/544/commits/a4fccabe56241a0525db3128a9be19fb3b67eb13.

The rest of the changes I would prefer to see squashed into one commit, because some of the later commits undo or amend changes done in earlier ones, and we don’t need to have this back-and-forth recorded. Let me know if you disagree.

Could you bring things into such a shape of three commits using an interactive rebase (ask if you need help)?

omarArm:
Oh, sorry I must have missed it. I am not sure what would be the best way to implement it yet. Therefore, I would rather have it in a separate PR. That being said, I created an issue for it. IMO, we should create issues for any future changes even if they are planned to be done in the near future.

I cleaned up the commits as requested as well

cwalther:
Thanks for opening the issue!

The commits are not quite cleaned up yet, I’m afraid. Commits 2 and 3 still contain changes that belong in commit 1, e.g. the `catch (err)` → `catch` part. ([Documentation of how to split commits](https://git-scm.com/docs/git-rebase#_splitting_commits))

cwalther:
It still doesn’t seem quite right (removals in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/544/commits/424db4eb0040350428bb3e0e10a4aa6b5a6fdfbf?) and has somehow become four commits now, but I don’t have time for a closer look right now, will continue tomorrow.

omarArm:
@cwalther thanks for the feedback, could you please check now?

src/integration-tests/evaluate.spec.ts
```diff
@@ -539,10 +539,10 @@ describe('evaluate request global variables', function () {
             expression: 'global_int',
             frameId: scope.frame.id,
         });
-        expect(watchRes.body.memoryReference).to.be.a('string');
+        expect(watchRes.body.memoryReference).to.not.be.undefined;
     });
 
-    it('response for variables should include memory reference if the expression does not have an lvalue', async function () {
+    it('should not include memory reference if the expression does not have an lvalue', async function () {
```
cwalther: These have nothing to do with memory events but are amendments of the previous commit, and therefore belong squashed into the previous commit.

src/integration-tests/debugClient.ts
```diff
@@ -141,6 +141,8 @@ export class CdtDebugClient extends DebugClient {
         if (!args) {
             args = {
                 supportsRunInTerminalRequest: true,
+                supportsMemoryEvent: true,
+                supportsMemoryReferences: true,
```
cwalther, line 145: Strictly speaking the second line also doesn’t belong here, but depending on whether it is actually required for the tests to pass in the previous commit (I didn’t check), I could accept leaving it here.

cwalther (commented):
Sorry, still not good. See inline comments on commit “checking if client supports memory events”.

#### Here is what I would do to clean it up:

Name a reference to the previous state for more convenient access
```sh
git branch addMemoryReferenceToEvaluateRequest-backup
```

Split the commit that contains too many changes into a temporary one with the extraneous changes and the real one with the intended changes (following [instructions](https://git-scm.com/docs/git-rebase#_splitting_commits))
```sh
git rebase -i origin/main
# change second commit from "pick" to "e"
git reset HEAD^
git add src/integration-tests/evaluate.spec.ts
git add -p src/integration-tests/debugClient.ts
# e
# remove the supportsMemoryEvent line but keep the supportsMemoryReferences one
git commit -m "temp"
git add -u
git commit -C addMemoryReferenceToEvaluateRequest^
git rebase --continue
```

Squash the temporary commit into its parent
```sh
git rebase -i origin/main
# change temp "temp" commit from "pick" to "f"
```

Check whether the result is as expected
```sh
git diff addMemoryReferenceToEvaluateRequest-backup addMemoryReferenceToEvaluateRequest
# diff should be empty
git range-diff addMemoryReferenceToEvaluateRequest-backup...addMemoryReferenceToEvaluateRequest
# should show desired modifications to the first two commits, no change to commit messages
git branch -D addMemoryReferenceToEvaluateRequest-backup
```

Use `git status` and `git diff --staged` anywhere in between to see what you are doing.

Does that help?

src/integration-tests/debugClient.ts
omarArm: done

omarArm:
@cwalther please have a look now. This was actually a really fun git exercise for me. Thank you for nudging me towards it

cwalther:
Thank you, now it’s looking good!

Glad the nudging helped. I can’t spend that effort every time, but if it helps people get more proficient and need less nudging in the future then it’s well spent.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2674,10 +2679,35 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                               varobjName: varobj.varname,
                           })
                         : 0;
+                let memoryReferenceResult:
+                    | MIGDBDataEvaluateExpressionResponse
+                    | undefined;
+                let hasMemoryReference = false;
```
jreineckearm, line 2685: Nit: probably one of the two variables would have been enough (`undefined` for `memoryReferenceResult` meaning no memory reference). Maybe next time we touch this area.

jreineckearm (commented):
Thanks for the review, @cwalther ! And apologies for chiming in only now. Was on holiday again, but should be back now for a bit.

> I leave it to you (or @jreineckearm if he is around) whether to fix this here or keep it as a known deficiency to be fixed later. If “later” is not in the near future, an issue should be opened for it. Let me know what you prefer (or if you disagree entirely).

@omarArm , let's have a look at what can be done. I believe there was some heuristics to determine a pointer in other parts of the code. But I also remember that those parts needed review for corner cases.

I see you have already raised https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/547 . Thanks!

PR #546, opened 2026-06-02
Not showing error response as a notification pop-up when evaluate request context is repl

### LINKED ISSUE #545: Error messages from invalid debug console input should only appear in the debug console
# Current behaviour:

Whenever the adapter catches an error in the `evaluate request`, it sends an `error response` without setting the destination. Which prompts vscode to show the error to users as a pop up AND also as a message in the debug console. 

Image

# Suggested fix

The current behaviour is not a bad idea. However, if the `evaluate request` has the context `repl`, errors should appear only in the debug console and not as a pop-up, since that would mean users are using the debug console.

omarArm:
This should fix #545 

image

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2692,8 +2692,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.sendErrorResponse(response, {
                     id: 1, // Id = 1 has always been the case for evaluateRequests
                     format: err instanceof Error ? err.message : String(err),
-                    showUser: false,
-                    sendTelemetry: undefined,
+                    showUser: false
                 });
             } else {
                 this.sendErrorResponse(
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2694,7 +2694,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     1,
                     err instanceof Error ? err.message : String(err),
                     undefined,
-                    2 // Send error only to telemtry to appear only in debug console and not in UI popups
+                    2 // Send error only to telemetry to appear only in debug console and not in UI popups
                 );
             } else {
                 this.sendErrorResponse(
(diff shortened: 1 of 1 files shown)


cwalther:
I am against any sort of telemetry (euphemism for “surveillance“).

If `context === 'repl'` does guarantee that the user has the debug console open and the error message has already been posted there as text output from GDB (which I do not know offhand), then why not just skip reporting it altogether?

Or what happens when you just remove the `User` bit without adding the `Telemetry` bit (i.e. pass 0 as `dest`)?

cwalther:
Come to think of it, at least when `isCliCommand`, we might even treat that condition as success from the DAP point of view? We successfully sent a command to GDB and successfully got an error message back?

omarArm:
@cwalther I removed the telemetry option and went with using the `Message` interface from the debug protocol. In it, there is an optional property called `showUser`. Do you think that would be acceptable?

Actually thinking of your second comment

> Come to think of it, at least when isCliCommand, we might even treat that condition as success from the DAP point of view? We successfully sent a command to GDB and successfully got an error message back?

I guess you do have a point, from an MSDAP point of view, it is a success and we should send just a normal response. Maybe we should add @jreineckearm and @jonahgraham to the discussion to see their opinion on the matter. I will convert the PR to a draft for the time being

cwalther:
> I removed the telemetry option and went with using the `Message` interface from the debug protocol. In it, there is an optional property called `showUser`. Do you think that would be acceptable?

Yes, it’s equivalent to passing `dest = 0`, see [implementation](https://github.com/microsoft/vscode-debugadapter-node/blob/476803c69b6b78b12980b422448b9dda02f4a591/adapter/src/debugSession.ts#L459-L478).

`sendTelemetry: undefined` is redundant, it disappears in the JSON serialization.

jreineckearm:
I agree on not using telemetry.

I think we should only check for context `repl` and not for `isCliCommand`. Assumption is that VS Code and other extensions use that context only for interactive REPL inputs/consoles.

jreineckearm:
@omarArm , code formatting check failed.

cwalther (approved):
Tested, looks good to me.

jreineckearm:
I realize now that I confused things with my comment `isCliCommand`. Was referring to always send a not-error response if `repl`. But understand in the meanwhile also that a genuine error should be sent if something breaks internally with the varobj logic.

Let's merge as is. But aim to add another small PR to add the `isCliCommand` check in the error handler which decides if to send a success response or a proper error response (the latter still causing duplication of the log output line).

PR #548, opened 2026-06-08
Optionally keep stopped threads stopped on attaching

### LINKED ISSUE #528: How to keep stopped threads stopped on attaching?
Currently, when attaching to an inferior that already has some stopped threads, these threads are resumed by an indiscriminate _continue_ command in the final `continueIfNeeded` call at the end of the configuration phase. This is a problem for us: our systems can have breakpoints even when no debugger is attached, and when a thread stops on such a breakpoint, we want to be able to attach a debugger and analyze the thread in this stopped state, and not have it continued. It’s also inconsistent: continue*IfNeeded* is supposed to be a counterpart to pause*IfNeeded*, and pausing was not needed in this case because there was a paused thread already, so there shouldn’t be a matching _continue_ either.

This is easy enough to fix, and I had done it locally months ago, but now that I wanted to prepare it for submission in a PR, I found that it breaks half the tests. Drat. Apparently the thing I had regarded as a deficiency may in fact have been a deliberate feature.

What breaks the tests is that both [GDB with a native target](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Attach.html) and `gdbserver` (unlike our stub) stop all threads on attaching, but the tests expect the process to be running (and hit a breakpoint) after the configuration phase.

If the tests expect that behavior, then maybe some users expect it as well, so maybe we shouldn’t just change it? It even makes sense in a way from the point of a GDB `native` (local, type `gdb`) or `gdbserver` (remote, type `gdbtarget`) user, because the stop-on-attach behavior of those two destroys the original state of the inferior anyway, and there is no use preserving the destroyed state, rather what should be preserved is the state before it was destroyed, which was most likely “all running”.

I need it changed though, not just for the reason above, but also (and that’s why I hadn’t bothered submitting it until now), in a situation that I am currently working on, this change is not just nice to have, but things would not work at all without it. The situation is that the stub is not attached to an inferior yet at the time GDB connects to it, but I only issue a `-target-attach 1 &` command just before the end of the configuration phase, after receiving breakpoints. This is possible when using the `extended-remote` target in GDB. (I can go into details separately on _why_ I am doing this, if anyone is interested.) In this case, there are no threads yet at the beginning of the configuration phase, so pausing is not needed, but they arrive before the end of the configuration phase and may all be running at that time. An indiscriminate _continue_ in that case results in an error “Cannot execute this command while the selected thread is running.” from GDB that takes down the whole debug session.

Any opinions on how to handle this?

I can think of these ways:
- _(may violate user expectations)_ Change the behavior unconditionally.
- _(easy way out)_ Make it configurable using an option in the attach arguments.
- _(perhaps harebrained)_ Define a way for a stub to advertise to the adapter that it is capable of preserving the stopped/running state of inferior threads on attaching and doesn’t want an indiscriminate _continue_ command. Perhaps through a `monitor` command, or is there any other way for the adapter to talk directly to a stub?

CC @jreineckearm, @jonahgraham, @asimgunes

cwalther:
As discussed in #528:

Currently, when attaching to an inferior that already has some stopped threads, these threads are resumed by an indiscriminate _continue_ command in the final `continueIfNeeded` call at the end of the configuration phase. This is a problem for us: our systems can have breakpoints even when no debugger is attached, and when a thread stops on such a breakpoint, we want to be able to attach a debugger and analyze the thread in this stopped state, and not have it continued. It’s also inconsistent: continue*IfNeeded* is supposed to be a counterpart to pause*IfNeeded*, and pausing was not needed in this case because there was a paused thread already, so there shouldn’t be a matching _continue_ either.

### Commit 1

The first of the attached commits fixes that by inserting the appropriate `if (this.waitPausedNeeded)`. Because existing clients, and that in fact includes many of the tests, may expect the previous behavior, gate it behind an option in the launch/attach arguments: The new, optional property `"run"` can have values `"always"` or `"preserve"` (and potentially more in the future). `always` is the default and corresponds to the previous behavoir: always let all threads run. `preserve` is the new behavior: leave the running state of threads unchanged. Alternative suggestions for names of option and values are welcome.

This is not sufficient however to fully handle this situation: Because `async: stopped` notifications from GDB are not relayed to the client as DAP `stopped` events until `this.isInitialized`, which happens at the beginning of the configuration phase after the `initialized` event, and in this case the `stopped` notification from GDB comes earlier than that, the client would not know about the stopped thread and would display all threads as running. Even manually clicking _Pause_ on such a thread does not help, as that does not cause GDB to send another `stopped` notification because to it the thread is stopped already. Therefore, any `stopped` notifications from GDB that arrive during this time of suppression must not be simply discarded, but must be stored and delivered deferredly when the suppression ends. `running` notifications can still be discarded because running is the default state of a thread, but one that comes after a `stopped` notification for the same thread can cancel the latter. This is the rest of the first commit.

Moreover, in a situation that I am currently working on, this change is not just nice to have, but things would not work at all without it. The situation is that the stub is not attached to an inferior yet at the time GDB connects to it, but I only issue a `-target-attach 1 &` command just before the end of the configuration phase, after receiving breakpoints. This is possible when using the `extended-remote` target in GDB. (I can go into details separately on _why_ I am doing this, if anyone is interested.) In this case, there are no threads yet at the beginning of the configuration phase, so pausing is not needed, but they arrive before the end of the configuration phase and may all be running at that time. An indiscriminate _continue_ in that case results in an error “Cannot execute this command while the selected thread is running.” from GDB that takes down the whole debug session.

The first commit makes some tests in the _remote gdb-non-stop_ scenario fail with “AssertionError [ERR_ASSERTION]: '0' == 'breakpoint'”. That is not due to a problem with the code under test, but due to a problem with `DebugClient.hitBreakpoint()` from @vscode/debugadapter-testsupport. `hitBreakpoint` assumes that the very first `stopped` event comes from its breakpoint. That is not a reasonable assumption to make, and in fact is necessarily false in the situation that this commit seeks to enable, keeping stopped threads stopped. The only reason why it held so far is that we erroneously discarded early stop notifications before.

However, the problematic behavior of `hitBreakpoint` will be sidestepped as a side effect of the second commit, therefore I decided not to try to fix it here.

### Commit 2

The second commit is a further adaptation to this special situation of late attaching, but I think it does not hurt in the general case, so I am proposing it along with the first one: Unlike in the normal early-attach case (`remote` target), in the late-attach situation (`extended-remote` target), existing threads (a potentially large number of them) arrive from GDB via `async: running` notifications _after_ `this.isInitialized` is set, which therefore are no longer suppressed from going to the client as `continued` events. The client (at least VS Code) responds to each of them with a `threads` request, and if you have 110 threads as I do in my current example system, that results in a lot of completely unnecessary XML transmission over the remote connection and takes a long time. (I am guessing this is the reason why the suppression was introduced in the first place.) I therefore propose to defer continued/stopped events not just to the beginning, but to the _end_ of the configuration phase. This is what the second commit does. Does everyone agree that this does not cause any problems? The suppression could be extended even further without any harm, it does not cause us to miss anything except that the client only notices stopped/stopping threads once it is lifted, so there is no need to lift it until the user is ready for that.

This leaves the `isInitialized` variable unused, but since it is `protected` and may already be used by third-party subclasses, I am leaving it in.

As a side effect, because the longer suppression lets the `continued` event caused by the _continue_ command from `"run": "always"` cancel the deferred early `stopped` event that would confuse `hitBreakpoint`, this also avoids the test failures from the first commit. From here on, all tests should pass after every commit.

### Commit 3

A `launch` request with `"run": "preserve"` leaves GDB in the state where it has not started the program yet. To start the program, GDB needs a _run_ command. So far, there has been no way for the user to explicitly send that command through DAP, other than typing it in the debug console, it was only sent automatically in the `"run": "always"` (default) case. Therefore, in the `preserve` case, let the first DAP `continue` request send _run_ instead of _continue_ to GDB (the latter would not be accepted in this state anyway).

### Commits 4–6

Since all the existing tests run without a `run` option and therefore in the `always` mode, add some tests for explicit `run` options and the `preserve` mode in commit 6. To check whether threads are in the desired running/stopped state, these tests send the `-thread-info` command to GDB, which is much simpler than piecing together this information from DAP responses and events. Making this possible required some further modifications in commits 4 and 5:

- Accept GDB commands without a frame ID also in `gdb`, not just `gdbtarget` debug type – see also #414 and https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/202#pullrequestreview-3860526810 (middle).
- Return the result from an evaluated MI command.

Closes #528.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2502,7 +2502,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         response: DebugProtocol.EvaluateResponse,
         args: DebugProtocol.EvaluateArguments
     ): Promise<void> {
-        return this.doEvaluateRequest(response, args, false);
+        return this.doEvaluateRequest(response, args, true);
     }
 
     private extractExpressionFormat(
@@ -2546,7 +2546,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     protected async doEvaluateRequest(
         response: DebugProtocol.EvaluateResponse,
         args: DebugProtocol.EvaluateArguments,
-        alwaysAllowCliCommand: boolean // if true, allows evaluation of expression without a frameId
+        alwaysAllowCliCommand: boolean // if true, allows evaluation of expression without a frameId (kept for compatibility)
     ): Promise<void> {
         response.body = {
             result: 'Error: could not evaluate expression',
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -212,7 +212,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     // keeps track of where in the configuration phase (between initialize event
     // and configurationDone response) we are
     protected configuringState: ConfiguringState = ConfiguringState.INITIAL;
-    protected isInitialized = false;
+    protected isInitialized = false; // unused here but kept for compatibility
     protected deferredStopEvents: any[] = [];
 
     /**
@@ -609,10 +609,6 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         }
         this.sendEvent(new InitializedEvent());
         this.isInitialized = true;
-        for (const resultData of this.deferredStopEvents) {
-            this.handleGDBStopped(resultData);
-        }
-        this.deferredStopEvents.splice(0);
     }
 
     protected async attachRequest(
@@ -778,12 +774,19 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.waitPaused = undefined;
     }
 
+    protected setConfiguringStateDone(): void {
+        this.configuringState = ConfiguringState.DONE;
+        for (const resultData of this.deferredStopEvents) {
+            this.handleGDBStopped(resultData);
+        }
+        this.deferredStopEvents.splice(0);
+    }
+
     protected async continueIfNeeded(): Promise<void> {
         if (this.pauseCount > 0) {
             this.pauseCount--;
             if (this.pauseCount === 0) {
                 if (this.configuringState === ConfiguringState.FINISHING) {
-                    this.configuringState = ConfiguringState.DONE;
                     if (
                         this.runAfterConfiguration == RequestArgRun.ALL ||
                         this.waitPausedNeeded
@@ -794,6 +797,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                             await mi.sendExecRun(this.gdb);
                         }
                     }
+                    this.setConfiguringStateDone();
                 } else if (this.waitPausedNeeded) {
                     if (this.gdb.isNonStopMode()) {
                         await mi.sendExecContinue(
@@ -1731,7 +1735,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.configuringState = ConfiguringState.FINISHING;
                 await this.continueIfNeeded();
             } else {
-                this.configuringState = ConfiguringState.DONE;
+                this.setConfiguringStateDone();
             }
             this.sendResponse(response);
         } catch (err) {
@@ -3209,7 +3213,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     }
                 }
                 this.updateIsRunning();
-                if (this.isInitialized) {
+                if (this.configuringState == ConfiguringState.DONE) {
                     this.handleGDBResume(resultData);
                 } else {
                     // while we are deferring events, cancel previous stop events for the same thread
@@ -3288,7 +3292,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                         newThreadsConfirmed ||
                         (wasRunning && !this.isRunning))
                 ) {
-                    if (this.isInitialized) {
+                    if (this.configuringState == ConfiguringState.DONE) {
                         this.handleGDBStopped(resultData);
                     } else {
                         this.deferredStopEvents.push(resultData);
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -2385,15 +2385,16 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     /**
      * Send command to backend.
      * @param expression Command to be executed
+     * @returns Promise for: - MI response as JS object if MI command, - `undefined` if CLI command
      */
     protected async sendCommandToGdb(
         gdb: IGDBBackend,
         expression: string,
         frameRef: FrameReference | undefined
-    ): Promise<void> {
+    ): Promise<any> {
         if (expression.startsWith('-')) {
             // GDB/MI command
-            await gdb.sendCommand(expression);
+            return await gdb.sendCommand(expression);
         } else {
             // GDB CLI command
             await mi.sendInterpreterExecConsole(gdb, {
@@ -2444,9 +2445,13 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         frameRef: FrameReference | undefined
     ): Promise<void> {
         const trimmedExpression = expression.trim();
-        await this.sendCommandToGdb(this.gdb, trimmedExpression, frameRef);
+        const result = await this.sendCommandToGdb(
+            this.gdb,
+            trimmedExpression,
+            frameRef
+        );
         response.body = {
-            result: '\r',
+            result: result ? JSON.stringify(result) : '\r',
             variablesReference: 0,
         };
         await this.sendCommandToOtherGdbs(
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -42,6 +42,7 @@ import {
     ObjectVariableReference,
     MemoryRequestArguments,
     CDTDisassembleArguments,
+    RequestArgRun,
 } from '../types/session';
 import { IGDBBackend, IGDBBackendFactory } from '../types/gdb';
 import { getInstructions } from '../util/disassembly';
@@ -177,6 +178,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
 
     protected updateThreadInfo: 'missing' | 'when-requested' | 'never' =
         'missing';
+    protected runAfterConfiguration: RequestArgRun = RequestArgRun.ALL;
 
     /**
      * State variables for pauseIfNeeded/continueIfNeeded logic, mostly used for
@@ -211,6 +213,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     // and configurationDone response) we are
     protected configuringState: ConfiguringState = ConfiguringState.INITIAL;
     protected isInitialized = false;
+    protected deferredStopEvents: any[] = [];
 
     /**
      *  customResetCommands from launch.json
@@ -350,6 +353,16 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                     "Valid values are: 'missing', 'when-requested', or 'never'."
             );
         }
+        if (
+            args.run !== undefined &&
+            args.run !== RequestArgRun.ALL &&
+            args.run !== RequestArgRun.PRESERVE
+        ) {
+            throw new Error(
+                `Invalid value for 'run': '${args.run}'. ` +
+                    "Valid values are: 'all', 'preserve'."
+            );
+        }
     }
 
     /**
@@ -363,6 +376,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.steppingResponseTimeout =
             args.steppingResponseTimeout ?? DEFAULT_STEPPING_RESPONSE_TIMEOUT;
         this.updateThreadInfo = args.updateThreadInfo ?? 'missing';
+        this.runAfterConfiguration = args.run ?? RequestArgRun.ALL;
     }
 
     /**
@@ -595,6 +609,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         }
         this.sendEvent(new InitializedEvent());
         this.isInitialized = true;
+        for (const resultData of this.deferredStopEvents) {
+            this.handleGDBStopped(resultData);
+        }
+        this.deferredStopEvents.splice(0);
     }
 
     protected async attachRequest(
@@ -766,10 +784,15 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             if (this.pauseCount === 0) {
                 if (this.configuringState === ConfiguringState.FINISHING) {
                     this.configuringState = ConfiguringState.DONE;
-                    if (this.isAttach) {
-                        await mi.sendExecContinue(this.gdb);
-                    } else {
-                        await mi.sendExecRun(this.gdb);
+                    if (
+                        this.runAfterConfiguration == RequestArgRun.ALL ||
+                        this.waitPausedNeeded
+                    ) {
+                        if (this.isAttach) {
+                            await mi.sendExecContinue(this.gdb);
+                        } else {
+                            await mi.sendExecRun(this.gdb);
+                        }
                     }
                 } else if (this.waitPausedNeeded) {
                     if (this.gdb.isNonStopMode()) {
@@ -3188,6 +3211,20 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.updateIsRunning();
                 if (this.isInitialized) {
                     this.handleGDBResume(resultData);
+                } else {
+                    // while we are deferring events, cancel previous stop events for the same thread
+                    if (this.deferredStopEvents.length > 0) {
+                        if (resultData['thread-id'] === 'all') {
+                            this.deferredStopEvents.splice(0);
+                        } else if (resultData['thread-id'] !== undefined) {
+                            this.deferredStopEvents =
+                                this.deferredStopEvents.filter(
+                                    (stopResultData) =>
+                                        stopResultData['thread-id'] !==
+                                        resultData['thread-id']
+                                );
+                        }
+                    }
                 }
                 break;
             case 'stopped': {
@@ -3253,6 +3290,8 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 ) {
                     if (this.isInitialized) {
                         this.handleGDBStopped(resultData);
+                    } else {
+                        this.deferredStopEvents.push(resultData);
                     }
                 }
                 break;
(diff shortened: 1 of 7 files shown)


cwalther:
Force push to fix tests on Windows (skip tests that require async with GDB versions that don’t support it).

It also occurs to me that I haven’t added any tests for commit 3 (send run instead of continue) yet. Will try that in an additional commit. This doesn’t affect the rest.

cwalther:
Tests for commit 3 added. No more changes planned for now, ready for review.

jreineckearm:
Had a first look at the implementation and the direction looks good. Thanks!

Would like to test a couple of things and how this could map to some of our connection strategies before approval.

Regarding elevating the first DAP `continue` request to a GDB `run` command: would this impact all threads given it doesn't pass a thread ID? Or is behavior GDB server specific? (Note that I haven't checked the spec yet)

Name `run` looks sensible, given we are explicitly talking about an initial GDB `run` command. And something like `runAfterConnect` might become misleading (and also involves the overloaded term `connect`).

Delaying stopped events makes sense on first glance. But needs some extra manual testing for regressions.

Will let you know when I had some time for testing.

cwalther:
> Regarding elevating the first DAP `continue` request to a GDB `run` command: would this impact all threads given it doesn't pass a thread ID? Or is behavior GDB server specific? (Note that I haven't checked the spec yet)

There are no threads yet at that point, the program is not running yet. I don’t think `-exec-run` looks at `--thread` at all, at least the docs don’t mention it (but then they don’t for `-exec-continue` either).

jreineckearm:
Did some initial testing: `always` mode seems to operate like before, couldn't spot any regressions yet.

For `preserved`, `stopped` events before the end of the configuration phase are not captured and sent later. I am using an all-stop target (single thread after connect) and I see the expected GDB notifications coming. But something's not right with the `this.isRunning` logic at the end of the handler.

Got interrupted while investigating and calling a day soon. Will have another look unless something comes to mind on your end.

cwalther:
By “not captured and sent later”, do you mean that they are sent immediately, or not sent at all? Could be that I messed up all-stop mode, as I only use non-stop (tests should cover it, but maybe not thoroughly enough). I will take a look (results unlikely before Monday).

cwalther:
Okay, I can see in _test-logs/attach remote/remote/can attach without continuing stopped threads.log_ that it actually reproduces the problem. No `stopped` event is sent to the client during the handling of `configurationDone`, unlike in the non-stop case. Maybe I should check for DAP events in the tests after all and not just rely on `-thread-info`.

jreineckearm:
I see you have reproduced it....

```
                if (
                    !suppressHandleGDBStopped &&
                    (this.gdb.isNonStopMode() ||
                        (wasRunning && !this.isRunning))
                ) {

```

This condition at the end of `handleGDBAsync()` becomes `false`. I suspect because `this.isRunning` is still at it's default value of `false` before evaluating the first event. Probably need to make it tristate (`boolean`|`undefined`) to understand if we are looking at an uninitialized value or a known `false` state. Should come with minimal impact to existing logic, only the need to explicitly check for `undefined` in places like here.

No need to rush, just wanted to catch up with PRs.

cwalther:
Thinking out loud, I am wondering what the purpose of the condition `wasRunning && !this.isRunning` is anyway. In what circumstances, other than this initialization situation, can it even be false, and why would we want to suppress the event in that case? Don’t bother trying to answer that, I will probably figure it out on my own. But I’m off now.

jreineckearm:
Got a theory but would need to confirm before sharing. ;-)
Enjoy a long weekend!

cwalther:
Note: I am still working on this. My first attempt fixed all-stop but broke sync mode.

cwalther:
Regarding my above question: I guess the reasoning behind the `wasRunning && !this.isRunning` condition was to avoid unnecessary invalidation of object references (handles) as done in `sendStoppedEvent()`. (Maybe @jonahgraham who wrote it in 50e38d392b7c8ad9b2e4fa32c56766fe5b694faf has any thoughts.) I am still having a hard time though thinking of situations where `wasRunning` would be false (`!this.isRunning` is necessarily true), other than the situation in question, the transition from not-running-because-there-are-no-threads to not-running-because-there-are-stopped-threads, where we do need the event. Possibly something involving the late out-of-band errors from #415, the only place where `isRunning` is written outside of `updateIsRunning`.

Anyway, I hesitate to consider this reason enough to throw it out entirely, so here is a tri-state solution, inserted as an additional commit before the others to make sure it doesn’t break anything by itself. Let’s call it “[Commit 0](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/548/commits/ba03060d48cdba663a6ebfede4c4b3458a743d08)” so I don’t have to renumber all the others.

Also fix an issue with canceling deferred events when thread-id=all (occurs in all-stop) ([Commit 1](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/548/commits/4385793b277eec616497ff61f7292a4acdb22bec)), add some DAP event checks to the new tests that would have caught the all-stop breakage (Commits [6](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/548/commits/b09323dd619c2221bf6514e1e178b2fd5eb4f0fc)–[7](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/548/commits/172526773848f7560c9938ffad7a2d1d3ff6c9c3)), and rebase on main. (See `git range-diff 204bc01...1725267`.)

I also have [another commit on top of Commit 0 (“Commit 0.5”)](https://github.com/indel-ag/cdt-gdb-adapter/compare/keepstopped...indel-ag:cdt-gdb-adapter:refs/heads/awaitrunningconfirmation) that is not included here, and I am uncertain whether to include it. (They were not developed in that order, it was a required part in an earlier attempt.) Any opinions would be appreciated. On the one hand, I think it makes things slightly more correct in terms of deciding whether to pause before setting breakpoints. On the other hand, it makes things more complicated, increasing the likelihood of introducing new bugs, and it cannot completely solve the fundamental problem that `isRunning` is generally a snapshot from the past, and we may always make wrong decisions based on it in race conditions where things have already changed on the target but we have not received or processed the events yet that tell us about it. I tend toward leaving it out as long as we don’t have a demonstrable case where it fixes a practical problem.

jreineckearm:
Thanks for the updates, @cwalther ! Will have another look as soon as I can.

Regarding the reason for this: Like you, I strongly suspect that this is to suppress unnecessary window updates in the IDE and the related chain of scopes, threads, stack, variables, etc requests. They can be quite expensive already on desktop with a debugger attached via USB. And I can imagine this is getting worse in a browser environment and/or a genuine remote connection of the debugged HW.

cwalther:
Thanks, take your time. I will be away again Friday through Monday. Will probably attend the meeting on Monday but will not be in the office. I’m sorry to unleash such a monster on you, it keeps getting bigger, but since you said you are interested in the functionality as well I am hoping it will benefit all of us.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -3142,15 +3194,35 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
-            this.isRunning = this.threads.length ? true : false;
-            for (const thread of this.threads) {
-                if (!thread.running) {
-                    this.isRunning = false;
-                }
+    private updateIsRunning() {
+        let haveStopped = false;
+        let haveRunning = false;
+        let haveUnknown = false;
+        for (const thread of this.threads) {
```
jreineckearm: Nit: Probably equally fine to have three `some` statements:. But means multiple iterations over the same array. So, no real need to bother with this comment (prettier would probably make them multi-line statements anyway).
```
  const haveStopped = this.threads.some(thread => thread === false);
  const haveRunning = this.threads.some(thread => thread === true);
  const haveUnknown = this.threads.some(thread => thread === undefined);
```

jreineckearm (commented):
Another code review. Looks good.
Another test follows later today or tomorrow morning. (Sorry, I know I said I'd look at this earlier this week).

jreineckearm:
I'd like to look at https://github.com/indel-ag/cdt-gdb-adapter/tree/awaitrunningconfirmation separately. Main concern is introducing regressions due to changed timings as you stated yourself.

src/gdb/GDBDebugSessionBase.ts
cwalther: Good point, I like that, it’s more concise. Also, the variables can then be inlined, so that some of the multiple iterations may be skipped. prettier does not mangle it.

```typescript
        const newIsRunning = this.threads.some((t) => t.running === false) // have stopped
            ? false
            : this.threads.some((t) => t.running === undefined) // have unknown
              ? undefined
              : this.threads.some((t) => t.running === true); // have running
```

Will add that next time I rebase.

cwalther:
Version 4 (`git range-diff 1725267...59194fe`):
- Streamline `updateIsRunning` as per review comment (no functional change).
- Rebase on main.

jreineckearm (approved):
Tested the behavior with an all-stop target and appears to do what is described. I don't have a non-stop target to test its behavior, hence rely on testing from Christian and others.
- Our GDB remote server in all-stop mode always stops the target.
- When I launch/attach with `run`: "always" it behaves as before and runs the target after the `initCommands`.
- When I launch/attach with `run`: "preserve" it now keeps the target in the stopped state as forced by the remote server on connection.

Previous problems with not seeing the stopped state in the IDE is fixed now. Thanks!

Some second thoughts on the option names:
`always` and `preserve` feel like a bit of a mismatch: `always` refers to the "run" operation, `preserve` refers to the thread states.
I was initially thinking about a better name for `preserve`. But on second thought, the setting is more about the thread states, not the run operation. Maybe we should use `all` instead of `always` to better reflect that? That also leaves the door open for something like `selected threads`.

Note that these second thoughts are not overly strong. Hence approving in case you decide to stay with what you have.

cwalther:
> `always` and `preserve` feel like a bit of a mismatch: `always` refers to the "run" operation, `preserve` refers to the thread states.
> I was initially thinking about a better name for `preserve`. But on second thought, the setting is more about the thread states, not the run operation. Maybe we should use `all` instead of `always` to better reflect that? That also leaves the door open for something like `selected threads`.

Good point, thanks for helping my too-deep-in-the-weeds thinking out with a fresh mind. I like the idea of changing `always` to `all`. In my mind, the setting _was_ more about the run operation (that bugged me) than about the thread states, but if interpreting it the latter way allows us to choose more consistent naming, then it’s perhaps the better way. My documentation descriptions in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/238 were also written from the former point of view, so I guess I should try to reword those as well. I have tentatively done that in [87eb439](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/238/commits/87eb439c0cd152f0494337d6c56eb40598be3a7a). Does that sound good to you?

As a further test, let’s think about what the other two hypothetical options I describe in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/528#issuecomment-4371840233 would be named: “always stop” would obviously be `none`. For “never send continue” I struggle to come up with anything (`uninterrupted`?), but maybe that is a useless option anyway. In the “inconsistent” way I would have called them `stop` and `never`.

jreineckearm:
> I have tentatively done that in [87eb439](https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/238/commits/87eb439c0cd152f0494337d6c56eb40598be3a7a). Does that sound good to you?

The `gdb` > `launch` description is still a bit off. But personally I'd be OK for the time being to accept that. 
It could be handled by different configuration values for that particular case and treat them as aliases internally. Strictly speaking, there we'd have `all`. And instead of `preserve` it could be `manually` or `later` because the point in time it does the "run" is coupled with the next "continue"-like operation.

Agree, the "always stop" could be `none` or `interrupt`/`stop` because it actively brings running threads into an interrupted state (opposed to preserving their state).

"never send continue": IIRC, this happens if we pause the target while connecting but do not set it running again, right? I am still trying to understand the use case except from stop hiding when we unexpectedly mess with the target state during connection. Which rather sounds like if someone is concerned, then they'd rather want us to stop silently pausing the thread.
I'd probably not worry too much about this case. And if someone needs it in future, then they can probably help us with finding a suitable name. 🙂

cwalther:
I agree that `gdb launch` is the hard one. The real anomaly there is that our `gdbtarget launch` acts more like an attach than a launch because it interprets “launch” as “launch gdbserver”, not “launch the program”. If that were not the case, I would agree with having different values for the two `launch` cases. But now that three out of four cases are the same, I am inclined to consider it a feature that the option values are the same in all four cases, even if they don’t quite fit in one.

Personally, it’s more the `all` that bugs me for `gdb launch`, because you can’t start any less than “all” of the program, whereas “preserving” GDB’s original state sounds okay to me. But `all` is the default value so nobody actually needs to write it and be confused by it.

If you have any additional suggestions for how to improve the `gdb launch` description, I am happy to accept them. Myself, I can’t right away think of anything that sounds like an improvement to me. Also, does anyone actually use `gdb launch`? It seems like most users that I hear of use this adapter for remote-debugging embedded systems with `gdbtarget`. So maybe this is not worth putting additional effort into.

> "never send continue": IIRC, this happens if we pause the target while connecting but do not set it running again, right? I am still trying to understand the use case except from stop hiding when we unexpectedly mess with the target state during connection.

I don’t have a use case either (and am not planning to implement this, I am just trying to be prepared for the case that someone might want to implement it, or something like it), but yes, the idea came from a desire “don’t mess with the target any more than necessary”.

---

Anyway, here is version 5 (`git range-diff 59194fe...8dc71ab`):
- Change `always` to `all`.
- Add PR and commit number to commit messages, because GitHub’s _Rebase and merge_ does not do that (unlike _Squash and merge_).

If there are no last-minute objections, I will merge that.

jreineckearm (approved):
OK, I see now why you had `always` in the beginning. I could live with that, too. So, if you prefer to go with the original names, then I am good. In any case approved, functionality did look good.

@asimgunes , Paul Bell mentioned that Renesas may want to test this one, too. Not sure if that concerns you or someone else.

cwalther:
Let’s not do any further back-and-forth unless Renesas or anyone else has a strong opinion. We can still add `always` as an alias for `all` later if desired. Giving Asim a day or two to respond – I assumed that Paul would have alerted him already, but maybe not.

@jreineckearm, how are your plans for a release going? Can I see somewhere in your new planning boards what is planned for that? Or is that not among their purposes? (I am unfamiliar with “projects” on GitHub.) I am planning to finish the other work required on my side for my users to take advantage of this in the next few days, after that a release would be useful for me. But no pressure.

jreineckearm:
On the release: I'd like to make one soon, a couple of fixes and enhancements have gone in already since last release. But realistically happens earliest next week.

Yes, the board (or related Milestones) should eventually indicate such things. At least track which tickets are planned to go into which release. But internal work made me stall progress on the board again... 🙁

What I would like to get in from our side is Omar's follow-up PR on the active source highlighting in cdt-gdb-vscode. But that should be closed out soon.

Another PR coming soon is about auto-completion for the Debug Console. That can be released separately though if we don't finish the work until then. That's all I know of what's in flight.

datphan9798:
Hi @cwalther , @jreineckearm ,
I did some testing with this PR, and I didn't notice any impact on our attach use cases. Looks good from what I've seen.

jreineckearm:
Thanks a lot, @datphan9798 ! Good to hear it doesn't break anything for the Renesas use case.
Should be good to merge then.

cwalther:
Thank you @datphan9798 and @jreineckearm for the reviews and testing! Merging.

PR #550, opened 2026-06-09
Make capabilities extensible by subclasses

cwalther:
Third-party debug session subclasses may want to advertise different capabilities to clients in the `initialize` handshake than GDBDebugSessionBase. However, doing that by overriding `initializeRequest()` is inconvenient, because such an override cannot call the superclass to retrieve its capabilities and then modify them before sending them to the client, because the superclass method already sends its own result to the client. Therefore factor out the filling-in of capabilities into a side-effect-free method that can be more conveniently overridden and super-called.

I was not sure whether the new `getCapabilities` method should _return_ the capabilities as a new object or fill them into an existing object, and chose the latter just because that matches the previous code more closely. I cannot think of much compelling reason for or against either, the ease of overriding is about the same – any opinions?

(The reason I want this is that I want to turn off `supportsTerminateRequest` and `supportTerminateDebuggee`, because my debuggee is an embedded system that cannot terminate, and I am currently working on a `launch` debug configuration, for which VS Code would otherwise put “terminate” on the prominent button in the debug toolbar and “disconnect”, which I want, only on its secondary function. That wouldn’t currently actually be much of a problem, because currently “terminate” eventually also results in a disconnection, but as I argue in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/522#issuecomment-4236866156, I think that is wrong.)

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -427,25 +427,29 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         this.supportsGdbConsole =
             os.platform() === 'linux' && this.supportsRunInTerminalRequest;
         response.body = response.body || {};
-        response.body.supportsConfigurationDoneRequest = true;
-        response.body.supportsEvaluateForHovers = true;
-        response.body.supportsSetVariable = true;
-        response.body.supportsConditionalBreakpoints = true;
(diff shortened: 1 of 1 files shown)


jreineckearm (approved):
Thanks, looks good to me!

Modifying the object in `getCapabilities` is fine with me.

cwalther:
Merging in absence of objections, we can still change it if any still come up.

PR #552, opened 2026-06-10
GDB server needs timeout to exit gracefully when disconnect

datphan9798:
When disconnecting, the GDB server may be terminated before it finishes cleanup. Add a timeout to allow it to exit gracefully.

### DIFF
--- a/src/desktop/GDBTargetDebugSession.ts
+++ b/src/desktop/GDBTargetDebugSession.ts
@@ -915,7 +915,7 @@ export class GDBTargetDebugSession extends GDBDebugSession {
         args: DebugProtocol.DisconnectArguments
     ): Promise<void> {
         try {
-            await this.doDisconnectRequest(0, args?.terminateDebuggee);
+            await this.doDisconnectRequest(this.serverDisconnectTimeout, args?.terminateDebuggee);
             if (this.sessionInfo.disconnectError) {
                 this.sendErrorResponse(
                     response,


jreineckearm:
I believe this is the point of the `disconnect` request. See `Debug session end` on https://microsoft.github.io/debug-adapter-protocol/overview . The `terminate` request is the one that should allow a graceful shutdown. IIRC, VS Code will also just terminate the debug adapter process without any further warning if the `disconnect` takes too long. Are you making use of `terminate` with a timeout?

> Debuggee launched: if a debug adapter supports the [terminate](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Terminate) request, the development tool uses it to terminate the debuggee gracefully, i.e. it gives the debuggee a chance to cleanup everything before terminating. If the debuggee does not terminate but continues to run (or hits a breakpoint), the debug session will continue, but if the development tool tries again to terminate the debuggee, it will then use the [disconnect](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect) request to end the debug session unconditionally. The disconnect request is expected to terminate the debuggee (and any child processes) forcefully.

datphan9798:
Hi @jreineckearm ,
> Are you making use of terminate with a timeout?

Yes, I am using `terminate` with the timeout.
Thank you for your explanation! I understand. I will close this PR.

PR #553, opened 2026-06-11
Fix variables view for anonymous unions/structs

### LINKED ISSUE #230: Unable to inspect variables inside anonymous union/struct in VSCode debugger
**Description**: When a struct contains an anonymous union or struct, the debugger fails to display or evaluate the variables inside that anonymous member. The members are missing from the VARIABLES view and cannot be accessed via the debug console.

**Expected behavior**: The debugger should show the anonymous union/struct members as if they belong directly to the parent struct, allowing full inspection and evaluation.

cwalther:
Fixes https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/issues/230.

- There was some mix-up between varobj names (internal machine-readable identifiers) and user-visible names. The former are in field `name` in MI and not present in DAP. The latter are in field `exp` in MI and in field `name` in DAP. This only worked by accident when the two were the same, but they are not for anonymous unions or structs
- There were several things wrong with computing the `evaluateName` for anonymous unions/structs and their members. This is what is used when choosing _Copy as Expression_ from the context menu on an entry in the Variables view. I don’t know if it is used for any other purposes.

Changes are split into individual commits for easier review, they can be squashed when merging. See individual commit messages for details.

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -3638,9 +3638,20 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             });
         }
         // Grab the full path of parent.
-        const topLevelPathExpression =
-            varobj?.expression ??
-            (await this.getFullPathExpression(parentVarname, gdb));
+        let topLevelPathExpression = varobj?.expression;
+        // When -var-info-path-expression returns '', it's probably an anonymous
+        // struct or union - strip trailing .-separated components from the
+        // varobj name until we reach a non-anonymous ancestor
+        while (!topLevelPathExpression && parentVarname) {
+            topLevelPathExpression = await this.getFullPathExpression(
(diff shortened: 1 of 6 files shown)


jreineckearm:
Thanks for the fix, code changes look good. We need to check a couple of views on our end that make use of the returned `evaluateName` etc.

thorstendb-ARM:
Tested with Arm RTX and FreeRTOS across Component Viewer, XRTOS Views, generic flashing, and debugging workflows.

thorstendb-ARM (approved):
LGTM

jreineckearm (approved):
Thanks for helping with review and testing, Thorsten!

Also did a quick check with the RTOS Views extension and RTX. That extension initially uncovered problems in this area.

Good to go based on a quick code review and Thorsten's test results.

cwalther:
Thanks for testing!

PR #559, opened 2026-07-02
support Completion requests

### LINKED ISSUE #560: Add support for completions for GDB CLI in the debug console
I think it would be a nice feature if the adapter would support auto completion features for users using GDB CLI in the debug console.

omarArm:
Fixes https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/560
I added a handler for `completionsRequest` as per the MS-DAP [documentation](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Completions) and GDB/MI [documentation](https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Miscellaneous-Commands.html#GDB_002fMI-Miscellaneous-Commands).

The `completion` request is called whenever users start typing in the debug console.

Note: VSCode caches the completions returned and does not call a new `completion` request unless the user closed the open list of suggestions.

Note: So far this only works for actual GDB commands starting with the `>` operator. I am planning to extend it to support simple expressions that do not have to be prepended with any operators

image

### DIFF
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -880,7 +880,10 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             const effectiveColumn = this.columnStartAt1
                 ? args.column - 1
                 : args.column;
-            if (!text.startsWith('>') || args.column <= beginningOfCommand) {
+            if (
+                !text.startsWith('>') ||
+                effectiveColumn <= beginningOfCommand
+            ) {
                 // All GDB commands must start with a '>' character. If expression doesn't, return no completions.
                 this.sendResponse(response);
                 return;
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -880,10 +880,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             const effectiveColumn = this.columnStartAt1
                 ? args.column - 1
                 : args.column;
-            if (
-                !text.startsWith('>') ||
-                effectiveColumn <= beginningOfCommand
-            ) {
+            if (!text.startsWith('>') || effectiveColumn < beginningOfCommand) {
                 // All GDB commands must start with a '>' character. If expression doesn't, return no completions.
                 this.sendResponse(response);
                 return;
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -888,9 +888,11 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 return;
             }
             const trimmedWhiteSpaceLength = args.text.length - text.length;
-            const endOfCommandToComplete =
-                effectiveColumn - trimmedWhiteSpaceLength;
-            const commandToComplete = text.slice(1, endOfCommandToComplete);
+            // Slicing from > to end of command to complete
+            const commandToComplete = args.text.slice(
+                trimmedWhiteSpaceLength + 1,
+                effectiveColumn
+            );
             const completions = await mi.sendCompletions(
                 this.gdb,
                 commandToComplete
@@ -902,7 +904,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                         length: commandToComplete.length,
                         start:
                             trimmedWhiteSpaceLength +
-                            (this.columnStartAt1 ? 2 : 1), // +1 for the '>' character
+                            (this.columnStartAt1 ? 2 : 1), // +1 for the '>' character if 0-based, +2 if 1-based
                     };
                 }),
             };
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -160,6 +160,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
     // threads yet, this is the last certain value.
     protected isRunning = false;
 
+    protected columnStartAt1 = true;
     protected supportsRunInTerminalRequest = false;
     protected supportsMemoryReferences = false;
     protected supportsMemoryEvent = false;
@@ -441,6 +442,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             args.supportsRunInTerminalRequest === true;
         this.supportsMemoryReferences = args.supportsMemoryReferences === true;
         this.supportsMemoryEvent = args.supportsMemoryEvent === true;
+        this.columnStartAt1 = args.columnsStartAt1 === true;
         this.supportsGdbConsole =
             os.platform() === 'linux' && this.supportsRunInTerminalRequest;
         response.body = response.body || {};
@@ -465,6 +467,7 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
         cap.supportsTerminateRequest = this.isRemote;
         cap.supportsDataBreakpoints = true;
         cap.supportsBreakpointLocationsRequest = true;
+        cap.supportsCompletionsRequest = true;
         cap.supportTerminateDebuggee = this.isRemote;
         cap.breakpointModes = this.getBreakpointModes();
     }
@@ -858,6 +861,66 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
 
         // The promise resolves when pushed resolveFunc gets called.
         return pausePromise;
+    }
+
+    protected async completionsRequest(
+        response: DebugProtocol.CompletionsResponse,
+        args: DebugProtocol.CompletionsArguments
+    ): Promise<void> {
+        try {
+            if (!this.canRequestProceed()) {
+                this.logger.verbose(
+                    'Debug adapter cannot process completions request, skipping it.'
+                );
+                this.sendResponse(response);
+                return;
+            }
+            const text = args.text.trimStart();
+            const effectiveColumn = this.columnStartAt1
+                ? args.column - 1
+                : args.column;
+            if (
+                !text.startsWith('>') ||
+                args.column <= args.text.indexOf('>') + 1
+            ) {
+                // All GDB commands must start with a '>' character. If expression doesn't, return no completions.
+                this.sendResponse(response);
+                return;
+            }
+            const trimmedWhiteSpaceLength = args.text.length - text.length;
+            const endOfCommandToComplete =
+                effectiveColumn - trimmedWhiteSpaceLength;
+            const commandToComplete = text.slice(1, endOfCommandToComplete);
+            const completions = await mi.sendCompletions(
+                this.gdb,
+                commandToComplete
+            );
+            response.body = {
+                targets: completions.matches.map((completion) => {
+                    return {
+                        label: completion,
+                        length: commandToComplete.length,
+                        start:
+                            trimmedWhiteSpaceLength +
+                            (this.columnStartAt1 ? 2 : 1), // +1 for the '>' character
+                    };
+                }),
+            };
+            this.sendResponse(response);
+        } catch (err) {
+            if (err instanceof Error && err.message.includes('complete')) {
+                response.body = {
+                    targets: [],
+                };
+                this.sendResponse(response);
+            } else {
+                this.sendErrorResponse(
+                    response,
+                    1,
+                    err instanceof Error ? err.message : String(err)
+                );
+            }
+        }
     }
 
     protected async dataBreakpointInfoRequest(
--- a/src/gdb/GDBDebugSessionBase.ts
+++ b/src/gdb/GDBDebugSessionBase.ts
@@ -876,13 +876,11 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 return;
             }
             const text = args.text.trimStart();
+            const beginningOfCommand = args.text.indexOf('>') + 1;
             const effectiveColumn = this.columnStartAt1
                 ? args.column - 1
                 : args.column;
-            if (
-                !text.startsWith('>') ||
-                args.column <= args.text.indexOf('>') + 1
-            ) {
+            if (!text.startsWith('>') || args.column <= beginningOfCommand) {
                 // All GDB commands must start with a '>' character. If expression doesn't, return no completions.
                 this.sendResponse(response);
                 return;
--- a/src/mi/miscellaneous.ts
+++ b/src/mi/miscellaneous.ts
@@ -1,0 +1,25 @@
+/*********************************************************************
+ * Copyright (c) 2026 Arm Limited and others
+ *
+ * This program and the accompanying materials are made
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *********************************************************************/
+
+import { IGDBBackend } from '../types/gdb';
+
+export interface MICompletion {
+    completion?: string;
+    matches: string[];
+    max_completions_reached: string;
+}
+
+export async function sendCompletions(
+    gdb: IGDBBackend,
+    command: string
+): Promise<MICompletion> {
+    const miCommand = `-complete "${command.replace(/(["\\])/g, '\\$1')}"`;
+    return gdb.sendCommand<MICompletion>(miCommand);
+}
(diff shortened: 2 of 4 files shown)


omarArm:
@datphan9798 I couldn't request your review from github, so I am mentioning you here as I thought it might be interesting for you to review this one

src/mi/miscellaneous.ts
```diff
@@ -0,0 +1,25 @@
...
+ * available under the terms of the Eclipse Public License 2.0
+ * which is available at https://www.eclipse.org/legal/epl-2.0/
+ *
+ * SPDX-License-Identifier: EPL-2.0
+ *********************************************************************/
+
+import { IGDBBackend } from '../types/gdb';
+
+export interface MICompletion {
+    completion: string;
```
cwalther:
```suggestion
    completion?: string;
```

src/mi/miscellaneous.ts
```diff
@@ -0,0 +1,25 @@
...
+    matches: string[];
+    max_completions_reached: string;
+}
+
+export async function sendCompletions(
+    gdb: IGDBBackend,
+    command: string
+): Promise<MICompletion> {
+    const miCommand = `-complete "${command}"`;
```
cwalther: This is wrong for commands that contain quotation marks. I guess the correct solution would be the same as in #507 but I haven’t checked.

cwalther (changes requested):
This does not work correctly for me.

When I type `>int`, it gives me
Image
(so far, so good), but when I then type _tab_ to accept, I get
`>intinterpreter-exec`

src/mi/miscellaneous.ts
omarArm: I changed it now

src/mi/miscellaneous.ts
omarArm: oh sorry about that, I will check your old PR and fix it accordingly

omarArm:
@cwalther that is weird, it works just as expected on my machine. I press `tab` and it fills in with whatever is highlighted on the list. Let me try and test it on a different machine, maybe I will catch something


Edit: I can reproduce it now, I am investigating how to fix it

Edit: I think I fixed the problem, let me know if the issue persists

cwalther:
It works in the case I showed above now, but I can still come up with situations where it doesn’t.

1. Type `> interru`
2. _esc_ to dismiss completion
3. 4× _←_ so you have `> int|erru` (where `|` is the cursor)
4. _ctrl-space_ to invoke completion, it suggests `interrupt`
5. _tab_ to accept
6. Actual result: `interrupt|erru`
Whether leaving the `erru` at the end is correct is debatable (GDB does the same in its CLI), but erasing the `> ` at the beginning is certainly wrong.

To behave the same as the GDB CLI, I guess you should only take text up to the cursor as `commandToComplete`.

You don’t take into account `line`, but then we don’t seem to support multi-line input anyway (you can do that with shift-return in VS Code). It would be nice to avoid the _GDB command completion failed_ errors in that case though.

omarArm:
@cwalther Thanks for pointing this one out. I don't take into account the parameter `line` indeed because we don't allow multi line commands anyway. Now the problem is that when you have the cursor in between the `text`, the `column` argument has a value lower than that of the length of the `text`, which entails a negative `start`, which then becomes set to `1` anyway. I changed it so that it falls back to `2` or `3` depending on whether or not the user had a white space in between the `>` and the rest of the text or not. Obviously, that would cause a problem that if users have more than a single white space before the actual expression, then that would be changed to a single space, but I guess that is an acceptable trade off?

cwalther:
I have not understood yet where the trade-off is. Why can’t this be done correctly? Send the text between the `>` plus any whitespace and the cursor to GDB, and then replace that same range by the completion that comes back? Or if you insist on taking the text to the end of the line, not just to the cursor, then replace _that_ range? There is no need for `args.column` to be involved in the computation of `startingPosition`, it’s simply where the “`>` plus whitespace” ends, something like `/^\s*>\s*/.exec(args.text)[0].length`. (Caution, take into account any trimming you do – the client wants offsets in the untrimmed text.)

Edit: And don’t forget to take into account `columnsStartAt1`. In the previous (faulty) code that canceled out, but not if you do it the above way.

omarArm:
I overcomplicated things indeed. I had the idea of trimming before sending the text to GDB, which caused all sorts of problems. I think now it should be acceptable, right?

src/gdb/GDBDebugSessionBase.ts
cwalther: There is still a trimming here that you don’t seem to take into account.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -879,23 +879,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
                 this.sendResponse(response);
                 return;
             }
-            const commandToComplete = text.slice(1).trim();
+            const commandToComplete = text.slice(1, args.column - 1);
```
cwalther: The `- 1` should probably depend on `columnsStartAt1`.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -879,23 +879,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
                 this.gdb,
                 commandToComplete
             );
-            const startingPosition =
-                args.column > commandToComplete.length
-                    ? args.column - commandToComplete.length
-                    : text.slice(1)[0] === ' '
-                      ? 3
-                      : 2;
+            const startingPosition = args.column - commandToComplete.length + 1;
```
cwalther: Where does this `+ 1` come from?

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -879,23 +879,17 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
             response.body = {
                 targets: completions.matches.map((completion, index) => {
```
cwalther: Note the linter warning on this line (I see it in the diff).

cwalther (dismissed):

src/gdb/GDBDebugSessionBase.ts
omarArm: the trimming in here is for the sake of avoiding users doing something like `    > inter`. I thought that would be acceptable, isn't it?

cwalther: I am not objecting to the trimming (it should match whatever is done in `evaluateRequest`). But it needs to be taken into account in the following position calculations.

src/gdb/GDBDebugSessionBase.ts

src/gdb/GDBDebugSessionBase.ts
omarArm: thanks for point it out, adjusted it

src/gdb/GDBDebugSessionBase.ts

src/gdb/GDBDebugSessionBase.ts
omarArm: For calculating the new starting position, I am using `args.column - commandToComplete.length`. The `args.column` does count the trimmed character in the position calculation. Or am I missing something?

cwalther: Correct, that calculation is okay. But the calculation of `commandToComplete` is not (unless I am missing something, as I am unable to test it right now). Did you test this?

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,59 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+            }
+            const text = args.text.trim();
+            if (!text.startsWith('>')) {
+                // All GDB commands must start with a '>' character. If expression doesn't, return no completions.
+                this.sendResponse(response);
+                return;
+            }
+            const commandToComplete = text.slice(
+                1,
+                this.columnsStartAt1 ? args.column - 1 : args.column - 2
```
cwalther: I think the last `- 2` should be `- 0`, i.e. absent – no correction needed when columns start at 0, like everything else.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,59 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+            );
+            const completions = await mi.sendCompletions(
+                this.gdb,
+                commandToComplete
+            );
+            const startingPosition = this.columnsStartAt1
+                ? args.column - commandToComplete.length
+                : args.column - commandToComplete.length - 1;
```
cwalther: `args.column` is already in the units corresponding to `columnsStartAt1`, no need to correct for it again.

src/gdb/GDBDebugSessionBase.ts

src/gdb/GDBDebugSessionBase.ts
omarArm: I have tested it yes, I tried both test cases that you mentioned above in the PR conversation and it worked. I am not sure if there are anymore test cases that I am not aware of.

Edit: I tried adding extra whitespace before `>` and then continued with the second test case you provided and it didn't work. I will check out what might the problem be and update the code accordingly

cwalther: Yes, that’s what I meant – test with a case where the `trim` in question actually does anything.

omarArm: Thanks for your support and your patience, I think with the latest push things should work smoothly. Please test it and let me know

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,60 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                this.sendResponse(response);
+                return;
+            }
+            const text = args.text.trim();
+            if (!text.startsWith('>')) {
+                // All GDB commands must start with a '>' character. If expression doesn't, return no completions.
+                this.sendResponse(response);
+                return;
+            }
+            const trimmedWhiteSpaceLength = args.text.split('>')[0].length;
```
cwalther: `args.text.indexOf('>')` would probably be slightly more efficient.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,60 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                ? args.column - trimmedWhiteSpaceLength - 1
+                : args.column - trimmedWhiteSpaceLength;
+            const commandToComplete = text.slice(1, endOfCommandToComplete);
+            const completions = await mi.sendCompletions(
+                this.gdb,
+                commandToComplete
+            );
+            const startingPosition = this.columnsStartAt1
+                ? args.column - commandToComplete.length
+                : args.column - commandToComplete.length;
```
cwalther: ?: operator with two identical expressions is redundant.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,60 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+            response.body = {
+                targets: completions.matches.map((completion) => {
+                    return {
+                        label: completion,
+                        length: commandToComplete.length,
+                        start: trimmedWhiteSpaceLength + startingPosition,
```
cwalther: This confuses me. I think it should be just `start: startingPosition`, but presumably you tested it with all combinations of whitespace this time? (By the way, you should really put these cases in the automated tests too, so you don’t accidentally break them again.)

I am still not corrently set up to test things, and that will likely stay so for the rest of the week.

src/gdb/GDBDebugSessionBase.ts
omarArm: I accidentally erased a `-1` if `columnStartAt1` was not supported

src/gdb/GDBDebugSessionBase.ts
omarArm: I will try and add some tests sure, the only problem is that I can only add tests for the completionsRequest itself, not how vscode is going to perceive them

omarArm: I changed the variable name as well

src/gdb/GDBDebugSessionBase.ts
omarArm: good catch

src/gdb/GDBDebugSessionBase.ts
cwalther: That was not accidental, that was correct, done in response to my comment https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/559#discussion_r3536874583.

cwalther:
You seem to be getting lost in a mess. I made this sketch while reviewing the code, maybe something like this would help you too? My scrawl is unreadable, but I hope you get the idea.
IMG_2369

jreineckearm:
Gave it a try with https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/559/commits/17e8e5cf54dd02332aa4a5e4a4bc93ca789ab3ad. Was able to trigger auto-completion with this kind of string `>  mon` and the cursor position being before the `>`. This should be guarded as long as we don't have support for expressions other than GDB CLI commands.

jreineckearm:
The guard still let's some corner cases through. Let's look at it together offline.

jreineckearm:
Working better now. Please add tests with and without extra spaces before and after `>`. And differing column values on the boundaries (right before and after `>`, right before the started command, etc).

jreineckearm (approved):
Changes look OK to me. Tested various corner cases after latest fixes.

cwalther:
I have tried it now and I have to concede that it appears to work, however I don’t understand why. The way VS Code interprets the `start` value it gets makes no sense to me and does not seem to conform to the DAP specification. The values still look wrong to me, but for some reason it does the right thing anyway.

All consideration of `columnsStartAt1` has disappeared now. That can’t be right.

cwalther:
In fact, it still does the right thing even if I return `start: Math.floor(10 * Math.random())`. In other words, it appears to ignore the value entirely. I guess I’ll have to look for where this is handled in the vscode source to check what it really does.

omarArm:
I did some investigation, and apparently VSCode doesn't use `CompletionItem.start` in accordance to what the documentation is saying. As far as I can guess, it doesn't use it at all even. That is why having it wrong didn't cause any visible problems.

Anyway, I adjusted that and everything should be clean now

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,68 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                this.sendResponse(response);
+                return;
+            }
+            const text = args.text.trimStart();
+            const effectiveColumn = this.columnStartAt1
+                ? args.column - 1
+                : args.column;
+            if (
+                !text.startsWith('>') ||
+                args.column <= args.text.indexOf('>') + 1
```
jreineckearm:
```suggestion
                effectiveColumn <= args.text.indexOf('>') + 1
```

I think this has to be the effective as per `columnStartAt1` setting.

Note you can launch the test debug client with different capabilities like here: https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/blob/14926db63c2a5fb531fc8c8bc0d348c782625a71/src/integration-tests/initializeCapabilities.spec.ts#L36 So, you should be able to write testes for both `columnStartAt1` values.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -860,6 +863,69 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+                return;
+            }
+            const text = args.text.trimStart();
+            const beginningOfCommand = args.text.indexOf('>') + 1;
+            const effectiveColumn = this.columnStartAt1
+                ? args.column - 1
+                : args.column;
+            if (
+                !text.startsWith('>') ||
+                effectiveColumn <= beginningOfCommand
```
cwalther: I think this could be `<` instead of `<=` so that completing an empty command is allowed – GDB seems to support that. But that may be a matter of opinion.

cwalther (commented):
The code looks correct now, however the new tests for `columnStartAt1 = false` do not. They are identical to the ones for `columnStartAt1 = true`, which they shouldn’t be, and only pass because they actually run with `columnStartAt1 = true`. I suspect there are now two `initialize` requests of which the wrong one takes effect.

I have checked the vscode source: The relevant code is [here](https://github.com/microsoft/vscode/blob/af2c64423e7ee5d1030a000c82a0bb774d043351/src/vs/workbench/contrib/debug/browser/repl.ts#L270-L310), and it indeed does not use `start` for positioning, but effectively assumes it is always `column - length` (clamped). It does use it for `filterText`, which seems a bit strange.

src/gdb/GDBDebugSessionBase.ts

cwalther (approved):
No more objections!

I will leave the merging / final approval to @jreineckearm.

jreineckearm:
Thanks both! Let's get this now out into a release. 🙂

PR #562, opened 2026-07-13
1.10.0 Release Preparations

jreineckearm:
- Bumps version to 1.10.0
- CHANGELOG

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.10.0
 
-- Implements [`#528`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/528): Add `run` option to configure if to always run all threads after connect, or if to preserve their state encountered after connect.
+- Implements [`#528`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/528): Add `run` setting to configure if stopped threads are set running or kept stopped on attach.
 - Implements [`#539`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/539): Extend `evaluate` response by `memoryReference` field where possible and if supported by client. A client can use this to determine if an expression maps to a memory address.
 - Implements [`#550`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/550): Allow override of capabilities reported by `GDBDebugSessionBase` class.
 - Implements [`#560`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/560): Support `completions` request for GDB CLI commands (with `>` prefix). This enables auto-completion features in the IDE, for example in a debug console.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Change Log
 
+## 1.10.0
+
+- Implements [`#528`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/528): Add `run` option to configure if to always run all threads after connect, or if to preserve their state encountered after connect.
```
cwalther: This sounds a little bumpier than what you suggested in https://github.com/eclipse-cdt-cloud/cdt-gdb-vscode/pull/238/changes but maybe you deliberately wanted to include the option values in the wording.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+- Implements [`#539`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/issues/539): Extend `evaluate` response by `memoryReference` field where possible and if supported by client. A client can use this to determine if an expression maps to a memory address.
+- Implements [`#550`](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/550): Allow override of capabilities reported by `GDBDebugSessionBase` class.
```
cwalther, line 7: Is this a “notable code change”, or do we only list incompatible ones as such?

cwalther (approved):
Checked changelog against commit history, looks good to me!

cwalther:
Oh, maybe https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/commit/c120dbd117b0a20c879a99ca83e9fab75d2917c8 would be worth a separate mention, even though I snuck it in as part of something else?

And likewise https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/commit/9b887b52c791ad93189cd2e58e8ec33930a66911, although maybe that’s a bit too obscure.

CHANGELOG.md
jreineckearm: Simply had forgotten about this, should be aligned with that changelog. Thanks for reminding me. Comes with next push.

CHANGELOG.md
jreineckearm: No, anything that makes a downstream developer's life harder or easier can go. As long as we keep it within reason for a changelog/release notes. Will check the code diff since last release and your suggestions further down. Comes with next push.

jreineckearm:
> Oh, maybe [c120dbd](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/commit/c120dbd117b0a20c879a99ca83e9fab75d2917c8) would be worth a separate mention, even though I snuck it in as part of something else?
> 
> And likewise [9b887b5](https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/commit/9b887b52c791ad93189cd2e58e8ec33930a66911), although maybe that’s a bit too obscure.

The first one makes sense. For the second, probably not so much the string instead of `\r` thing, but definitely the changed return type (`void` -> `any`)

jreineckearm:
Changes in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/562/commits/fa93e92ca14f9dc454976f4ef007e645aa7f8856 . Reviewed the rest of code changes since last release. Can't see anything else to put in there except from those you mentioned. The rest goes a bit too far to explain in the changelog and need review by developers as a whole.

cwalther:
Looks good, thanks! Good point about the changed return type, I had only considered that from the point of view of the caller (where it doesn’t matter), not from the one of an overrider (where it does).

Issue #424, opened 2025-08-11
Support global and file static variables


jreineckearm:
The CDT GDB adapter currently only displays local variables and registers in the Variables view.
However, you also may want to see global variables and/or file-static variables (see [here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Variables.html))

GDB supports enumeration and access through https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Symbol-Query.html and https://sourceware.org/gdb/current/onlinedocs/gdb.html/GDB_002fMI-Variable-Objects.html#GDB_002fMI-Variable-Objects

We should add support for global/file-static variables. You currently can only inspect them from the Watch or the Debug Console if you know a certain variable name, or through GDB CLI commands.

Potential caveats
* Depending on the program, you may have a significant number of such variables. Showing them as a flat list may become expensive on Variable view updates for embedded targets due to the target accesses.
* File-statics could be always shown or only when they are in scope, i.e. PC is at code from a specific file.

Hence, we need to keep an eye on performance during work on this feature. And may need a setting as a fail-safe to turn this off.

PR #566, opened 2026-08-10
Show global variables in Variables view

datphan9798:
This PR adds support for displaying global variables in the Variables View.

A new Global section is introduced below the existing Local section. All available global variables are collected and displayed under this Global section, allowing users to view and inspect global variables directly from the Variables View during a debug session.

Changes:
- Add a new Global scope in the Variables View.
- Display global variables under the Global section.

There are some concerns about the performance of loading all global variables and their address and value if there are a lot of variables. I applied pagination for Global section in this PR, hope it could help for this case.


jreineckearm (changes requested):
Thanks for the contribution, @datphan9798 .

We had worked on something similar in the past and had similar concerns about performance. But also about how a user would nicely navigate this in the Variables window. That's when we stopped this.

I don't want to stand in the way of adding the feature. Especially since we don't offer a good alternative yet. And I am happy to review and test in the coming days. But can we please make it optional through a launch/attach configuration item?

datphan9798:
Hi @jreineckearm ,
Thank you for taking a look at this!
 I've updated it so that displaying global variables is configurable through a launch/attach option. Could you please help me check it again?

datphan9798:
Hi @asimgunes , @jonahgraham ,
Could you please help me review this PR?
Thank you!

omarArm:
I think it would be a good idea if @cwalther could give it a try. As last time I tried that, he had a proper use case that really highlighted the performance issue.

cwalther:
Will do shortly, thanks for the ping.

For reference, this is related to #424, and last time we discussed it was in https://github.com/eclipse-cdt-cloud/cdt-gdb-adapter/pull/443#issuecomment-3271108294.

cwalther:
I have tried it on my current project, which is a bit larger than the one I have previously talked about, with 15484 global variables. Findings:
- I have nothing against adding this feature if it stays gated behind the `showGlobalVariables` option. I could even live with having it on by default, since our extension programmatically sets most of the launch options anyway. But I prefer off by default as it is now.
- The `-symbol-info-variables` MI command that is used to fetch the variables still takes more than 2 minutes to execute and delivers a megabyte-sized response. The GUI is responsive during that time, but any action that results in a GDB command just stalls as the command is queued. This makes the feature unusable for us – having to wait minutes between hitting a breakpoint and being able to step ahead (with no indication that anything is going on, just an empty variables view) is inacceptable.
- The pagination helps with not creating unnecessary variable objects. Expanding one page of 100 variables takes about 8 seconds, which is acceptable.
- However, even disregarding the `-symbol-info-variables` performance, the feature seems useless in our situation, because you have no idea which of the 155 pages contain the variable that you are interested in.
- It doesn’t look like much can be done about the `-symbol-info-variables` performance, I don’t see anything in the [docs](https://sourceware.org/gdb/current/onlinedocs/gdb#index-_002dsymbol_002dinfo_002dvariables) that would allow pagination at that level or anything like that. One could pass a `--max-results` limit of 1000 or something, but I’m not sure that would achieve much, beyond preventing a user who has accidentally turned on the feature from thinking the debug session has locked up, because the resulting functionality is still useless (likely the variables you wanted wouldn’t be among the 1000 selected).

I have not looked at the code yet.

cwalther:
Would it be possible to issue the `-symbol-info-variables` command only when the _Global_ entry in the variables view is expanded? In that case the `showGlobalVariables` option might be unnecessary (or could be on by default). What does the `expensive` flag in the DAP [Scope](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Scope) object achieve in VS Code, would it cause some kind of progress indication in that case?

datphan9798:
Thank you for your checking, @cwalther !
Sorry to hear this isn't helpful for your case. I tested it with a project containing about 5,000 global variables, and -symbol-info-variables completed in roughly 0.5 seconds. That may be because the variables in my test project are all simple types. I also agree that there may not be a practical way to improve the performance of -symbol-info-variables for this particular case.

> Would it be possible to issue the `-symbol-info-variables` command only when the _Global_ entry in the variables view is expanded? In that case the `showGlobalVariables` option might be unnecessary (or could be on by default). What does the `expensive` flag in the DAP [Scope](https://microsoft.github.io/debug-adapter-protocol/specification#Types_Scope) object achieve in VS Code, would it cause some kind of progress indication in that case?

The reason I'm issuing -symbol-info-variables during Global scope initialization (in scopesRequest) is that I need the total number of global variables to enable pagination in the Global section. About `expensive` flag, marking the scope as expensive would keep the Global section collapsed by default after launch.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -3863,6 +3894,186 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+        start: number,
+        count: number
+    ): Promise<DebugProtocol.Variable[]> {
+        const variables: Array<DebugProtocol.Variable | undefined> = [];
+        const GDB_BATCH_SIZE = 100;
+        const GLOBAL_FRAME_REFERENCE = { threadId: -1, frameId: -1 };
+        const GLOBAL_DEPTH = 0;
+        if (this.auxGdb && this.isRunning) {
+            this.logger.verbose(
+                'Skipping variable frame request while target is running'
```
jreineckearm, line 3926: Probably should say "global variable frame".

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -3863,6 +3894,186 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+            );
+            return [];
+        }
+        const globalNames = await this.getGlobalSymbolNames();
+        const sliceNames = globalNames.slice(start, start + count);
+        const existingVarObjs: Array<{ varobj: VarObjType; index: number }> =
+            [];
+        const newNames: Array<{ name: string; index: number }> = [];
+
```
jreineckearm, line 3935: Haven't done a detailed comparison of the rest of the method to the existing variables handling. Just wondering if we have avoidable code duplication here.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -3943,11 +4154,44 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
...
+        if (cacheAddr) {
+            return cacheAddr;
+        }
+        const addr = await this.getAddr(varobj, gdb);
+        this.globalAddrCache.set(varobj.varname, addr);
+        return addr;
+    }
+
+    private async getGlobalSymbolNames(): Promise<string[]> {
+        if (this.globalSymbolNames.length === 0) {
```
jreineckearm, line 4186: If I understand the code correctly, we get the symbol names once and then keep the list unchanged over the debug session. What if a user loads additional symbol files from the Debug Console?

src/integration-tests/var.spec.ts
```diff
@@ -22,6 +22,8 @@ import {
     verifyRegister,
     fillDefaults,
 } from './utils';
+import { DebugProtocol } from '@vscode/debugprotocol';
+// import { ILocation } from '@vscode/debugadapter-testsupport/lib/debugClient';
```
jreineckearm, line 26: Do we need this commented out line?

jreineckearm (commented):
Feature overall working for me with a simple target project, thanks.

As @cwalther confirmed there still is potential performance concerns which can strongly depend on how your debug target is connected (remote over network, serial connection, USB, JTAG debug unit in the way, etc) and how big the application is. Hence, I'd vote for keeping the feature off by default to avoid surprises for users of slow-ish debug connections.

See a couple of comments in the code and a few more observations below:
- Did you do measurements with and without the pagination/batched accesses? Just curious if there is proof that the additional complexity helps.
- Editing global variable values through the variables window doesn't seem to work. I enter a new value but I get an error message. 

Image

- Please don't forget to add the new configuration to `cdt-gdb-vscode` (new configuration option in package.json and docs).

cwalther:
> I tested it with a project containing about 5,000 global variables, and -symbol-info-variables completed in roughly 0.5 seconds. That may be because the variables in my test project are all simple types.

Hmm, that’s quite a discrepancy. I have done a few experiments to figure out where it comes from, but have not found anything conclusive. It cannot have anything to do with the speed of the remote connection – there is no traffic on the remote connection during that time, the variable information comes from the ELF file, not from the target.

One notable point of data: it takes the same two minutes even if I say `-symbol-info-variables --max-results 20`, so that wouldn’t help.

How many source files does your 5000-variable example have? My project has 15505 variables across 1228 files, and the ELF file (with debug info) is 178 MB.

> The reason I'm issuing -symbol-info-variables during Global scope initialization (in scopesRequest) is that I need the total number of global variables to enable pagination in the Global section.

Okay, that’s a bummer. (Weird design – surely an _expensive-to-compute_ number of variables that cannot be supplied up front and a _large_ number of variables that makes pagination useful often occur together, so why make these things mutually exclusive?)

Potential ways out I can think of – up to you whether you want to pursue them:
- Add another level of hierarchy between the _Global_ scope and the individual variables, with a known number of elements (probably 1).
- Set `indexedVariables` in the original scope response to an arbitrary number that is higher than the actual number (would probably need to be user-configurable) – it looks like VS Code can deal with receiving fewer than announced.

It also occurs to me that a potential change that may even make the feature usable in our situation (at least more so than the current implementation) would be this: Instead of presenting the variables in a flat list (paginated by the client), present them in the same structured form that GDB supplies them in: grouped by source file.

jonahgraham:
> the ELF file (with debug info) is 178 MB.

Without knowing much about either example project, I wanted to highlight that the type of variables, as well as the total number of times various big includes are in the project (C++ std lib or boost), and even level of debug info (-g0 vs -g3) can all make a huge difference. 

I ran a quick experiment, for 1,500 files and 15,000 globals if I make all globals `int` it takes a fraction of a second. If I make them `std::string` and include `<string>` in each file it takes 10s of seconds.

> present them in the same structured form that GDB supplies them in: grouped by source file.

Using `-file-list-exec-source-files` can show all source files. But we still need the big query to get all variables.

---

An alternative approach is to not use gdb at all to get the names of the global variables, instead use objdump to populate a list of globals. This can be done asynchronously from any gdb query. IIUC this is how some other people are doing it (e.g. cortex debug). Note that objdump can often produce this information much faster than gdb itself can.

This information can be used in one of two ways:

A) Populate Global in the tree, perhaps by using some hierarchy so we don't have trees of 10k variables
B) Populate a chooser that adds them to the watch window with command `editor.debug.action.selectionToWatch` or `debug.addToWatchExpressions`

jreineckearm:
> An alternative approach is to not use gdb at all to get the names of the global variables, instead use objdump to populate a list of globals.

This assumes objdump to be around in your installation. Which should be fine if you use the GNU toolchain for everything. But for other (embedded) toochains such tools may be called differently, e.g. fromelf in Arm Compiler 5 with slightly different output format IIRC. Also these objdump tools may use slightly different interpretation of ELF, DWARF, etc where the specs leave room for such interpretation. That's why you normally prefer to use the debug engine itself (GDB client) as the single source of truth. Otherwise you might see inconsistent behavior at runtime.

But anyway, if it helps your use cases, it may still be worth a try.

> Populate a chooser that adds them to the watch window

Sounds like a good idea (would need to live in cdt-gdb-vscode).
We had a similar approach in past tooling where we had a Symbols window showing all functions, data symbols, types, etc. And offered a context menu to send requests to the watch window for display.
https://www.keil.com/support/man/docs/uv4cl/uv4cl_db_dbg_symwin.htm (note we have the same for our Cortex-M variant of uVision, just having trouble at the moment to access the link directly: https://support.arm.com/documentation/101407/0543/Debugging/Debug-Windows-and-Dialogs/Symbols-Window?lang=en )

One day I'd still like to have something like this again. This is why I also one day would love to have a DAP extension that enumerates global symbols without reading from target.

cwalther:
> > present them in the same structured form that GDB supplies them in: grouped by source file.
> 
> Using `-file-list-exec-source-files` can show all source files. But we still need the big query to get all variables.

Correct. That suggestion was in response to the usability issue, not to the performance issue.

jonahgraham:
@datphan9798 There is a lot of discussion on this issue. Are you clear on a path forward here? Perhaps you can share a (brief) plan for review before implementing additional changes to this PR.

datphan9798:
Hi @jonahgraham and everyone,
Sorry for keeping everyone waiting. I'm still checking and validating the approaches discussed above. Here are the changes I'm planning for the upcoming commits:

- Use objdump/nm to enumerate global variables instead of -symbol-info-variables, to address the performance concerns.
- Populate variables in the tree, grouped by source file.
- Address the remaining review comments: global variable value editing not working, handling the case of loading additional symbol files from the Debug Console, measurements with/without batched.

I will update again once these are ready for another round of review. Thank you for your patience!

datphan9798:
Hi @jonahgraham, @jreineckearm, @cwalther 

Thanks for your patience! I've pushed a new commit addressing the feedback from the previous review. In this commit:

- Added `objdumpPath `to configure the path to objdump. If `showGlobalVariables `is true and `objdumpPath `is provided, global symbols are read from the symbol file using objdump. If `showGlobalVariables `is true and `objdumpPath `is not provided, global symbols are obtained from GDB via the `-symbol-info-variables` command instead. I kept the `-symbol-info-variables` path as a fallback for when objdumpPath isn't configured, since it's still reliable and useful for smaller projects, and works regardless of the toolchain in use.
`objdumpPath `is currently only honored for target/remote debug sessions — a plain local launch always uses the GDB command regardless of `objdumpPath`. If you want to specifically test the objdump-backed path, please use a target/remote launch config.

- `GNUObjdumpSymbolReader `implements the objdump-based reader. I've tested it with GCC ARM 10.3-2021.10 (DWARF 4) and 13.3.rel1 / 14.2.rel1 (DWARF 5).

- Global variables are now grouped by source file and expanded on demand, so the batched-request optimization from the earlier design is no longer needed — I removed it.

- Fixed an issue that prevented editing Global variables.

This isn't fully resolved yet — the Windows test suite still fails in gdb-async-off mode. If you have time, I'd really appreciate you reviewing the approach and trying this commit out in your own environment.

Thanks a lot!

src/integration-tests/var.spec.ts
```diff
@@ -620,3 +626,1195 @@ describe('Variables Test Suite', function () {
...
+        INITIAL_STOP: 0,
+        RETURN: 0,
+    };
+
+    before(function () {
+        resolveLineTagLocations(varsGlobalsSrc, lineTags);
+    });
+
+    beforeEach(async function () {
+        dc = await standardBeforeEach('debugTargetAdapter.js');
```
jonahgraham, line 1122: I assume the intention is for this to work with remote and non-remote?

I suggest this and removing this from the individual tests:

```typescript
        if (!isRemoteTest) {
            this.skip();
        }
```

```suggestion
        dc = await standardBeforeEach();
```

src/integration-tests/var.spec.ts
```diff
@@ -620,3 +626,1195 @@ describe('Variables Test Suite', function () {
...
+    const varsGlobalsProgram = path.join(testProgramsDir, 'vars_globals');
+    const varsGlobalsSrc = path.join(testProgramsDir, 'vars_globals.c');
+    const lineTags = {
+        INITIAL_STOP: 0,
+        RETURN: 0,
+    };
+
+    before(function () {
+        resolveLineTagLocations(varsGlobalsSrc, lineTags);
+    });
```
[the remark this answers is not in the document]
jonahgraham: Better still, make these tests work in non-remote.

jonahgraham:
I'm working on a real person review at the moment. I am using this PR to also test the coderabbit code review option that has been on my backlog. The code rabbit review is being doing in https://github.com/jonahgraham/cdt-gdb-adapter/pull/4

My review will be posted later.

src/integration-tests/var.spec.ts
```diff
@@ -620,3 +626,1195 @@ describe('Variables Test Suite', function () {
...
+
+    afterEach(async function () {
+        await dc.stop();
+    });
+
+    const findVar = (
+        vars: DebugProtocol.Variable[],
+        name: string
+    ): DebugProtocol.Variable => {
+        const v = vars.find((v) => v.name === name);
```
jonahgraham, line 1153: Why does remote vs non-remote has name in different form? If I search endsWith vars_global.c then this test can pass.

src/gdb/GDBDebugSessionBase.ts
```diff
@@ -2209,6 +2225,32 @@ export abstract class GDBDebugSessionBase extends LoggingDebugSession {
             ],
         };
 
+        if (this.globalSymbolsProvider) {
+            const frame = this.frameHandles.get(args.frameId);
+            if (!frame) {
+                this.sendErrorResponse(response, 1, 'Unknown frame');
```
jonahgraham, line 2231: Why does lack of valid frame fail the whole scopesRequest? Should it simply fail the to add globals to the list of scopes?

src/desktop/GNUObjdumpSymbolReader.ts
```diff
@@ -0,0 +1,1208 @@
...
+}
+
+interface ParsedDwarfInfo {
+    compilationUnits: DwarfCompilationUnit[];
+    diesByOffset: Map<number, DwarfDie>;
+}
+
+/**
+ * Mutable, streaming-friendly state for the "--dwarf=info" parser.
+ * Everything that used to be a local variable inside the parseDwarfInfo
```
jonahgraham, line 63: Try to remove this verbose AI comments that are related to earlier versions of the code that have now been discarded.

jonahgraham (commented):
I have completed a partial review and once some of these first questions are answered I can do a more complete one.

Please make sure to attribute AI that you are using (my assumption based on the code style and comments is AI was used to write this). See `Assisted-by:` in https://www.eclipse.org/projects/handbook/#genai

Does the dwarf information need to be processed? I though global symbol table was available without it. I believe that cortex debug doesn't use dwarf.

Have you performance tested this code against the large examples previously discussed? I attached a simply python script that helps generate huge projects. You can edit it to change types of variables and how many to create across how many files. [generate_c_project.py](https://github.com/user-attachments/files/32485613/generate_c_project.py)

> objdumpPath is currently only honored for target/remote debug sessions — a plain local launch always uses the GDB command regardless of objdumpPath. If you want to specifically test the objdump-backed path, please use a target/remote launch config.

Why? I don't understand why the target (aka remote) and local debug paths are different. Can you expand on why this may be? My best guess is this is related to a complication of Windows vs non-Windows toolchains. On Windows PE it won't be possible to use objdump. But on Linux ELF is standard, so objdump should work.

Also, a test for the rather complex objdump reader is needed. I created this one (which you can integrate into your PR) so I could step through the code a little more easily than running the full debug target.

```typescript
/*********************************************************************
 * Copyright (c) 2026 Kichwa Coders Canada Inc. and others
 *
 * This program and the accompanying materials are made
 * available under the terms of the Eclipse Public License 2.0
 * which is available at https://www.eclipse.org/legal/epl-2.0/
 *
 * SPDX-License-Identifier: EPL-2.0
 *********************************************************************/

import * as path from 'path';
import { expect } from 'chai';
import { GNUObjdumpSymbolReader } from '../desktop/GNUObjdumpSymbolReader';
import { testProgramsDir } from './utils';

describe('GNUObjdumpSymbolReader Test Suite', function () {
    const varsGlobalsProgram = path.join(testProgramsDir, 'vars_globals');
    let reader: GNUObjdumpSymbolReader;
    let variablesByFile: Map<string, string[]>;

    before(async function () {
        reader = new GNUObjdumpSymbolReader('objdump');
        variablesByFile =
            await reader.readGlobalVariablesByFile(varsGlobalsProgram);
    });

    const findFileEntry = (fileName: string): [string, string[]] => {
        const entry = [...variablesByFile.entries()].find(
            ([file]) => file.endsWith('/' + fileName) || file === fileName
        );
        expect(entry, `No entry found for source file '${fileName}'`).to.exist;
        return entry as [string, string[]];
    };

    it('associates the global variables with vars_globals.c', function () {
        const [, variables] = findFileEntry('vars_globals.c');
        expect(variables).to.include.members([
            's0',
            'p_s0',
            's1',
            'p_s1',
            'global_int',
        ]);
    });

    // TODO add tests for static variables (both scoped and global)
});

```

